Issue #4, opened 2021-01-29
Postgres jsonb support


Follpvosten:
I've noticed that the `.json()` datatype on ColumnDef results in the `json` datatype on the postgres side. Postgres has two json datatypes; `json` (which gets validated, but then stored in plain text) and `jsonb` (which gets stored in a decomposed binary format, so it is indexable and you can check for object containment and existence).

Looking at the [mysql docs](https://dev.mysql.com/doc/refman/8.0/en/json.html), it seems like the json datatype on the mysql side is the equivalent to the `jsonb` type (and they don't have an equivalent to postgres `json`, you're probably supposed to just use text).

Would it be an option to a) change the default `ColumnType::Json` type on the postgres side to yield `jsonb`, or b) change the `ColumnType::Json` type on the mysql side to yield `text`, and add a `json_binary()` type that results in `json` on mysql and `jsonb` on postgres?

tyt2y3:
Thank you for your suggestion! We are keen to make this change. JSON support on SQL databases is a messy matter, even MySQL and MariaDB are different. We are not a day to day user of Postgres, so its nice to have your opinion! 
When people use `json` in Postgres, do they usually use `json` or `jsonb`? I think option a) is a 'reasonable default', while b) is more being more explicit. I personally incline on b), but I want to know your preference as a heavy Postgres user.

Follpvosten:
"Heavy postgres user" is a bit too flattering for me, this is the first time I'm actually using it for a real project :D 

I've now looked at the mariadb side and wow, that really doesn't help this case. So MariaDB's json type with a JSON_VALID check is equivalent to the postgres `json` type, while MySQL's new json type is equivalent to postgres `jsonb`.

I'm basically also just learning this myself, so for reference, [here's all the info about json in postgres](https://www.postgresql.org/docs/current/datatype-json.html). In short, json is a bit cheaper on insert, but jsonb allows indexing and querying of the data, so it's more useful for data you might want to query/filter.
I would probably take the `.json_binary()` approach and just document that there's no difference on the mysql side (and possibly also document the difference between mariadb and mysql at that point, since mysql is technically always json_binary and mariadb is technically always json).

tyt2y3:
thank you for your input! this sounds like a reasonable solution at this point. we will implement this change and make a new release

PR #7, opened 2021-01-31
Column type json

tyt2y3:
#4


tyt2y3:
I think it is redundant to have Integer and IntegerDefault, as it is not a good use of the type system.
The refactor is internal implementation details that shouldnt break the API.
Please review @billy1624

billy1624 (commented):
Looks good! Thanks Chris

Issue #3, opened 2021-01-28
Provide a custom derive for Iden


Follpvosten:
Looking at how you have to manually implement Iden at the moment, it seems like a natural fit for a minimal derive macro.
Something like this:
```rust
#[derive(Iden)]
#[iden(rename_all = "snake_case")]
pub enum Character {
    #[iden("character")]
    Table,
    Id,
    Character,
    FontSize,
    SizeW,
    SizeH,
    FontId,
}
```

Going one step further, you could theoretically derive it on a struct as well, auto-generating the enum:

```rust
#[derive(IdenEnum, PostgresMapper)]
#[iden(table = "character")]
pub struct Character {
    id: i64,
    character: String,
    font_size: i64,
    size_w: i64,
    size_h: i64,
    font_id: Option<i64>,
}
// This would generate the enum above with an Iden implementation based on the struct's fields.
// As for the name of the enum, that could be chosen by another iden attribute, and default to
// something like {struct name}Def, ie CharacterDef.
// The implementation would transform the field names in the enum to PascalCase by default.
// And then you can directly use the struct with tokio-pg-mapper (Character::from_row)
```

This is mostly playing with ideas, but I'd be interested in contributing something like that if there is interest. Possibly something like that is already planned for SeaORM, but it would be nice to have it customizable and independent of the bigger ORM project.

Follpvosten:
I'm currently using this workaround to reduce the boilerplate:
```rust
#[derive(serde::Serialize)]
#[serde(rename_all = "snake_case")]
pub enum UserV1 {
    Table,
    Id,
    Name,
    Email,
}

impl Iden for UserV1 {
    fn unquoted(&self, s: &mut dyn Write) {
        match self {
            Self::Table => write!(s, "user"),
            f => write!(s, "{}", serde_json::to_string(f).unwrap()),
        }
        .unwrap();
    }
}
```
But the practicality and performance of that is questionable.

If you'd be interested in me potentially contributing a proc macro like this, please tell me; I'd be happy to start working on it if it lines up with the goals of sea-query.

tyt2y3:
SeaORM would include a code generator for these definitions so it'd be more ergonomic.
But for SeaQuery, yes it is manual and it'd be nice to ease the process a bit. I think providing macro is good as long as we don't force people into using them.
I think you nailed a smart hack! Never thought of implementing things this way.
[...]

PR #10, opened 2021-02-05
add the sea-query-derive code, update docs

Follpvosten:
I've moved the LICENSE files to the root of the repo again because they now apply to all of the crates, I've adjusted the links in the README, I've already integrated the crate into sea-query as an optional dependency, and I've updated the docs and README accordingly.

I mirrored how most big crates do the workspace dependencies (path deps). I think for publishing you'll have to first publish the derive crate, adjust the path dep to a version dep and the publish the main crate, but I have never done this, so I'm not sure.


tyt2y3:
#3

Issue #28, opened 2021-03-03
JOIN with alias


Follpvosten:
Support the following:
`LEFT JOIN table AS alias`
This is already possible for fields with `expr_alias` and for the FROM clause via `from_alias`, but not for joins (as far as I can tell, would love to be proven wrong). This is important as soon as you want to JOIN the same table twice.

Note that this might require an `Expr::alias(alias, column)` as well, but not necessarily since you can just use `Expr::tbl` with an  `Alias` everywhere.

tyt2y3:
Thank you! Would you mind if I rename the method to `join_as`?

tyt2y3:
Thank you! Included in `0.6.1`

Follpvosten:
Thanks!

PR #29, opened 2021-03-04
add join_alias methods to SelectStatement

Follpvosten:
#28 
I implemented this myself because I need it; I can also provide shortcuts like `.left_join_alias` etc if you'd like that.

### DIFF
--- a/src/query/select.rs
+++ b/src/query/select.rs
@@ -746,6 +746,48 @@ impl SelectStatement {
         self
     }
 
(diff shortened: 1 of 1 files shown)


tyt2y3:
Thank you! Would you mind if I rename the method to join_as?

Follpvosten:
That would be great! Maybe you can also rename expr_alias to expr_as while you're at it, but only if that's ok for you :)

Issue #34, opened 2021-04-16
Use IntoIterator in public api functions


Progdrasil:
In order to be more [idiomatic](https://rust-lang.github.io/api-guidelines/flexibility.html?highlight=IntoIterator#functions-minimize-assumptions-about-parameters-by-using-generics-c-generic) the use of generics would be preferable to asking for `Vec` as parameters.

Take for example SelectStatement::columns which has a signature as follows:
```rust
pub fn columns<T>(&mut self, cols: Vec<T>) -> &mut Self
    where T: IntoColumnRef
```
This could become
```rust
pub fn columns<T, I>(&mut self, cols: I) -> &mut Self
    where
        T: IntoColumnRef,
        I: IntoIterator<Item = T>
```

This stems from the discussion in #15.

duchainer:
Started https://github.com/duchainer/sea-query/tree/more-generic-into-iterator-instead-of-vec

tyt2y3:
Thanks! Let me know when you think it's ready.

tyt2y3:
Thank you so much of the big effort!

PR #37, opened 2021-04-22
More generic `IntoIterator` trait instead of `Vec` for public API

duchainer:
Fixes https://github.com/SeaQL/sea-query/issues/34


duchainer:
Don't hesitate to tell me if some of these changes are considered out of scope, for example https://github.com/SeaQL/sea-query/pull/37/commits/b8842ba5a08256bfaa3d92f0bd9a8a8fc64a2b87

Issue #43, opened 2021-05-14
Add traits for all queries for `build` and `to_string`


nitnelave:
Can we add a trait for all query types to factor the interface/implementation of `build` and `to_string`? That way, if we want a function that takes any query and a builder and builds the query, we can just be generic over the trait, not each specific query type.

WDYT?

tyt2y3:
Thank you for your suggestion. I am aware of the lack of a unified interface among different Statement builders. The problem is, query statements (DQL & DML) have value bindings but schema statements (DDL) do not.

I'd actually prefer to have the `QueryStatement` enum expose a `build` API and `SchemaStatement` enum expose another, and then impl a set of `Into<QueryStatement>` for `SelectStatement` etc.

PR #44, opened 2021-05-15
Implement traits for query builders and schema builders

nitnelave:
Fixes https://github.com/SeaQL/sea-query/issues/43#issuecomment-841167153

Two new traits are added:

- `traits::QueryStatementBuilder`, for queries (with values)
- `traits::SchemaStatementBuilder`, for schema-altering queries (without values).

They are imported publicly by every relevant statement builder types, so that users' code will not break due to a missing trait.

The `backend::SchemaBuilder` trait has also been added to gather the `backend::{TableBuilder, IndexBuilder, ForeignKeyBuilder}` traits. Since the `SchemaStatementBuilder` trait is now taking a `SchemaBuilder` as argument instead of the relevant sub-trait, there is the slight downside that it's now not possible to use the methods with a builder that implements `TableBuilder` but not `IndexBuilder` and `ForeignKeyBuilder`. However, all of the implementations in the crate implement all 3, and a search in GitHub for `backend::TableBuilder` (and others) brings up no result. While technically a breaking change, I don't think it's a big issue.

Some doc examples were removed (the `to_string`, `build` and `build_any` functions for the query statements) because they are present in the trait, and the example and usage is clear enough that I don't think the variant for each implementor is needed.

Note: due to my editor stripping the extra whitespace at the end of lines, I have a bigger diff than needed. I split the whitespace changes in a separate commit, which I can remove if needed.

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -21,59 +21,61 @@ pub use sqlite::*;
 
 pub trait GenericBuilder: QueryBuilder + TableBuilder + IndexBuilder + ForeignKeyBuilder {}
 
+pub trait SchemaBuilder: TableBuilder + IndexBuilder + ForeignKeyBuilder {}
+
 pub trait QueryBuilder {
-    
+
     /// Translate [`InsertStatement`] into SQL statement.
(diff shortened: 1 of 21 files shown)


tyt2y3:
Thank you for the contribution.
I would still like to maintain interface compatibility for now, so I added back some methods as wrappers over the trait implementation.
Hope this won't break your usage.

Issue #45, opened 2021-05-16
Replace `and_where` and `or_where` with `any` and `all`


nitnelave:
Currently, the way to write conditions is too close to SQL, yet different. To express `WHERE a AND b`, there are two possibilities:
`.and_where(a).and_where(b)` or `.or_where(a).and_where(b)`.

When we add more operations, things get messy:
`.and_where(a).or_where(b).and_where(c)` actually resolves differently depending on whether it's on a Select or a Delete/Update.

- In a Select, it will resolve to `WHERE a OR (b AND c)` due to the higher precedence of `AND` vs `OR`.
- In a Delete/Update, it will resolve to `WHERE (a OR b) AND c` because of enforced left-associativity.

In both cases, the semantics are not clear. Instead of `AND` and `OR`, the syntax should be based on `any` and `all`, and then resolved by the builder with the correct semantics. Something along these lines:

```rust
Query::delete()
  .from(Char::Table)
  .where(
    condition::any()
    .add(a)
    .add(
      condition::all()
      .add(b)
      .add(c)))
```

This would resolve to:
```sql
DELETE FROM char WHERE a OR (b AND c);
```

We can also add the macros `any!` and `all!` to get:
```rust
  .where(any![a, all![b, c]])
```

Regarding implementation, the existing `and_where` can be marked deprecated, and implemented as an `Either` with the any/all implementation, with a panic if both are used. Then with a major version bump we can remove the `and_where`.

WDYT?

tyt2y3:
I totally support adding a generic `where` method accepting an expression tree as parameter.
Regarding the naming, why not simply name them as `logical::and()` and `logical::or()`, this is more intuitive I guess?
But I don't think it is needed to deprecate `and_where`, as they make the resulting code more readable.
May be we can simply deprecate `or_where` and warn people against mixing `and_where` and `or_where`.

nitnelave:
There is precedent for the `any`/`all` with iterators: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.any
In my mind, `or` makes more sense for inline operations, but `any`/`all` are better (and more common) for tree-style conditions like this.

+1 to keeping the `and_where` and deprecating the `or_where`. We'd still have to make it incompatible with the generic `where` though.

PR #46, opened 2021-05-16
Create the OrderedStatement and ConditionalStatement traits

nitnelave:
This refactoring reduces the amount of code and increases the consistency between the different queries.

Based on https://github.com/SeaQL/sea-query/pull/44.

Note that this changes the semantics of `and_where` and `or_where` for
Delete and Update, to align them with Select. The difference is in the
operator precedence (see https://github.com/SeaQL/sea-query/issues/45)

Previously, we had:

```rust
Query::delete()
  .from_table(Glyph::Table)
  .or_where(Expr::col(Glyph::Id).eq(1))
  .or_where(Expr::col(Glyph::Id).eq(2))
  .and_where(Expr::col(Glyph::Id).eq(3))
```

equal to

```sql
DELETE FROM "glyph" WHERE (("id" = 1) OR ("id" = 2)) AND ("id" = 3)
```

I.e. it was using left-associativity. With this change, we now have the
less surprising:

```sql
DELETE FROM "glyph" WHERE "id" = 1 OR "id" = 2 AND "id" = 3
```

I.e. using the natural operator precedence from SQL, with AND before OR.
In that case, the parentheses would be:

```sql
DELETE FROM "glyph" WHERE "id" = 1 OR ("id" = 2 AND "id" = 3)
```

### DIFF
--- a/src/backend/mysql/query.rs
+++ b/src/backend/mysql/query.rs
@@ -133,9 +133,11 @@ impl QueryBuilder for MysqlQueryBuilder {
             false
         });
 
-        if let Some(wherei) = &update.wherei {
+        if !update.wherei.is_empty() {
             write!(sql, " WHERE ").unwrap();
-            self.prepare_simple_expr(wherei, sql, collector);
+        }
+        for (i, log_chain_oper) in update.wherei.iter().enumerate() {
+            self.prepare_logical_chain_oper(log_chain_oper, i, update.wherei.len(), sql, collector);
         }
 
         if !update.orders.is_empty() {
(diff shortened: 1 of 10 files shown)


tyt2y3:
First of all, thank you for the big contribution!
Yes, you are right. The idea for the AND & OR semantics was to 'delegate to  SQL' to resolve the associativity.
A refactor was done to the SELECT query at some point, and 'forgotten' for UPDATE and DELETE, causing the inconsistency.

nitnelave:
Rebased with the merge of https://github.com/SeaQL/sea-query/pull/44

PR #48, opened 2021-05-16
Factor builders' code into the trait

nitnelave:
The query builders have especially a lot in common, with very little difference between them. Factoring the code together shortens it and highlights the differences.

Based on https://github.com/SeaQL/sea-query/pull/47


tyt2y3:
Thank you for your contribution. I am still unsure about whether `OrderedStatement` and `ConditionalStatement` are implementation details that the user shouldn't need to know, or is a best practice to break interfaces down.

Anyway I added some shim to not break existing code for now. May we'd got to decide when approaching 1.0

WDYT?

nitnelave:
I think that `OrderedStatement` and `ConditionalStatement` make sense: you can reason the same way about all queries that implement them, and you have the guarantee that the functions are the same (like the issue with Delete and Update not being up to date in the condition implementation vs Select).

I'd say that it reduces cognitive load for users by reducing the interface, especially when you have things like `cond_where`, my other PR with the `any`/`all` change. By the way, I also want to expand that to `having` for select: it's as simple as introducing a `cond_having` method taking an `any`/`all`.

tyt2y3:
Thank you for your input. That makes sense. Is the other two PRs not needed now?

nitnelave:
You merged them already, so we're good :) I'll just prepare one last one to introduce `cond_having`, but that's going to be a small one.

PR #50, opened 2021-05-19
Refactor `Condition` 

tyt2y3:
Follow up on #49 

Since now `Condition` is shared between `where` and `having`, we might rename it as simply Condition.

And also, I think any() and all() being methods instead of free standing functions adhere more to the library's style

Agree? @nitnelave

### DIFF
--- a/src/query/condition.rs
+++ b/src/query/condition.rs
@@ -1,28 +1,28 @@
 use crate::{expr::SimpleExpr, types::LogicalChainOper};
 
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub enum ConditionWhereType {
+pub enum ConditionType {
     Any,
     All,
 }
 
 /// Represents the value of an [`any()`] or [`all()`]: a set of disjunctive or conjunctive conditions.
 #[derive(Debug, Clone)]
-pub struct ConditionWhere {
-    pub(crate) condition_type: ConditionWhereType,
+pub struct Condition {
+    pub(crate) condition_type: ConditionType,
     pub(crate) conditions: Vec<ConditionExpression>,
 }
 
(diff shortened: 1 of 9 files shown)


nitnelave:
Yeah, it makes sense. Thanks!

tyt2y3:
Thanks for the understanding. Do you have anything more to add? I'll prepare a new release.

src/query/condition.rs
```diff
@@ -1,28 +1,30 @@
 use crate::{expr::SimpleExpr, types::LogicalChainOper};
 
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub enum ConditionWhereType {
+pub enum ConditionType {
     Any,
     All,
 }
 
 /// Represents the value of an [`any()`] or [`all()`]: a set of disjunctive or conjunctive conditions.
```
nitnelave, line 9: Make sure you update the doc links to `any` to `Condition::any`

src/query/condition.rs
```diff
@@ -31,85 +33,85 @@ impl ConditionWhere {
...
-        condition_type: ConditionWhereType::All,
-        conditions: Vec::new(),
+impl std::convert::From<SimpleExpr> for ConditionExpression {
+    fn from(condition: SimpleExpr) -> Self {
+        ConditionExpression::SimpleExpr(condition)
     }
 }
 
-/// Macro to easily create an [`any()`].
+/// Macro to easily create an [`Cond::any()`].
```
nitnelave, line 114: Condition?

nitnelave (approved):

tyt2y3:
Sure I am adding one section to readme too.

PR #53, opened 2021-05-23
Create a SQLx-SQLite driver

### LINKED ISSUE #51: sqlx-sqlite driver
Right now we had sqlx-mysql and sqlx-postgres and rusqlite driver support, but without sqlx-sqlite.
It would be great to have 1) driver 2) usage example for `sqlx-sqlite`.

Hirtol:
Should fix #51 

It's mostly a copy paste from the sqlx-postgres driver.
The only real changes are that `TinyUnsigned` and `SmallUnsigned` don't have to be cast to u32, as SQLx has `Encode` implementations for u8 and u16 for SQLite.

The sqlx_example is much the same as the postgres one as well, making use of an in-memory database instead, as well as removing the `returning_col()` on the read example (which SQLite now supports, so could be something to consider adding in the future).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,7 @@ members = [
     "examples/rusqlite",
     "examples/sqlx_mysql",
     "examples/sqlx_postgres",
+    "examples/sqlx_sqlite",
 ]
 
 [package]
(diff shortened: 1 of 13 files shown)


tyt2y3:
Thank you so much for your contribution.

PR #60, opened 2021-06-04
Implements add_option for Condition

code-mm:
* Adds `add_option` to `Condition` struct.
* Adds unit tests for Postgres, Mysql, and Sqlite backends.

### DIFF
--- a/src/query/condition.rs
+++ b/src/query/condition.rs
@@ -65,6 +65,22 @@ impl Condition {
         self
     }
 
+    /// Add an optional condition to the set.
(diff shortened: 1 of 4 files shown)


code-mm:
Adds a method `add_option` to the `Condition` struct, accepting an `Option` as argument.
If option is `None` the condition is left empty and therefore not part of the query.
In certain cases a type annotation is required, due to the trait bound.

tyt2y3:
Thank you. You can now also use `and_where_option` too.

PR #92, opened 2021-08-06
remove anyhow as dependency

Progdrasil:
While working on some of the issues I raised, I found that anyhow was being used simply for its re-export of the [`Result`](https://doc.rust-lang.org/std/result/enum.Result.html) enum. This was without using anyhow's [`Error`](https://docs.rs/anyhow/1.0.42/anyhow/struct.Error.html) type.

So this just changes the type alias to use directly the standard library's `Result` and removes anyhow from the list of dependencies


billy1624:
Seems this is included in #93.

tyt2y3:
Nice catch. I forgot the reason we required on `anyhow` indeed.

tyt2y3:
close for #93

PR #93, opened 2021-08-06
#89 flattening iden enums

### LINKED ISSUE #89: flattening iden enums
Basically the equivalent to `#[serde(flatten)]` but for the `Iden` derive.

It allows us to be DRYer by using composition for common columns across many table.
For example for keeping track which user created an entry and when:

```rust
#[derive(Iden)]
enum Creation {
    CreatedBy,
    CreationDate,
}
#[derive(Iden)]
enum Product {
    Name,
    Price,
    #[iden(flatten)]
    CreationInfo(Creation),
}

#[derive(Iden)]
enum Sale {
    ProductId,
    Discount,
    EndDate,
   #[iden(flatten)]
    CreationInfo(Creation)
}
```

Currently I'm manually doing the following implementation which can easily be derived for expansion:

```rust

#[automatically_derived]
impl Iden for Product {
    fn unquoted(&self, s: &mut dyn std::fmt::Write) {
        write!(
            s,
            "{}",
            match self {
                Self::Name => "name",
                Self::Price => "price",
                Self::CreationInfo(c) => return c.unquoted(s),
            }
        )
        .unwrap()
    }
}

### PR DESCRIPTION
**Pull request by Progdrasil**
First let me apologize in advance for the monstrosity that is this pull request's size.

This pull request addresses #89. Even though the maintainers didn't voice any thoughts on the matter, I figured I'd try to implement it anyways rather than do a custom derive in my project.

I thought this was going to be a simple change but it turned out to be quite a endeavour. It ended up as pretty much an complete rewrite.

Along the way to add the `#[iden(flatten)]` attribute, I added the following:
* Error management instead of silently failing  on invalid attributes.
* 2 utility structs
  * `IdenAttr` manages the actual attribute used on a struct or variant.
  * `IdenVariant` manages a single enum variant according to its `IdenAttr` and the previously defined rules.
* tests for the derive crate. For this, a few test dependencies were added
  * [trybuild](https://docs.rs/trybuild/1.0.43/trybuild/)
  * [strum](https://docs.rs/strum/0.21.0/strum/)
  * and the sea-query crate for the trait.

`IdenVariant` has the additional advantage to implement `quote::ToTokens` which will expand into a match arm. This way you can just pass a variable of this type to `quote!{}` and it will expand according to the given rules.

For the tests I haven't touched the CI configuration since I'm not sure if you wish to keep test the workspace together or each crate separately.

Closes #89

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -53,14 +53,12 @@ jobs:
       - uses: actions-rs/cargo@v1
         with:
           command: build
-          args: --all-features
-        working-directory: sea-query-derive
+          args: --package sea-query-derive --all-features
 
       - uses: actions-rs/cargo@v1
         with:
           command: test
-          args: --all-features
-        working-directory: sea-query-derive
+          args: --package sea-query-derive --all-features
 
   sqlite:
     name: SQLite
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -36,6 +36,32 @@ jobs:
           command: test
           args: --all-features
 
+  derive-test:
+    name: Derive Tests
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+          override: true
+
+      - uses: Swatinem/rust-cache@v1
+
+      - uses: actions-rs/cargo@v1
+        with:
+          command: build
+          args: --all-features
+        working-directory: sea-query-derive
+
+      - uses: actions-rs/cargo@v1
+        with:
+          command: test
+          args: --all-features
+        working-directory: sea-query-derive
+
   sqlite:
     name: SQLite
     runs-on: ubuntu-20.04
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ target
 Cargo.lock
 *.sublime*
 .vscode
+wip/
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,7 +40,6 @@ chrono = { version = "^0", optional = true }
 postgres-types = { version = "^0", optional = true }
 rust_decimal = { version = "^1", optional = true }
 uuid = { version = "^0", optional = true }
-anyhow = { version = "^1" }
 thiserror = { version = "^1" }
 
 [features]
--- a/sea-query-derive/Cargo.toml
+++ b/sea-query-derive/Cargo.toml
@@ -18,3 +18,8 @@ syn = { version = "1", default-features = false, features = [ "derive", "parsing
 quote = "1"
 heck = "0.3"
 proc-macro2 = "1"
+
+[dev-dependencies]
+trybuild = "^1.0"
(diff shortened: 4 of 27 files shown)

### CODE REVIEW

**Comment by billy1624** [comment]
Hey @Progdrasil, thank you so much for the huge PR lolll.

I think we can also tests the `sea-orm-derive` on CI.

**Comment by Progdrasil** [comment]
Just watch out before merging it includes the commit from #92, I forgot to switch branches when i started work on this. I tried rebasing but it seems it didnt work.

**Comment by billy1624** [comment]
Hahaaa we mention the same thing a few seconds apart lol

**Comment by tyt2y3** [comment]
Hi there, nice work on the rewrite. You are now the derive macro guru!
I am not too concerned as it seems to pass our tests.
May I know what will happen if `flatten` is being applied at multiple levels?
I am not saying we absolutely should support from than 1 level flattening,
just want to define the expected behavior and perhaps panic if we can't support that.

**Comment by Progdrasil** [comment]
> I'd like to include the tests of sea-query-derive on CI too.

Do you want it run in a separate job, or in the same one as the other tests?

>May I know what will happen if flatten is being applied at multiple levels?

Since it's delegated it should be infinitely flattenable, I can add more tests for that though to make sure it works.
I'm pretty sure limiting it to one level is not really possible due to not having information on the inner enum.

> On the other hand, I think we should align the API by changing (or accepting both for now)
#[method = "custom_to_string"] to #[iden(method) = "custom_to_string"]

I agree, it would also simplify the code to handle the attribute having something like `#[iden(method = "custom_to_string")]`, `#[iden(rename = "custom_iden")]` and `#[iden(flatten)]` would make the api more uniform.
We can accept both, unfortunately I don't know how to do warning in derive macros to do a soft deprecation of the previous api.

**Comment by tyt2y3** [comment]
I tried to run the workflow though, and seeing
```
The workflow is not valid. .github/workflows/rust.yml (Line: 57, Col: 9): Unexpected value 'working-directory' .github/workflows/rust.yml (Line: 63, Col: 9): Unexpected value 'working-directory'
```

**Comment by Progdrasil** [comment]
I tried making derive run in a separate workflow, but I'm not really familiar with github actions. I found `working-directory` through a stack overflow link. I'm not too sure how to make it run in a subdirectory.

Otherwise the alternative wouldbe to add the argument `--workspace` in the tests to run everything in one job.

**Comment by tyt2y3** [comment]
It seems that `uses` is not compatible with `working-directory`
but may be we can simply use the `--package` argument of `cargo`

**Comment by billy1624** [comment]
Hey @Progdrasil, can you try something like

```
derive-test:
    name: Derive Tests
    runs-on: ubuntu-20.04
    steps:
      - uses: actions/checkout@v2

      - uses: actions-rs/toolchain@v1
        with:
          profile: minimal
          toolchain: stable
          override: true

      - uses: Swatinem/rust-cache@v1

      - uses: actions-rs/cargo@v1
        with:
          command: build
          args: --manifest-path sea-query-derive/Cargo.toml

      - uses: actions-rs/cargo@v1
        with:
          command: test
          args: --manifest-path sea-query-derive/Cargo.toml
```

Progdrasil:
I tried the first suggestion with the `--package` argument

PR #94, opened 2021-08-09
Add support for ESCAPE

Sytten:
Hello!
First contribution here :)
I am unsure if the way I implemented it is correct (specially in the Expr) but it does work as expected.

### DIFF
--- a/src/backend/query_builder.rs
+++ b/src/backend/query_builder.rs
@@ -603,6 +603,7 @@ pub trait QueryBuilder: QuotedBuilder {
                 BinOper::Mul => "*",
                 BinOper::Div => "/",
                 BinOper::As => "AS",
+                BinOper::Escape => "ESCAPE",
                 #[allow(unreachable_patterns)]
                 _ => unimplemented!(),
             }
@@ -1403,7 +1404,7 @@ pub trait QueryBuilder: QuotedBuilder {
         sql: &mut SqlWriter,
         collector: &mut dyn FnMut(Value),
     ) {
-        let no_paren = matches!(op, BinOper::Equal | BinOper::NotEqual);
+        let no_paren = matches!(op, BinOper::Equal | BinOper::NotEqual | BinOper::Escape);
         let left_paren = left.need_parentheses()
             && left.is_binary()
             && *op != left.get_bin_oper().unwrap()
--- a/src/expr.rs
+++ b/src/expr.rs
@@ -2065,6 +2065,36 @@ impl SimpleExpr {
         self.binary(BinOper::Sub, right.into())
     }
 
+    /// Select an espace character in a LIKE .
+    ///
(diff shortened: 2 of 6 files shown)


tyt2y3:
Thank you so much for your contribution. I have a puzzle, why is the `\` in the last don't have to be escaped themselves?

```SQL
SELECT `character` FROM `character` WHERE `character` LIKE 'A' ESCAPE '\'
```
vs

```SQL
SELECT `character` FROM `character` WHERE `character` LIKE 'A' ESCAPE '\\'
```

Sytten:
I think it will escape it automatically? I compared it against an r string so thats why it might be misleading. AFAIK you dont need to escape it further than that.
We used a custom expression before I wrote that and it didnt require an explicit escape. I think it might be because the input is a char not a string?
I will test further to make sure.

ikrivosheev:
@Sytten do you want to finish this PR?

Sytten:
I still need it but I have not had the time, I will circle back today and let you know.

ikrivosheev:
@Sytten if you want, I can do this PR.

Sytten:
@ikrivosheev @tyt2y3 @billy1624 
I fixed the PR to support the latest version.
It works correctly, but I am not sure if we should merge as is since the escape only makes sense when following a like so I think a separate struct for Like would be better (but it would most likely be a breaking change).
Maybe @ikrivosheev you want to take a shot at it?

ikrivosheev:
@tyt2y3 @billy1624 perhaps we can transform `BinOper::Like` -> 
```
pub struct Like {
    pattern: String,
    escape: Option<String>
}
```

and move to `SimpleExpr`? I cannot image how it makes as part of `BinOper`

billy1624:
For me API wise, current API is better

```rust
.and_where(Expr::col(Char::Character).like("A").escape('\\'))
```

If we require user to specify `ESCAPE` when constructing `LIKE`, unavoidably the API would be like

```rust
.and_where(Expr::col(Char::Character).like("A"))
.and_where(Expr::col(Char::Character).like_escape("A", '\\'))
```

Sytten:
Agreed the API would not change but the return type of like would.
The problem with the current api is you can write:
```rust
.and_where(Expr::col(Char::Character).escape('\ \'))
```
Which is not valid.

ikrivosheev:
@billy1624 @Sytten I think we can safe API, no problem)

Sytten:
Do you want to take it over @ikrivosheev?

ikrivosheev:
@Sytten can I merge you PR into internal branch (of course with save your commits) and continue work in internal branch?

Sytten:
Go ahead!

ikrivosheev:
@Sytten https://github.com/SeaQL/sea-query/pull/318

PR #97, opened 2021-08-11
Implement `IntoSelf` trait for `Self`

billy1624:
Related to SeaQL/sea-orm#65


tyt2y3:
Done?

billy1624:
Should we implement for all other similar `IntoSelf` trait?

PR #98, opened 2021-08-11
Postgres fulltext search support

code-mm:
* Adds library feature `postgres-fulltext-search`.
* Adds operators (matches, contains, contained).
* Adds helper methods and unit tests for operators to `Expr`.
* Adds Postgres database functions and unit tests to `Function` and `Func`.

A few remarks.
* The add methods on `Expr` take a `Into<SimpleExpr>` in contrast to other operator methods like `eq`, which take a `Into<Value>`. That's because most of the time they are used with the also added database functions.
* The naming of the functions on `Func` is according to the Postgres function naming (snake case to camel case). Looks a bit odd for some of them, but I preferred to stick to them precisely.
* I represent the Postgres `Regconfig` data type (text search configuration) as `u32` as it's an 4 byte integer in the database anyway. This could be improved with an implemented data type like `enum` or `const` in the future.
* The functions `ts_rank` and `ts_rank_cd` have fewer parameters than the actual database functions. This should be improved in the future.


tyt2y3:
First of all thank you for the big commit! It looks good so far.

There is one thing, I indeed planned to include all built-in functions from supported databases for convenience.
So, the design was, to have `Func` be functions that can be abstracted over different databases,
and have separate files and struct perhaps called `PgFunc`, `MyFunc`, `LiteFn`,
so that these files can be feature guarded by `backend-postgres` etc.

What's your view on it? Will `PgFunc` be unergonomic for you?

code-mm:
> What's your view on it? Will `PgFunc` be unergonomic for you?
No, that sounds reasonable to me. I was actually thinking the same about the binary operators, but it might be a bit more difficult for those. Do you want to have the `PgFunc` struct also in the `func.rs` file?

> Oh, and perhaps if we add a nested enum variant (now that Rust has nested pattern matching!) to the `Function` enum, it could saves us many lines of `#[cfg(feature = "postgres-fulltext-search")]` too.
Never used that. Good suggestion! I will have a look...

tyt2y3:
In that case, we can simply move it to a `pg_func.rs` file, and use only one guard on the mod.
Looking through the code again, I think we don't need a separate feature `postgres-fulltext-search`, just reuse the `backend-postgres` feature guard is suffice!
I think people using Postgres will want to have access to these operators and functions anyway.
But we need to think about the optional arguments of built-in functions, can we simply have the optional arguments wrapped in `Option`?
Thank you so much for everything up to now.

code-mm:
> Looking through the code again, I think we don't need a separate feature `postgres-fulltext-search`, just reuse the `backend-postgres` feature guard is suffice.
> I think people using Postgres will want to have access to these operators and functions anyway.

I removed the `postgres-fulltext-search` feature and used the `backend-postgres` on the new module.

> But we need to think about the optional arguments of built-in functions, can we simply have the optional arguments wrapped in `Option`?

I already did it that way on the helper methods for the `regconfig` parameters. I currently don't see any problem with this approach.

I moved the text search functions from the `Function` struct to a `PgFunction` struct in the `pg_func` module.
The only thing I don't like in the current implementation is that we would have a nested pattern match for the Postgres functions in the query builder. I would suggest to move this to a separate function in the query builder to match all Postgres functions there.
What do you think?

tyt2y3:
Thank you for the great work all in all!
I think for the purpose of this PR we are done )
I will think about how to refactor and structure the code better.
But those are internal details that won't impact the API, so I can probably take my time.
I will make a new release soon.

code-mm:
Thank you for accepting my PR! :-)

tyt2y3:
[0.12.12](https://github.com/SeaQL/sea-query/releases/tag/0.12.12)

Issue #78, opened 2021-07-30
Align ColumnType and Value variants


forrest-akin:
It would be nice to have the expected mappings from [database types](https://www.postgresql.org/docs/current/datatype.html#DATATYPE-TABLE) to [Value](https://docs.rs/sea-query/0.12.8/sea_query/value/enum.Value.html) types, similar to [sqlx](https://docs.rs/sqlx/0.5.5/sqlx/postgres/types/index.html#types). It's been challenging figuring out which `Value` variant I need to construct when setting a default value for a column in a table schema based on a dynamic column type, especially since there isn't a one-to-one mapping of [`ColumnType`](https://docs.rs/sea-query/0.12.8/sea_query/table/enum.ColumnType.html) to Value.

tyt2y3:
You usually do not have to specify the Value variant manually. The intended usage is `into`, e.g.
```rust
let v: Value = 123.into();
```

It is the underlying driver's responsibility to bind the respective value type to the binary protocol. So yes, if you are using SQLx as driver, you'd look at it's documentation. The responsibility of the `Value` enum is only to hold the query parameters temporarily until it's time for prepare. 

So yes, there is not a one-to-one mapping for column type to value type. For example, when comparing a value with a decimal column, you can using any numeric values.

I agree that could be a bit confusing though. At what scenario you need specify the variant manually?

forrest-akin:
My use case is dynamically generating `CREATE TABLE` statements so based on the column type I need to know what rust type to cast the input to before converting to a `Value`. So you're right, I'm not constructing `Value` variants manually (I'm using `Value::from`), but I do need to know which rust type to cast to based on the column type.

PR #107, opened 2021-08-19
Revamp Value

billy1624:
## The problem

When using a prepared statement, the parameter binding must be of the same type as when prepared. As such, `0i32` is not compatible with `None<bool>`, it has to be `None<i32>`. To mitigate this, we have to remove the generic `Null` variant, and users have to indicate the associated type of the `NULL`. 

## The changes

- Fix Postgres integration tests (SeaQL/sea-orm#91) with typed null value
- Rename and add new enum variants to `Value` for type better mapping #78 

It will be a rather big breaking change, but hopefully users would not have to change a lot of their existing code.
An upgrade guide will be provided afterwards.

### DIFF
--- a/src/value.rs
+++ b/src/value.rs
@@ -7,15 +7,20 @@ use serde_json::Value as Json;
 use std::str::from_utf8;
 
 #[cfg(feature = "with-chrono")]
-use chrono::{DateTime, FixedOffset, NaiveDateTime};
+use chrono::{DateTime, FixedOffset, NaiveDate, NaiveDateTime, NaiveTime};
 
 #[cfg(feature = "with-rust_decimal")]
 use rust_decimal::Decimal;
+
+#[cfg(feature = "with-bigdecimal")]
+use bigdecimal::BigDecimal;
 
 #[cfg(feature = "with-uuid")]
 use uuid::Uuid;
 
 /// Value variants
+///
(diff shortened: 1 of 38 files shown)


src/value.rs
```diff
@@ -569,32 +652,60 @@ pub fn sea_value_to_json_value(value: &Value) -> Json {
...
+        Value::Bytes(Some(s)) => Json::String(from_utf8(s).unwrap().to_string()),
+        Value::Json(Some(v)) => v.as_ref().clone(),
+        #[cfg(feature = "with-chrono")]
+        Value::Date(_) => PostgresQueryBuilder.value_to_string(value).into(),
+        #[cfg(feature = "with-chrono")]
+        Value::Time(_) => PostgresQueryBuilder.value_to_string(value).into(),
         #[cfg(feature = "with-chrono")]
         Value::DateTime(_) => PostgresQueryBuilder.value_to_string(value).into(),
         #[cfg(feature = "with-chrono")]
         Value::DateTimeWithTimeZone(_) => PostgresQueryBuilder.value_to_string(value).into(),
```
tyt2y3, line 696: I realized it cannot compile if the Postgres backend is not enabled. I will refactor it later.

billy1624: Thanks!!!

PR #110, opened 2021-08-24
Format better and remove unnecessary borrows

benarmstead:
Title self explanatory


tyt2y3:
Thank you so much for the contribution.
Do share if us, if you have a bigger plan.

benarmstead:
> Thank you so much for the contribution.
> Share with us, if you have a bigger plan )

No problem at all, if I have any ideas I will let you know.

PR #117, opened 2021-08-30
Initial lock support

nappa85:
As discussed in https://github.com/SeaQL/sea-orm/issues/111, this PR adds the basis for row locking support.

As always, I'm open for discussion and improvement

### DIFF
--- a/src/backend/query_builder.rs
+++ b/src/backend/query_builder.rs
@@ -343,7 +343,7 @@ pub trait QueryBuilder: QuotedBuilder {
         .unwrap();
     }
 
-    /// Translate [`SelectDistinct`] into SQL statement.
+    /// Translate [`LockType`] into SQL statement.
     fn prepare_select_lock(
         &self,
         select_lock: &LockType,
--- a/src/backend/query_builder.rs
+++ b/src/backend/query_builder.rs
@@ -121,6 +121,11 @@ pub trait QueryBuilder: QuotedBuilder {
             write!(sql, " OFFSET ").unwrap();
             self.prepare_value(offset, sql, collector);
         }
+
+        if let Some(lock) = &select.lock {
+            write!(sql, " ").unwrap();
+            self.prepare_select_lock(lock, sql, collector);
+        }
     }
 
     /// Translate [`UpdateStatement`] into SQL statement.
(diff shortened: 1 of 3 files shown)


src/query/select.rs
```diff
@@ -1345,3 +1366,10 @@ impl ConditionalStatement for SelectStatement {
         self
     }
 }
+
+/// List of lock types
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum LockType {
```
tyt2y3, line 1372: Very tiny comment here. I usually put type definitions before implementations.

src/backend/query_builder.rs
```diff
@@ -338,6 +343,24 @@ pub trait QueryBuilder: QuotedBuilder {
...
+    fn prepare_select_lock(
+        &self,
+        select_lock: &LockType,
+        sql: &mut SqlWriter,
+        _collector: &mut dyn FnMut(Value),
+    ) {
+        write!(
+            sql,
+            "{}",
+            match select_lock {
```
tyt2y3, line 356: This trait impl things that are common across MySQL, Postgres and SQLite. Seemingly we cannot provide a default impl here, as things are different for Postgres and SQLite (which has no row based lock indeed).

nappa85: To be honest, I didn't check for sqlite before, and in fact now I find out it doesn't support row locking.
Postgres supports the same lock commands as MySQL, the docs are a little confusionary, all the lock types described are used internally.

I think we can leave this impl here and overwrite it in sqlite part

nappa85: How do you disable features in a single DB flavour?
ATM I've put a panic, but I don't feel good with it, at the same time writing code that does nothing IMHO isn't good too.

src/query/select.rs
nappa85: I've copied the comment style for SelectDistinct

src/backend/query_builder.rs
tyt2y3: I think 'do nothing' is the usual behavior

nappa85: Ok, I've removed the panic

src/query/select.rs
nappa85: Oh, I've realized only now you maybe intended the comments on the lock methods...
I was waiting for your approval on the design before spending time on docs and examples

tyt2y3: Yes, the API looks good to me.

nappa85: Ok, I've added some docs and examples

PR #122, opened 2021-09-01
Postgres chained concatenate expression

billy1624:


billy1624:
Do you think having a alias `concat` for `concatenate` would be good? @tyt2y3

tyt2y3:
> Do you think having an alias `concat` for `concatenate` would be good? @tyt2y3

Sounds nice!

billy1624:
> Do you think having an alias `concat` for `concatenate` would be good? @tyt2y3

Added

src/expr.rs
```diff
@@ -1733,4 +1743,44 @@ impl SimpleExpr {
...
+    pub fn concatenate<T>(self, right: T) -> Self
+    where
+        T: Into<SimpleExpr>,
+    {
+        self.binary(BinOper::Concatenate, right.into())
+    }
+
+    /// Alias of [`SimpleExpr::concatenate`]
+    #[cfg(feature = "backend-postgres")]
+    pub fn concat<T>(self, right: T) -> Self
```
tyt2y3, line 1780: Yes. an alias is nice

tyt2y3 (approved):

PR #125, opened 2021-09-02
Joins accept any IntoCondition

### LINKED ISSUE #124: `join()` which accept Condtion instead of SimpleExpr?
> Is this a good idea to add func join_cond() which accept Condtion instead of SimpleExpr ?

Regarding the change of API... I think we can do this without breaking anything and always use `Condition` internally? @tyt2y3 

```rust
pub fn join<R, C>(&mut self, join: JoinType, tbl_ref: R, condition: C) -> &mut Self
where
    R: IntoTableRef,
    C: IntoCondition,
{
    // ...
}
```

Considering we have...

```rust
impl IntoCondition for SimpleExpr {
    fn into_condition(self) -> Condition {
        Condition::all().add(self)
    }
}

impl IntoCondition for Condition {
    fn into_condition(self) -> Condition {
        self
    }
}
```

_Originally posted by @billy1624 in https://github.com/SeaQL/sea-query/discussions/123#discussioncomment-1269599_

billy1624:
Resolve #124

### DIFF
--- a/src/backend/query_builder.rs
+++ b/src/backend/query_builder.rs
@@ -601,10 +601,7 @@ pub trait QueryBuilder: QuotedBuilder {
         collector: &mut dyn FnMut(Value),
     ) {
         match join_on {
-            JoinOn::Condition(c) => {
-                write!(sql, "ON ").unwrap();
-                self.prepare_simple_expr(c, sql, collector);
-            }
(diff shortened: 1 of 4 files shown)


tyt2y3 (changes requested):
Seems like we need some test cases?

PR #130, opened 2021-09-10
Add getters to TableForeignKey

billy1624:
Related to SeaQL/sea-orm#143


tyt2y3 (commented):
I was trying to say, if we make the enum impl Copy, we don't have to clone

billy1624:
> I was trying to say, if we make the enum impl Copy, we don't have to clone

Good call!

tyt2y3 (approved):

PR #66, opened 2021-06-23
Implements returning and returning_col for update statement

code-mm:
* Adds `returning` and `returning_col` to `UpdateStatement`.
* Adds doc tests for `returning` and `returning_col` methods.
* Adds preparation call to `QueryBuilder` for update statements in Postgres backend.

PR #136, opened 2021-09-14
Support `returning` for `DeleteStatement`

### LINKED ISSUE #120: Support `returning` for `DeleteStatement`
Hi, i'm trying to write delete statement that return deleted rows:
```sql
delete from sessions
where id = any($1::uuid[])
returning id, account, expire_at, created_at
```
and it seems that `DeleteStatement` missing `returning` method.

Although I can find `returning` method for `InsertStatement` and `UpdateStatement`.

MuhannadAlrusayni:
Fix #120

### DIFF
--- a/src/backend/query_builder.rs
+++ b/src/backend/query_builder.rs
@@ -215,6 +215,8 @@ pub trait QueryBuilder: QuotedBuilder {
             write!(sql, " LIMIT ").unwrap();
             self.prepare_value(limit, sql, collector);
         }
+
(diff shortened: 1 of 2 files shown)


MuhannadAlrusayni:
I'm not sure if I done this correctly since all I did was coping #66, I have run `cargo test && cargo build` without any issue.

tyt2y3 (approved):
Thank you very much. Looks good indeed.

PR #137, opened 2021-09-17
Postgres driver bind `DateTime<FixedOffset>`

billy1624:


tyt2y3:
May be we can add some test case by expanding the `sqlx_postgres` example.
Btw, the `postgres` example already has.
https://github.com/SeaQL/sea-query/blob/20d1abc5aae3aa1561324f71e60afefe304d210f/examples/postgres_json/src/main.rs#L53

billy1624:
> May be we can add some test case by expanding the `sqlx_postgres` example.
> 
> Btw, the `postgres` example already has.
> 
> https://github.com/SeaQL/sea-query/blob/20d1abc5aae3aa1561324f71e60afefe304d210f/examples/postgres_json/src/main.rs#L53

We can do it on another PR

PR #140, opened 2021-09-22
Add `QueryValue<DB>` trait for dynamic types

Acidic9:
### Summary

This PR creates a trait `QueryValue<DB>` which can be used in place of the `Value` enum for support for custom types.

### Problem

Using custom types in queries is not supported, as the `Value` enum is used as the type accepted by functions throughout sea-orm.

### Solution

A new trait `QueryValue<DB>` has been added which replaces almost all instances where `Value` is used, allowing support for custom types that implement this new trait.

### How it's implemented

Anywhere `Value` could previously be used has been replaced with the `QueryValue<DB>` trait. This means most types now have a `DB` generic attached to it.

The definition for this new trait:
```rust
/// Indicates that a SQL type is supported for use in queries.
/// Convert a value to a String for use in queries.
pub trait QueryValue<DB> {
    /// Returns the value as an escaped string safe for use in queries.
    fn query_value(&self) -> String;
}
```

Most primitive types implement QueryValue in this commit, including integers, floats, strings, bools, DateTime.

The major benefit of this change is that users can define new types for use in queries.
For example:
```rust
struct Email(String);

impl<DB> QueryValue<DB> for Email {
    fn query_value(&self) -> String {
        let mut buf = String::new();
        DB::write_string_quoted(self.0, &mut buf);
        buf
    }
}
```
```rust
enum Gender {
    Male,
    Female,
}

impl<DB> QueryValue<DB> for Gender {
    fn query_value(&self) -> String {
        let mut buf = String::new();
        match self {
            Self::Male => DB::write_string_quoted("male", &mut buf);
            Self::Female => DB::write_string_quoted("female", &mut buf);
        }
        buf
    }
}
```
The types can then be used in queries as you previously could only with `Value`:
```rust
let email = Email("example@domain.com".to_string());
let gender = Gender::Male;

let query = Query::update()
    .table(User::Table)
    .value(User::Name, &"John Doe")
    .value(User::Email, &email)
    .value(User::Gender, &gender)
    .and_where(Expr::col(User::Id).eq(&1))
    .to_owned();
```

This is a major change to the project, and the main reason behind it is for support for SeaORM custom types. Eg, enums, etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,11 +8,11 @@
     <strong>🌊 A dynamic query builder for MySQL, Postgres and SQLite</strong>
   </p>
 
-  [![crate](https://img.shields.io/crates/v/sea-query.svg)](https://crates.io/crates/sea-query)
-  [![docs](https://docs.rs/sea-query/badge.svg)](https://docs.rs/sea-query)
-  [![build status](https://github.com/SeaQL/sea-query/actions/workflows/rust.yml/badge.svg)](https://github.com/SeaQL/sea-query/actions/workflows/rust.yml)
+[![crate](https://img.shields.io/crates/v/sea-query.svg)](https://crates.io/crates/sea-query)
+[![docs](https://docs.rs/sea-query/badge.svg)](https://docs.rs/sea-query)
+[![build status](https://github.com/SeaQL/sea-query/actions/workflows/rust.yml/badge.svg)](https://github.com/SeaQL/sea-query/actions/workflows/rust.yml)
 
-  <sub>Built with 🔥 by 🌊🦀🐚</sub>
+<sub>Built with 🔥 by 🌊🦀🐚</sub>
 
 </div>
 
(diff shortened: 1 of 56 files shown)


tyt2y3:
Thank you for the big PR.

Though given the scope of changes, it scares me a bit. I am unable to understand what is `DB`? Why is it not bounded by any trait?

Regarding the change in API, in any case, we should not change the usage that abruptly.

I noticed 2 usage changes:
1) having to `&1` instead of `1`
2) having to `to_string::<PostgresQueryBuilder>()` (a turbo fish)

Is there any way we can avoid those?
For 2, you can probably reference https://github.com/SeaQL/sea-orm/blob/master/DESIGN.md#turbofish-and-inference

Acidic9:
Closing in favour of https://github.com/SeaQL/sea-query/pull/146

PR #145, opened 2021-09-28
Add Condition::not

6xzo:
See https://github.com/SeaQL/sea-orm/issues/147

See https://github.com/SeaQL/sea-query/discussions/143


tyt2y3:

PR #146, opened 2021-09-29
Replace `Value` enum with `dyn QueryValue`

Acidic9:
This PR deprecated the old `Value` enum, and creates a **new** struct: `Value(Box<dyn QueryValue>)`.

Any type that implements the new trait `QueryValue` can be used as a `Value`.
```rust
/// Indicates that a type is supported for use in SQL queries.
pub trait QueryValue: Clone + PartialEq {
    /// Returns the value as an escaped string safe for use in SQL queries.
    fn query_value(&self, query_builder: &dyn QueryBuilder) -> String;
}
```

Many of the primitive types implement QueryValue, including:
`&str`, `String`, `i8`, `i16`, `i32`, `i64`, `u8`, `u16`, `u32`, `u64`, `f32`, `f64`, `bool`, `()` (represents NULL), `Vec<u8>`, `serde_json::Value`, `chrono::NaiveDate`, `chrono::NaiveTime`, `chrono::NaiveDateTime`, `chrono::DateTime<chrono::FixedOffset>`, `uuid::Uuid`, `rust_decimal::Decimal`, `bigdecimal::BigDecimal`.

This means they can be used in queries (as they would previously), and calling `.into()` where necessary.


tyt2y3:
Thank you for the effort.

There seems to be a lot of useful ideas!

1. `().into()`
2. the `Default` on `ColumnSpec`
3. `V: Into<Value>` on `Update::value`

And the impact to users is seemingly small.

My only wonder is, what new possibilities does this change bring to users?

Acidic9:
> My only wonder is, what new possibilities does this change bring to users?

I finally can answer this question with the PR in Sea ORM https://github.com/SeaQL/sea-orm/pull/214

It allows support for `CAST($1 as type)`, which lets you cast into custom types, such as enums in Postgres.

src/primitive_value.rs
```diff
@@ -71,6 +74,58 @@ pub enum PrimitiveValue {
...
+            PrimitiveValue::SmallInt(value) => value.query_value(query_builder),
+            PrimitiveValue::Int(value) => value.query_value(query_builder),
+            PrimitiveValue::BigInt(value) => value.query_value(query_builder),
+            PrimitiveValue::TinyUnsigned(value) => value.query_value(query_builder),
+            PrimitiveValue::SmallUnsigned(value) => value.query_value(query_builder),
+            PrimitiveValue::Unsigned(value) => value.query_value(query_builder),
+            PrimitiveValue::BigUnsigned(value) => value.query_value(query_builder),
+            PrimitiveValue::Float(value) => value.query_value(query_builder),
+            PrimitiveValue::Double(value) => value.query_value(query_builder),
+            PrimitiveValue::String(value) => value.as_deref().cloned().query_value(query_builder),
```
Acidic9, line 91: I couldn't find a way to do this without `.cloned()`. I think some changes need to be made in `impl<T> QueryValue for Option<T>`, but I had issues getting it to work.

Acidic9:
With the new issue made https://github.com/SeaQL/sea-orm/issues/252, I wil close this PR for now. If needed it can be reopened.

On a side note, I feel over using dynamic trait objects is tempting but not always the solution.

PR #167, opened 2021-10-18
[Draft] support upsert sql 

baoyachi:
Inprogres work for https://github.com/SeaQL/sea-orm/issues/187

### DIFF
--- a/src/backend/query_builder.rs
+++ b/src/backend/query_builder.rs
@@ -1,3 +1,5 @@
+use crate::query::upsert::ActionExpr;
+use crate::upsert::{ConflictExpr, UpsertExpr};
 use crate::*;
 
 pub trait QueryBuilder: QuotedBuilder {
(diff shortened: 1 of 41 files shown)


billy1624:
Hey @baoyachi, sorry for the delay.

I'm on the way home now and I wonder if MySQL & SQLite supports similar upcert syntax?

Progdrasil:
[SQLite supports it](https://www.sqlite.org/lang_UPSERT.html) but MySQL is ... [Complicated](https://dev.mysql.com/doc/refman/8.0/en/insert-on-duplicate.html)

src/query/upsert.rs
```diff
@@ -0,0 +1,180 @@
...
+        let query = Query::insert()
+            .into_table(Glyph::Table)
+            .columns(vec![Glyph::Image])
+            .upsert(UpsertExpr::do_conflict_nothing())
+            .returning(Query::select().column(Glyph::Id).take())
+            .to_owned()
+            .to_string(PostgresQueryBuilder);
+
+        println!("{}", query);
+    }
```
billy1624, line 122: For all these test cases could we assert it against a String?

billy1624 (changes requested):

billy1624:
> [SQLite supports it](https://www.sqlite.org/lang_UPSERT.html) but MySQL is ... [Complicated](https://dev.mysql.com/doc/refman/8.0/en/insert-on-duplicate.html)

I see... basically MySQL don't have `DO NOTHING` and any extra conditioning. It just have `ON DUPLICATE KEY UPDATE`
```sql
INSERT [LOW_PRIORITY | DELAYED | HIGH_PRIORITY] [IGNORE]
  [INTO] tbl_name [PARTITION (partition_list)] [(col,...)]
  {VALUES | VALUE} ({expr | DEFAULT},...),(...),...
  [ ON DUPLICATE KEY UPDATE
    col=expr
      [, col=expr] ... ]
```

REF:
- https://mariadb.com/kb/en/insert-on-duplicate-key-update/
- https://dev.mysql.com/doc/refman/8.0/en/insert-on-duplicate.html

src/backend/query_builder.rs
```diff
@@ -48,9 +50,116 @@ pub trait QueryBuilder: QuotedBuilder {
...
         self.prepare_returning(&insert.returning, sql, collector);
     }
 
+    /// Upsert Statement
+    fn prepare_upsert(
+        &self,
+        upsert: &UpsertExpr,
+        sql: &mut SqlWriter,
+        collector: &mut dyn FnMut(Value),
+    ) {
```
billy1624, line 66: We should implement a separate version of `prepare_upsert` for MySQL

jamiepine:
Noticed the author of this wonderful PR hasn't returned to make the requested changes. I would love to help get this one moving as I need it in my project. 

Though, I am wondering if this will work for `insert_many` operations?

billy1624:
> Though, I am wondering if this will work for `insert_many` operations?

I assumed you are talk PostgreSQL specifically. I found the following snippet online.

```sql
INSERT INTO med.coinprices (symbol, currency, price_usd, volume_24h) VALUES
	('BTC', 'Bitcoin', 60000, 40000000000)
	, ('XDG', 'Dogecoin', 0.2181, 1983534547)
ON CONFLICT (symbol, currency)
DO UPDATE SET
	price_usd = EXCLUDED.price_usd
	, volume_24h = EXCLUDED.volume_24h;
```

https://towardsdatascience.com/postgresql-how-to-upsert-safely-easily-and-fast-246040514933

billy1624:
Just found MySQL & SQLite have similar features.

- SQLite also have a `excluded` keyword: https://www.sqlite.org/lang_UPSERT.html
- While, MySQL have a keyword named `new`: https://dev.mysql.com/doc/refman/8.0/en/insert-on-duplicate.html

jamiepine:
Oh I should have mentioned, in my case I'm using SQLite. I only just discovered that it supports `ON CONFLICT IGNORE` in the schema itself, so that was able to solve my issue with `insert_many`.
```sql
CREATE TABLE files (
-- ...
   meta_checksum TEXT NOT NULL UNIQUE ON CONFLICT IGNORE,
-- ...
);
```

```rust
let files: Vec<file::ActiveModel>;

file::Entity::insert_many(files).exec(db).await?;
```
So if anything this is a good workaround for SQLite users.

billy1624:
Hey @baoyachi, I'll convert this to draft due to inactive :)

billy1624:
I'm implementing a simple version of `ON CONFLICT` on #279

PR #173, opened 2021-10-27
Add binary method to expr

### LINKED ISSUE #172: Less than expression or Greater than expression
This is a feature request to add methods on `Expr` for  binary operations  between different expressions. Currently I have to build the binary expressions manually like so:

```rust
SimpleExpr::Binary(
    Box::new(Expr::tbl(Details::Table, Details::Quantity).into()),
    BinOper::SmallerThan,
    Box::new(Expr::tbl(Global::Table, Global::Limit).into()),
)
```

In order to not have a method for every single binary operation that exists, I would suggest a method allowing the following

```Rust
Expr::tbl(Details::Table, Details::Quantity).binary(BinOper::SmallerThan, (Global::Table, Global::Limit))
```

The method could look as follows:

```Rust
impl Expr {
    pub fn binary<T>(self, oper: BinOper, right: T) -> SimpleExpr
    where
        T: Into<SimpleExpr>
}
```
This would assume that something that implements `IntoColumnRef` also implements `Into<SimpleExpr>`.

Progdrasil:
This adds a `binary` method to `Expr` in order to create any sort of binary operation with any type of `SimpleExpr`

## To Discuss

1. This also changes the `Into` implementations  on `Expr` to `From` implementations which will provide the `Into` implementations automatically with the standard library's blanket implementation.
Was there a reason for using `Into` rather than `From`?

2. A blanket implementation for all types implementing `IntoColumnRef` for `SimpleExpr` was added. This is to allow the following call:
```rust
Expr::col(Char::SizeW).binary(BinOper::GreaterThan, Char::SizeH)
```
Otherwise the alternative without the blanket implementation is
```rust
Expr::col(Char::SizeW).binary(BinOper::GreaterThan, Expr::col(Char::SizeH))
```
Which would the maintainers prefer?

3. A `From<Value>` implementation was added to `SimpleExpr` To facilitate value passing, Unfortunately without specialization I cannot do a blanket implementation for all types that implement `Into<Value>` 
```rust
 Expr::col(Char::SizeW).binary(BinOper::SmallerThan, Value::from(10))
```
Otherwise the alternative is:
```rust
Expr::col(Char::SizeW).binary(BinOper::SmallerThan, Expr::value(10))
```
Which would the maintainers prefer?

Closes #172

### DIFF
--- a/src/expr.rs
+++ b/src/expr.rs
@@ -482,6 +482,41 @@ impl Expr {
         self.bin_oper(BinOper::GreaterThan, SimpleExpr::Value(v.into()))
     }
 
+    /// Express a greater than (`>`) expression to another expression.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use sea_query::{tests_cfg::*, *};
+    ///
+    /// let query = Query::select()
+    ///     .columns(vec![Char::Character, Char::SizeW, Char::SizeH])
+    ///     .from(Char::Table)
+    ///     .and_where(
+    ///         Expr::tbl(Char::Table, Char::SizeW)
+    ///         .greater_than(Expr::tbl(Char::Table, Char::SizeH))
+    ///     )
+    ///     .to_owned();
+    ///
+    /// assert_eq!(
+    ///     query.to_string(MysqlQueryBuilder),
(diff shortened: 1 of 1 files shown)


billy1624:
[0.18.2](https://github.com/SeaQL/sea-query/releases/tag/0.18.2)

PR #175, opened 2021-10-29
Helper methods for sea-orm soft delete

billy1624:
- Depended by SeaQL/sea-orm#276


ikrivosheev:
@billy1624 can you fix `clippy warnings`? I think this is very useful.

ikrivosheev:
@billy1624 do you plan to continue working?

tyt2y3:
Actually, I would prefer a non-mut style of API: instead of `fn merge(&mut self, b)`, `fn merge(&self, b) -> Self` would be better.

billy1624:
This is no longer needed in the updated soft delete https://github.com/SeaQL/sea-orm/pull/790
I'll close this

PR #182, opened 2021-11-08
Support `RETURNING` in sea-orm

billy1624:


billy1624:
Close for now. Because we only need `RETURNING` support for Postgres, discussed on SeaQL/sea-orm#292.

PR #186, opened 2021-11-15
removed `'static` and inplace, added the type `IntoTableRef`.

josh-codes:


src/table/create.rs
```diff
@@ -268,7 +268,13 @@ impl TableCreateStatement {
...
 
     pub fn get_table_name(&self) -> Option<String> {
-        self.table.as_ref().map(|table| table.to_string())
+        self.table.as_ref().map(|table| match table {
+            TableRef::Table(table) => table.to_string(),
+            TableRef::SchemaTable(schema, table) => {
+                format!("{}.{}", schema.to_string(), table.to_string())
+            }
+            _ => unimplemented!(),
+        })
```
billy1624, line 277: I guess it's better to return `&TableRef` directly

tyt2y3 (approved):
Thank you for your contribution

PR #194, opened 2021-11-30
Implement `RETURNING` for SQLite

SpyrosRoum:


tyt2y3:
The reason we left out `RETURNING` support for SQLite is because it is a recent feature addition.
Which means we probably would need to do version check somewhere.
But you are right, may be the *caller* should do the check themselves.

Sytten:
Diesel had the same problem. I think one elegant solution would be version flags but it might become hard to maintain.

SpyrosRoum:
I think a disabled by default feature would be the right thing to do.
This way it forces the user to think about it when enabling it.

As for becoming hard to maintain I don't see why that would be the case, can you elaborate?

tyt2y3:
I think a compile-time feature guard is not the best way.
The problem is there is no way for a user to explicitly opt-out.
If two crates in the same workspace happen to depend on sea-query, where `A` enabled this feature but `B` didn't, `B` will still be opt-in as a side effect.

So better just let the caller decides.

SpyrosRoum:
> So better just let the caller decides.

Alright, so how do we do that? 
If we mention it in the docs then doesn't the user decide by just not using it?

tyt2y3:
> Alright, so how do we do that? If we mention it in the docs then doesn't the user decide by just not using it?

Yes, that's exactly what I meant. If the user uses it, then it is their responsibility to make sure the database itself supports it.

SpyrosRoum:
Is there anything else blocking this?
Ideally the sqlx issue would be resolved as well, and I will try doing some investigation myself on it but no promises

tyt2y3 (approved):
Looks good. Thank you so much for the contribution!

PR #195, opened 2021-11-30
Implement ToTokens for IntervalField

autarch:
This is necessary for full support of the Interval type in sea-orm/sea-orm-codegen.

This is a bit of a flailing PR. I'm not sure adding a dependency on proc_macro2 and quote is what you'd want for the main crate in this workspace, but I'm not sure if it makes sense to put this in sea-query-derive instead.

I'm adding this so that the sea-orm tools can support this type. I'm happy to rework this if this is the wrong place for this code.


autarch:
I'm also not sure where to put tests for this.

billy1624:
Hi @autarch, thanks for the contribution!

Are you planning to open a PR on sea-orm as well? You can open it first, then we could go through both of it at together.

tyt2y3:
Well, I don't know much about `IntervalField`, can you explain briefly the use case?
Btw, yes, at least we should put a feature guard for it, probably
```toml
postgres-interval_field = ["proc-macro2", "quote"]
```

autarch:
Sorry for the slow response. I started working on this as part of some deep dive yak shaving for another project I'm working on. But for now I've decided to just go ahead and use `sqlx` for that project so I can move forward on it.

I do have some in-progress work on `sea-schema` for this as well, but nothing for `sea-orm` yet. I got kind of stuck on the other issue I reported for `sea-schema`, https://github.com/SeaQL/sea-schema/issues/28, and that prevented me from really playing with `sea-orm` with the database I wanted to use it for.

tyt2y3:
@billy1624 can help a bit on the feature guard?

billy1624:
> @billy1624 can help a bit on the feature guard?

Added `with-quote` feature
- https://github.com/SeaQL/sea-query/pull/195/commits/5ae4630c361457fe3c63b4b6a3efce98fc8c35a2

billy1624:
@autarch @tyt2y3 any comments?

autarch:
> @autarch @tyt2y3 any comments?

I don't know that the feature name `with-quote` is very clear. It might make sense if there are more types in the future that will need this support. But if this is only for this one type, I think a name like `pg-interval` might be clearer.

autarch:
> I don't know that the feature name `with-quote` is very clear. It might make sense if there are more types in the future that will need this support. But if this is only for this one type, I think a name like `pg-interval` might be clearer.

And following up on that ... it might make sense to rename the type to `PgInterval` too, since it really is entirely Pg-specific, AFAICT.

tyt2y3:
> I don't know that the feature name `with-quote` is very clear. It might make sense if there are more types in the future that will need this support. But if this is only for this one type, I think a name like `pg-interval` might be clearer.

Agreed

> And following up on that ... it might make sense to rename the type to PgInterval too, since it really is entirely Pg-specific, AFAICT.

Agreed

tyt2y3:
Sorry for re-visiting this after so long, but why and where was `ToToken` needed in the first place.
I have a desire to deprecate (or remove) that bit.

PR #202, opened 2021-12-09
Write database & schema prefix in `TableRef` for all databases

billy1624:
## PR Info
- Required by
    - SeaQL/sea-schema#26
    - SeaQL/sea-schema#29


tyt2y3 (approved):
"Let caller decide"

PR #205, opened 2021-12-13
Implemented 'Array' type for Postgres.

kev0960:
This introduces "Array" as a new Value variant.

It can be derived from all Vec<T> types, where T is a valid Value, except for u8, which is already handled by Bytes.

Note that sqlx and postgres-types already supports an array, so this can be implemented without too much hassle.


kev0960:
Hmm, actually, this will not work because we cannot provide bindings to Sqlx. 

Sqlx expects Vec<T> for an array, and we have to somehow convert Vec<Value> to Vec<T>. :(

billy1624:
Hey @kev0960, thanks for the contributions! I assumed you want to support `Vec<T>` in sea-orm?

kev0960:
Yes I do :)

kev0960:
I don't like the idea but we can probably workaround by defining bunch of definitions like

```
ArrayInt32(Option<Vec<i32>>),
ArrayInt64(Option<Vec<i64>>),
ArrayString(Option<Vec<String>>),
...
```

and treat them individually. I really cant think of the "clean" way to convert the generic `Array(Option<Vec<Value>>)` into `Vec<ConcreteType>` since "Value" does not really carry the type we have to convert to. 

Any ideas are welcomed :)

billy1624:
Let me think loll :P

src/value.rs
```diff
@@ -72,6 +72,10 @@ pub enum Value {
     #[cfg(feature = "with-bigdecimal")]
     #[cfg_attr(docsrs, doc(cfg(feature = "with-bigdecimal")))]
     BigDecimal(Option<Box<BigDecimal>>),
+
+    #[cfg(feature = "with-array")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "with-array")))]
+    Array(Option<Box<Vec<Value>>>),
```
psibi, line 78: Do you think a solution like this will work out:

```
pub enum Value<A> {
```

and then this becomes:

```
    Array(Option<Box<Vec<A>>>),
```

We would have to convert the other code to account for the generic argument, but I'm thinking if such a approach will work.

billy1624: > ```
> pub enum Value<A> {
> ```

So, `A` is unbounded? Why is this necessary?

tyt2y3 (approved):
Sorry for the late reply. Actually, this looks very good!

billy1624:
Hey @tyt2y3, I think something is missing in this PR and it shouldn't be merged into master yet.

Binding array values to SQLx PostgreSQL is not ready
- https://github.com/SeaQL/sea-query/runs/4502670225

PR #208, opened 2021-12-16
Add `ColumnSpec::Unsigned`

billy1624:
## PR Info
- Requires by
  - SeaQL/sea-schema#35
  - SeaQL/sea-orm#385


tyt2y3:
What about these then
https://github.com/SeaQL/sea-query/blob/b9876ba378937dd6a3c364f51699382e36e22b33/src/value.rs#L210

billy1624:
> What about these then
> 
> https://github.com/SeaQL/sea-query/blob/b9876ba378937dd6a3c364f51699382e36e22b33/src/value.rs#L210

Here in this PR, we specify UNSIGNED in terms of `ColumnSpec` instead of `ColumnType`. So, there is no way we can map `Value::TinyUnsigned` to the corresponding unsigned `ColumnType`, unless we add all four unsigned integer types in  `ColumnType`.

tyt2y3:
Closed in favour of https://github.com/SeaQL/sea-query/pull/211

Issue #209, opened 2021-12-20
Support `NULLS LAST` and `NULLS FIRST`


qyihua:
Hello,
I found sea-query doesn't support `NULLS LAST` and `NULLS FIRST`  in the `OrderedStatement` yet.
- In PostgreSQL, by default, null values sort as if larger than any non-null value
  ref: https://www.postgresql.org/docs/current/queries-order.html
- SQLite considers NULL values to be smaller than any other values for sorting purposes  
  ref: https://sqlite.org/lang_select.html
- Mysql seems doesn't support the syntax to sort nulls last in.

PR #210, opened 2021-12-20
Support 'NULLS LAST' and 'NULLS FIRST'

qyihua:
Hi,
In PostgreSQL, by default, null values sort as if larger than any non-null value
ref: https://www.postgresql.org/docs/current/queries-order.html
SQLite considers NULL values to be smaller than any other values for sorting purposes
ref: https://sqlite.org/lang_select.html
Mysql seems doesn't support the syntax to sort nulls last in.

#209

### DIFF
--- a/src/backend/postgres/query.rs
+++ b/src/backend/postgres/query.rs
@@ -105,7 +105,7 @@ impl QueryBuilder for PostgresQueryBuilder {
         match order_expr.nulls {
             None => (),
             Some(Nulls::Last) => write!(sql, " NULLS LAST").unwrap(),
-            Some(Nulls::First) => write!(sql, " NULLS FISRT").unwrap(),
+            Some(Nulls::First) => write!(sql, " NULLS FIRST").unwrap(),
         }
     }
 }
--- a/src/backend/postgres/query.rs
+++ b/src/backend/postgres/query.rs
@@ -102,10 +102,10 @@ impl QueryBuilder for PostgresQueryBuilder {
         self.prepare_simple_expr(&order_expr.expr, sql, collector);
         write!(sql, " ").unwrap();
         self.prepare_order(&order_expr.order, sql, collector);
-        match order_expr.nulls_last {
+        match order_expr.nulls {
             None => (),
-            Some(true) => write!(sql, " NULLS LAST").unwrap(),
-            Some(false) => write!(sql, " NULLS FISRT").unwrap(),
+            Some(Nulls::Last) => write!(sql, " NULLS LAST").unwrap(),
+            Some(Nulls::First) => write!(sql, " NULLS FISRT").unwrap(),
         }
     }
 }
--- a/src/backend/postgres/query.rs
+++ b/src/backend/postgres/query.rs
@@ -92,4 +92,20 @@ impl QueryBuilder for PostgresQueryBuilder {
             _ => QueryBuilder::prepare_simple_expr_common(self, simple_expr, sql, collector),
         }
     }
+
+    fn prepare_order_expr(
+        &self,
+        order_expr: &OrderExpr,
+        sql: &mut SqlWriter,
+        collector: &mut dyn FnMut(Value),
+    ) {
+        self.prepare_simple_expr(&order_expr.expr, sql, collector);
+        write!(sql, " ").unwrap();
+        self.prepare_order(&order_expr.order, sql, collector);
+        match order_expr.nulls_last {
+            None => (),
+            Some(true) => write!(sql, " NULLS LAST").unwrap(),
+            Some(false) => write!(sql, " NULLS FISRT").unwrap(),
+        }
+    }
 }
--- a/src/backend/postgres/query.rs
+++ b/src/backend/postgres/query.rs
@@ -86,8 +86,8 @@ impl QueryBuilder for PostgresQueryBuilder {
         self.prepare_order(&order_expr.order, sql, collector);
         match order_expr.nulls {
             None => (),
-            Some(Nulls::Last) => write!(sql, " NULLS LAST").unwrap(),
-            Some(Nulls::First) => write!(sql, " NULLS FIRST").unwrap(),
+            Some(NullOrdering::Last) => write!(sql, " NULLS LAST").unwrap(),
+            Some(NullOrdering::First) => write!(sql, " NULLS FIRST").unwrap(),
         }
     }
 }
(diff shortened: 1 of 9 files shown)


src/types.rs
```diff
@@ -153,6 +153,7 @@ pub enum JoinType {
 pub struct OrderExpr {
     pub(crate) expr: SimpleExpr,
     pub(crate) order: Order,
+    pub(crate) nulls_last: Option<bool>,
```
billy1624, line 156: It would be better to have an enum named `OrderNull` / `NullOrder` or `Nulls` with variants `First` and `Last` to represents it.

src/backend/query_builder.rs
```diff
@@ -637,6 +637,11 @@ pub trait QueryBuilder: QuotedBuilder {
         self.prepare_simple_expr(&order_expr.expr, sql, collector);
         write!(sql, " ").unwrap();
         self.prepare_order(&order_expr.order, sql, collector);
+        match order_expr.nulls_last {
+            Some(true) => write!(sql, " NULLS LAST").unwrap(),
+            Some(false) => write!(sql, " NULLS FISRT").unwrap(),
+            _ => (),
+        }
```
billy1624, line 644: This should be placed inside `src/backend/postgres/query.rs`. So, by default all other backend won't write order by nulls when preparing the select statement.

src/query/ordered.rs
```diff
@@ -95,4 +102,86 @@ pub trait OrderedStatement {
...
+    }
+
+    /// Order by custom string with nulls order option.
+    fn order_by_customs_nulls_last<T>(
+        &mut self,
+        cols: Vec<(T, Order)>,
+        nulls_last: bool,
+    ) -> &mut Self
+    where
+        T: ToString,
```
billy1624, line 157: Shouldn't this `nills_last` be in the `cols` tuple?

src/query/ordered.rs
```diff
@@ -95,4 +102,86 @@ pub trait OrderedStatement {
...
+    }
+
+    /// Order by vector of columns with nulls order option.
+    fn order_by_columns_nulls_last<T>(
+        &mut self,
+        cols: Vec<(T, Order)>,
+        nulls_last: bool,
+    ) -> &mut Self
+    where
+        T: IntoColumnRef,
```
billy1624, line 176: Same here

billy1624 (commented):
Hey @qyihua, thanks for the contribution!

I'm thinking is there any better name for these `order_by_xxx_nulls_last` methods?

qyihua:
@billy1624 Thank you for the good suggestions. 
I need convert the pr to draft and go on

qyihua:
Hi @billy1624 , 
I make some change, it's ready for review now. Thank you!

billy1624:
Btw... just found something interesting. We could actually implement `NULLS FIRST` and `NULLS LAST` for MySQL and MariaDB.

```sql
SELECT ... FROM ... ORDER BY reuestId IS NULL DESC, reuestId DESC
```

_Adapted from https://stackoverflow.com/a/9307657/7059723_

qyihua:
> Btw... just found something interesting. We could actually implement `NULLS FIRST` and `NULLS LAST` for MySQL and MariaDB.
> 
> ```sql
> SELECT ... FROM ... ORDER BY reuestId IS NULL DESC, reuestId DESC
> ```
> 
> _Adapted from https://stackoverflow.com/a/9307657/7059723_

It's great! Let me have a try

billy1624:
Thanks!!

src/backend/postgres/query.rs
```diff
@@ -92,4 +92,20 @@ impl QueryBuilder for PostgresQueryBuilder {
...
+        sql: &mut SqlWriter,
+        collector: &mut dyn FnMut(Value),
+    ) {
+        self.prepare_simple_expr(&order_expr.expr, sql, collector);
+        write!(sql, " ").unwrap();
+        self.prepare_order(&order_expr.order, sql, collector);
+        match order_expr.nulls {
+            None => (),
+            Some(Nulls::Last) => write!(sql, " NULLS LAST").unwrap(),
+            Some(Nulls::First) => write!(sql, " NULLS FISRT").unwrap(),
```
billy1624, line 108: Just discovered there is a typo, "FISRT".
Check the fix https://github.com/SeaQL/sea-query/commit/ce52bda96653b61d07c2cc9ebaf106169727a160

qyihua: Oh! Sorry for the stupid mistakes. 
fixed.

qyihua:
> Btw... just found something interesting. We could actually implement `NULLS FIRST` and `NULLS LAST` for MySQL and MariaDB.
> 
> ```sql
> SELECT ... FROM ... ORDER BY reuestId IS NULL DESC, reuestId DESC
> ```
> 
> _Adapted from https://stackoverflow.com/a/9307657/7059723_

Nulls order support for Mysql backend is ok now :)

billy1624:
I have added some more test cases on https://github.com/billy1624/sea-query/commit/e43245b1bf3713eec30b691b3ef60e140c6af078
You can cherrypick it to this PR

qyihua:
> I have added some more test cases on [billy1624@e43245b](https://github.com/billy1624/sea-query/commit/e43245b1bf3713eec30b691b3ef60e140c6af078) You can cherrypick it to this PR

I have cherrypick it to this PR, and all test is passed

src/types.rs
```diff
@@ -148,11 +148,19 @@ pub enum JoinType {
     RightJoin,
 }
 
+/// Nulls order
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Nulls {
```
tyt2y3, line 153: I think we don't need to pub this ? We can keep this as `pub(crate)`?
If not, we need to think of a better name than `Nulls`.
I am thinking may be `NullOrdering` or something

tyt2y3 (commented):
Sorry for the delay but there is now several conflicts. Can help resolving?

src/types.rs
qyihua: 1. If this keep as pub(crate), how do you think users should use these methods order_by_xxxs_with_nulls?
  I think it need pass bool as nulls order argument ?
2. I also think `NullOrdering` is better than `Nulls`

qyihua:
> Sorry for the delay but there is now several conflicts. Can help resolving?

Yeah, but something is wrong in `src/driver/sqlx_postgres.rs` 
```rust 
} else if value.is_array() {
    query.bind(value.as_ref_array())
```

billy1624:
Let me check

billy1624:
Should be fixed

qyihua:
> Should be fixed

Yes! It's fixed now. Thanks :)

tyt2y3:
Thank you for your contribution!

PR #211, opened 2021-12-22
Add `ColumnType` unsigned integer types

billy1624:
## PR Info
- Requires by
  - SeaQL/sea-schema#37
  - SeaQL/sea-orm#385


tyt2y3 (commented):
Great. But is there any reason for not adding `Unsigned` variants to `Value`?

billy1624:
> Great. But is there any reason for not adding `Unsigned` variants to `Value`?

I think we already have that?

https://github.com/SeaQL/sea-query/blob/79c58c550ee59ced3af182ef3ab52c9131b53f64/src/value.rs#L23-L40

tyt2y3:
Oh really. Awesome!

PR #214, opened 2021-12-30
Add SELECT EXISTS, Fix #118

### LINKED ISSUE #118: Support `EXISTS (subquery)` operator
Hi, I'm trying to do simple query and use `EXISTS (subquery)` operator, but it looks it's not supported yet in `sea-query`.

.e.g:
```sql
select exists(select 1 from accounts where email = $1 limit 1) as "exists!"
```

RomainMazB:
This would add the SELECT EXISTS statement with this api:
```rust
Query::select()
    .exists()
    .from(Char::Table)
    .and_where(Expr::col(Char::Character).eq("A"))
```

I had a doubt about the api to use, my second shot would have been this one, which would have been syntactically more equivalent to the final SQL query but more verbose (let me know if you'd preferred this one):
```rust
Query::exists(
    Query::select()
        .from(Char::Table)
        .and_where(Expr::col(Char::Character).eq("A")),
    Alias::new("exists")
)
```

I added some comments into the code to describe the process I used since it's relatively "desynchronised" with the rest of the `prepare_select_statement` due to one thing:
- The `SELECT EXISTS` statement needs to be executed "as it self", without any table to query in the main expression.

Due to this I couldn't use a function (because if no table is passed, it only results to a `SELECT ` string without any further expression).

### DIFF
--- a/src/backend/query_builder.rs
+++ b/src/backend/query_builder.rs
@@ -58,6 +58,36 @@ pub trait QueryBuilder: QuotedBuilder {
         sql: &mut SqlWriter,
         collector: &mut dyn FnMut(Value),
     ) {
+        if let Some(_) = &select.exists {
+            // Clone current query and clean all eventual select to perform a wildcard select
+            let mut sub_query = select.clone();
+            sub_query
+                .clear_selects()
+                .expr(Expr::expr(SimpleExpr::Custom("*".to_string())))
+                .exists = None;
+
+            // Wrap the subquery into an `EXISTS()` statement
+            let mut sub_sql = SqlWriter::new();
+            write!(sub_sql, " EXISTS (").unwrap();
+            self.prepare_select_statement(&sub_query, &mut sub_sql, collector);
+            write!(sub_sql, ")").unwrap();
+
+            // Return a query aliased to exists
+            let main_query =
+                Query::select()
+                    .expr_as(
+                        Expr::expr(
+                            SimpleExpr::Custom(sub_sql.result())
+                        ),
+                        Alias::new("exists")
+                    )
+                    .take();
+
+            self.prepare_select_statement(&main_query, sql, collector);
+
+            return ()
+        }
+
         write!(sql, "SELECT ").unwrap();
 
         if let Some(distinct) = &select.distinct {
--- a/tests/mysql/query.rs
+++ b/tests/mysql/query.rs
@@ -791,6 +791,19 @@ fn select_48() {
 }
 
 #[test]
+fn select_49() {
+    assert_eq!(
+        Query::select()
+            .exists()
+            .columns(vec![Char::Character])
+            .from(Char::Table)
+            .and_where(Expr::col(Char::Character).eq("A"))
+            .to_string(MysqlQueryBuilder),
+        "SELECT EXISTS (SELECT * FROM `character` WHERE `character` = 'A') AS `exists`"
+    );
+}
+
+#[test]
 #[allow(clippy::approx_constant)]
 fn insert_2() {
     assert_eq!(
@@ -904,3 +917,4 @@ fn delete_1() {
         "DELETE FROM `glyph` WHERE `id` = 1 ORDER BY `id` ASC LIMIT 1"
     );
 }
+ 
--- a/src/query/select.rs
+++ b/src/query/select.rs
@@ -51,6 +51,7 @@ pub struct SelectStatement {
     pub(crate) limit: Option<Value>,
     pub(crate) offset: Option<Value>,
     pub(crate) lock: Option<LockType>,
+    pub(crate) exists: Option<bool>,
 }
 
 /// List of distinct keywords that can be used in select statement
@@ -122,6 +123,7 @@ impl SelectStatement {
             limit: None,
             offset: None,
             lock: None,
+            exists: None,
         }
     }
 
@@ -140,6 +142,7 @@ impl SelectStatement {
             limit: self.limit.take(),
             offset: self.offset.take(),
             lock: self.lock.take(),
+            exists: self.exists.take(),
         }
     }
 
@@ -279,6 +282,12 @@ impl SelectStatement {
     /// Select distinct
     pub fn distinct(&mut self) -> &mut Self {
         self.distinct = Some(SelectDistinct::Distinct);
+        self
+    }
+
+    /// Select exists
+    pub fn exists(&mut self) -> &mut Self {
+        self.exists = Some(true);
         self
     }
 
@@ -1672,3 +1681,4 @@ impl ConditionalStatement for SelectStatement {
         self
     }
 }
+ 
--- a/tests/postgres/query.rs
+++ b/tests/postgres/query.rs
@@ -775,6 +775,19 @@ fn select_48() {
 }
 
 #[test]
+fn select_49() {
+    assert_eq!(
+        Query::select()
+            .exists()
+            .columns(vec![Char::Character])
+            .from(Char::Table)
+            .and_where(Expr::col(Char::Character).eq("A"))
+            .to_string(PostgresQueryBuilder),
+        r#"SELECT EXISTS (SELECT * FROM "character" WHERE "character" = 'A') AS "exists""#
+    );
+}
+
+#[test]
 #[allow(clippy::approx_constant)]
 fn insert_2() {
     assert_eq!(
--- a/tests/sqlite/query.rs
+++ b/tests/sqlite/query.rs
@@ -791,6 +791,19 @@ fn select_48() {
 }
 
 #[test]
+fn select_49() {
+    assert_eq!(
+        Query::select()
+            .exists()
+            .columns(vec![Char::Character])
+            .from(Char::Table)
+            .and_where(Expr::col(Char::Character).eq("A"))
+            .to_string(SqliteQueryBuilder),
+        "SELECT EXISTS (SELECT * FROM `character` WHERE `character` = 'A') AS `exists`"
+    );
+}
+
+#[test]
 #[allow(clippy::approx_constant)]
 fn insert_2() {
     assert_eq!(


src/backend/query_builder.rs
```diff
@@ -58,6 +58,36 @@ pub trait QueryBuilder: QuotedBuilder {
...
+                            SimpleExpr::Custom(sub_sql.result())
+                        ),
+                        Alias::new("exists")
+                    )
+                    .take();
+
+            self.prepare_select_statement(&main_query, sql, collector);
+
+            return ()
+        }
```
billy1624, line 89: This `clone` seems unnecessary and nested calling of `prepare_select_statement` method could be avoided.

We could incorporate the logic of "exists statement" into the body of `prepare_select_statement` method.

The main difference between a normal select and "exists statement" are
1. all select are ignored and removed with an "*"
1. normal select wrapped with "SELECT EXISTS ( ... ) AS xxx"

tests/mysql/query.rs
```diff
@@ -790,6 +790,19 @@ fn select_48() {
...
+    assert_eq!(
+        Query::select()
+            .exists()
+            .columns(vec![Char::Character])
+            .from(Char::Table)
+            .and_where(Expr::col(Char::Character).eq("A"))
+            .to_string(MysqlQueryBuilder),
+        "SELECT EXISTS (SELECT * FROM `character` WHERE `character` = 'A') AS `exists`"
+    );
+}
```
billy1624, line 804: Btw... do you think we should allow user to customize the alias? Let say someone want `SELECT EXISTS ( ... ) AS found`
```sql
SELECT EXISTS (SELECT * FROM `character` WHERE `character` = 'A') AS `found`
```
We could provide a method `exists_as` that takes any `IntoIden` as input.

src/backend/query_builder.rs
RomainMazB: Before pushing this, I was manually writing the `SELECT EXISTS()` statement without cloning the `SelectStatement`, just cleaning the selects and wrapping it result into `String` using `write!` before and after to add `SELECT EXISTS(` and `) AS exists` and then abort with return ().

With this simpler solution we are still forced to make a nested call to `prepare_select_statement` to generate the subquery with all the where clauses and resetting the `exists` value.

Without nesting and aborting, this could lead the end-user into SQL errors like `SELECT EXISTS(SELECT * FROM "characters") ORDER BY "characters"."font_size"`.

tests/mysql/query.rs
RomainMazB: Yes, this could be added.

src/backend/query_builder.rs
billy1624: > Without nesting and aborting, this could lead the end-user into SQL errors like `SELECT EXISTS(SELECT * FROM "characters") ORDER BY "characters"."font_size"`.

Would it be possible to write the SQL like `SELECT EXISTS(SELECT * FROM "characters" ORDER BY "characters"."font_size")`? (where the closing parenthesis moved to the very end of original select)

Or, we could even ignore the order by expression if it's an exists statement.

RomainMazB: The query you provided is valid (inefficient but valid ^^'), with my "nested" implementation, this is already possible because I clean only the columns. This would be the result of this:
```rust
Query::select()
    .exists()
    .from(Char::Table)
    .and_where(Expr::col(Char::Character).eq("A"))
    .order_by(Character::Id, Order::Desc)
```
> Or, we could even ignore the order by expression if it's an exists statement.

Yeah, we could ignore the order by expression if it's an exists statement, but we would also need to ignore all the further preparations too, including the further `prepare_select_statement`, the order by was an exemple, but it's related to all of them, the distinct, the joins, offset, limits and so on. This is why I ended up just in just cleanly aborting.

As an exemple, I removed the aborting and rerun the test without modification, this is the result:
```sql
SELECT EXISTS (SELECT * FROM `character` WHERE `character` = 'A') AS `exists`SELECT FROM `character` WHERE `character` = 'A'
                                                                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
The end section was appended because the current `SelectStatement` still includes table and conditions, we can get rid of the `SELECT` here by moving the block after `write!(sql, "SELECT ").unwrap();` and replacing our `SELECT EXISTS` by just `EXISTS` but we can't get rid of  the rest of the expression with:
- ignoring all blocks after
- or aborting

tests/mysql/query.rs
RomainMazB: thinking further about it, I can't see a use case where this could be needed:
- If the user requires only one exists, no need to alias it
- We can't get multiple `.select().exists()` from one query without nesting (example below)
- If the user requires multiple of them from two different table, this could be achieved with some nesting like:
```rust
let characters_exists =
    !format(
        "({})",
        Query::select()
                .exists()
                .from(Char::Table)
                .and_where(Expr::col(Char::Character).eq("A"))
    );

let glyphes_exists =
    !format(
        "({})",
        Query::select()
                .exists()
                .from(Glyph::Table)
                .and_where(Expr::col(Glyph::Aspect).eq(1.23))
    );
 
Query::select()
    .expr_as(
        Expr::cust(
            characters_exists,
            Alias::new('characters_exists')
        )
    )
    .expr_as(
        Expr::cust(
            glyphes_exists,
            Alias::new('glyphes_exists')
        )
    );
```
which should ends up with [this perfectly valid syntax](http://sqlfiddle.com/#!9/d2bf79/4/0):
```sql
SELECT
    (SELECT EXISTS (SELECT * FROM `character` WHERE `character` = 'A') AS `exists`) as `characters_exists`
    (SELECT EXISTS (SELECT * FROM `glyph` WHERE `aspect` = 1.23) AS `exists`) as `glyphes_exists`
```

billy1624: That make sense! tricky loll

src/backend/query_builder.rs
billy1624: I was thinking to refactor it into https://github.com/billy1624/sea-query/commit/aaf140e2f1e9395b5a68a9b2aa4b02801bbc579b to avoid `clone` and early exit

tests/mysql/query.rs
RomainMazB: Really tricky yep 👍🏻

Now that it has been developed, we could add a `multiple_exists` method which would accept a `Vec<(SelectStatement, Iden)>` to simplify this use case and avoid the ugly `format!()` to wrap the expressions into parentheses.

src/backend/query_builder.rs
RomainMazB: Hmm, this seems cleaner indeed 😄

billy1624: You can cherrypick the useful bits loll

RomainMazB:
@maintainers
An additional note on this one: if both [exists unary operator](https://github.com/SeaQL/sea-query/issues/216) and [wildcard](https://github.com/SeaQL/sea-query/issues/217) issues are fixed (I'll submit the PR for the wildcard and the draft-PR for exists unary operator (can't finish because it rely on wildcard) in few hours), we could refactor this one or even just not merge this at all because it would be now possible to do something like:
```rust
Query::select()
    .expr_as(
        Expr::exists(
            Query::select()
                .wildcard()
                .from(Char::Table)
        ),
        Alias::new('exists');
    )
```
Which is not so ugly imo, it should results to something like we wanted:
```sql
SELECT EXISTS(SELECT * FROM `character`) as `exists`
```
Plus! We could now chain multi-select-exists it more elegantly:
```rust
Query::select()
    .expr_as(
        Expr::exists(
            Query::select()
                .wildcard()
                .from(Char::Table)
        ),
        Alias::new('character_exists');
    )
    .expr_as(
        Expr::exists(
            Query::select()
                .wildcard()
                .from(Glyph::Table)
        ),
        Alias::new('glyph_exists');
    )
```
```sql
SELECT EXISTS(SELECT * FROM `character`) as `character_exists`, EXISTS(SELECT * FROM `glyph`) as `glyph_exists`
```

billy1624:
> ```rust
> Query::select()
>     .expr_as(
>         Expr::exists(
>             Query::select()
>                 .wildcard()
>                 .from(Char::Table)
>         ),
>         Alias::new('exists');
>     )
> ```

It looks elegant!

RomainMazB:
The two related PRs are ready for reviews, I made the `.wildcard()` call unnecessary to call in the `exists` `SelectStatement`, it will be automatically injected.

Feel free to not merge this one if both of the other two get merged.

src/backend/query_builder.rs
```diff
@@ -58,21 +58,27 @@ pub trait QueryBuilder: QuotedBuilder {
...
-            self.prepare_select_distinct(distinct, sql, collector);
-            write!(sql, " ").unwrap();
-        }
+        write!(sql, "SELECT ").unwrap();
 
-        select.selects.iter().fold(true, |first, expr| {
-            if !first {
-                write!(sql, ", ").unwrap()
+        if is_exists_stmt {
+            write!(sql, "EXISTS (SELECT * ").unwrap();
```
tyt2y3, line 66: First of all thank you! I liked the API overall. However, seeing the code generation part here makes me feel like something is missing. We definitely need to fully express this concept in the syntax tree.
We could probably expand the definition of `SimpleExpr::SubQuery`

```rust
SelectStatement {
    selects: [
        SelectExpr {
            expr: SimpleExpr::SubQuery(
                SubQueryOper::Exists,
                SelectStatement{ ... }
            ),
            alias: None
        }
    ]
    ...
}
```

tyt2y3:
Oh and if we place the test cases as doc tests, we can save some effort by not having to put the test in 3 different places.

ikrivosheev:
Close PR, continue: https://github.com/SeaQL/sea-query/pull/379

PR #218, opened 2021-12-30
Add `EXISTS` unary operator, fix #216

### LINKED ISSUE #216: Add WHERE EXISTS condition
Now that `SELECT EXISTS` may be about to get merged, the `EXISTS` as condition method should be supported too.

The `WHERE EXISTS` clause is really useful on relationship constraints:

As an example to retrieve all the game's character owned by a user with name starting by `bo`:
```sql
SELECT * FROM `characters`
WHERE EXISTS(
    SELECT * FROM `users`
    WHERE `users`.`name` LIKE "bo%"
        AND  `users`.`id` = `characters`.` user_id`
)
```

This function would accept the subquery as parameter and just add the `EXISTS`  `BinOp`, the same way `in_subquery` does.

Could tackle this tonight if this seems fine to everyone.

RomainMazB:
First draft to fix #216 , the next step would be to handle the wildcard to constraint a `SELECT *` without any further columns.

### DIFF
--- a/src/backend/query_builder.rs
+++ b/src/backend/query_builder.rs
@@ -342,6 +342,14 @@ pub trait QueryBuilder: QuotedBuilder {
             SimpleExpr::AsEnum(_, expr) => {
                 self.prepare_simple_expr(expr, sql, collector);
             }
+            SimpleExpr::UnarySelect(op, select) => {
+                self.prepare_un_oper(op, sql, collector);
+                write!(sql, " (").unwrap();
+                // TODO: When/If Expr::wildcard() is available:
+                //   clean the columns and add the wildcard to express only a SELECT *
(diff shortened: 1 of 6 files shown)


RomainMazB:
This PR is ok for review, but rely on #219 , so it should be merged before this one.

There will be some conflict inside the tests, I renamed the methods name so you can just manually merge the tests in the good order.

tyt2y3:
As discussed in another PR, I think we don't need to introduce a new `UnarySelect` variant, instead we can extend the existing `SubQuery` variant to include one more enum probably `SubQueryOper` which could itself be `Exists` or `None`

ikrivosheev:
Close PR, continue: https://github.com/SeaQL/sea-query/pull/379

PR #219, opened 2021-12-30
Add Expr::asterisk() and Expr::tbl_asterisk(table: DynIden) methods - Fix #217

### LINKED ISSUE #217: Add Expr::asterisk() and Expr::tbl_asterisk(IntoTableRef)
In most of the case, we don't need to list all the columns of the table, which could be tedious with some large tables.

We should be able to pass an `Expr::wildcard()` or `Expr::wildcard(IntoTableRef)` to the columns' list, which would result in an asterisk or a table name followed by an asterisk:
```rust
// SELECT * FROM `character`;
Query::select()
    .expr(Expr::wildcard())
    .from(Char::Table);

// SELECT `character`.*, `glyph`.`aspect`
// FROM `character`
// JOIN `glyph` ON ....;
Query::select()
    .expr(Expr::tbl_wildcard(Char::Table))
    .columns(Glyph::Table, Glyph::Aspect)
    .from(Char::Table)
    .join(Glyph::Table, /*...*/);
```

This could also open some other doors of features like the one I stuck on here: https://github.com/SeaQL/sea-query/issues/216#issuecomment-1003102381

RomainMazB:
This fix #217, adding two new methods:
- `Expr::wildcard()`
```rust
Query::select()
    .expr(Expr::wildcard())
    .from(Char::Table)
    .to_string(MysqlQueryBuilder);
```
```sql
SELECT *
FROM `character`
```
- `Expr::tbl_wildcard(table: DynIden)`
```rust
Query::select()
    .expr(Expr::tbl_wildcard(Char::Table))
    .column((Font::Table, Font::Name))
    .from(Char::Table)
    .inner_join(Font::Table, Expr::tbl(Char::Table, Char::FontId).equals(Font::Table, Font::Id))
    .to_string(MysqlQueryBuilder);
```
```sql
SELECT `character`.*, `font`.`name`
FROM `character`
INNER JOIN `font` ON `character`.`font_id` = `font`.`id`
```

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -63,6 +63,8 @@ pub enum ColumnRef {
     Column(DynIden),
     TableColumn(DynIden, DynIden),
     SchemaTableColumn(DynIden, DynIden, DynIden),
+    Wildcard,
+    TableWildcard(DynIden)
 }
 
 pub trait IntoColumnRef {
--- a/src/types.rs
+++ b/src/types.rs
@@ -63,8 +63,8 @@ pub enum ColumnRef {
     Column(DynIden),
     TableColumn(DynIden, DynIden),
     SchemaTableColumn(DynIden, DynIden, DynIden),
-    Wildcard,
-    TableWildcard(DynIden)
+    Asterisk,
+    TableAsterisk(DynIden)
 }
 
 pub trait IntoColumnRef {
--- a/src/backend/query_builder.rs
+++ b/src/backend/query_builder.rs
@@ -251,10 +251,10 @@ pub trait QueryBuilder: QuotedBuilder {
                         write!(sql, ".").unwrap();
                         column.prepare(sql, self.quote());
                     }
-                    ColumnRef::Wildcard => {
+                    ColumnRef::Asterisk => {
                         write!(sql, "*").unwrap();
                     }
-                    ColumnRef::TableWildcard(table) => {
+                    ColumnRef::TableAsterisk(table) => {
                         table.prepare(sql, self.quote());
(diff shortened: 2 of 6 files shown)


src/types.rs
```diff
@@ -63,6 +63,8 @@ pub enum ColumnRef {
     Column(DynIden),
     TableColumn(DynIden, DynIden),
     SchemaTableColumn(DynIden, DynIden, DynIden),
+    Wildcard,
+    TableWildcard(DynIden)
```
tyt2y3, line 67: @billy1624 how will this affect SeaORM / SeaSchema?

tyt2y3:
Sorry for the name argument, but `*` has not got an official name in SQL, but I am pretty sure it is not called 'wildcard' (which is `%`), so perhaps it could be named `Star` or `Asterisk`?

RomainMazB:
Yeah, no problem for the name: I call `*` wildcard because this is how it's called in filesystem paths libraries like [globset](https://docs.rs/globset/latest/globset/#example-configuring-a-glob-matcher).

I'll rename all the appellation `wildcard` to `asterisk` because even if `star` is shorter, I feel uncomfortable with `tbl_star()` method

tyt2y3 (approved):
Thank you

src/types.rs
billy1624: No impact for SeaSchema. But this part of code has to be updated in SeaORM.
https://github.com/SeaQL/sea-orm/blob/82bad4a376efb875ac32ab601ac2cd9cce191e3f/src/query/combine.rs#L45-L62

PR #221, opened 2022-01-03
Use double quotes for quuoting identifgiers for SQLite

### LINKED ISSUE #220: Why are why you using back-ticks for sqlite identifiers instead of double quotes?
If you switch to double quotes then you avoid this sqlx [issue](https://github.com/launchbadge/sqlx/issues/1531), which would allow `RETURNING` to work for SQLite for every backend (plus you are more correct according to the SQL standard))
[SQLite keywords](https://www.sqlite.org/lang_keywords.html)

SpyrosRoum:
This should allow `RETURNING` to work for SQLite with sqlx 
closes #220

### DIFF
--- a/src/backend/mysql/foreign_key.rs
+++ b/src/backend/mysql/foreign_key.rs
@@ -8,7 +8,7 @@ impl ForeignKeyBuilder for MysqlQueryBuilder {
     ) {
         write!(sql, "ALTER TABLE ").unwrap();
         if let Some(table) = &drop.table {
-            table.prepare(sql, '`');
+            table.prepare(sql, self.quote());
         }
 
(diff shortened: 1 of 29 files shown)


SpyrosRoum:
I didn't realize I had to use the `--all-features` flag when testing so I kept missing some tests until the ci told me.
It should be good now

tyt2y3:
@billy1624 Can take a look at the failed cases?

billy1624:
> @billy1624 Can take a look at the failed tests?

I reverted the binding of array values https://github.com/SeaQL/sea-query/commit/67a8ffba1351c754aae2559051930d4fe2fe70ae

tyt2y3:
Awesome, thank you for the diligence!

PR #222, opened 2022-01-06
Support the use of chrono::DateTime<Utc> using the type alias DateTim…

charleschege:
Support DateTime<Utc> natively.

After researching and diving into the code, I found that the best way is to convert the `DateTime<Utc>` into a `DateTime<FixedOffset>` with the `FixedOffset` being based on `chrono::Utc` . This is helpful especially in databases like MySQL which store the `TIMESTAMP` in UTC despite a timezone being specified.

### DIFF
--- a/src/backend/query_builder.rs
+++ b/src/backend/query_builder.rs
@@ -748,6 +748,8 @@ pub trait QueryBuilder: QuotedBuilder {
             Value::DateTime(None) => write!(s, "NULL").unwrap(),
             #[cfg(feature = "with-chrono")]
             Value::DateTimeWithTimeZone(None) => write!(s, "NULL").unwrap(),
+            #[cfg(feature = "with-chrono")]
+            Value::DateTimeUtc(None) => write!(s, "NULL").unwrap(),
             #[cfg(feature = "with-rust_decimal")]
             Value::Decimal(None) => write!(s, "NULL").unwrap(),
             #[cfg(feature = "with-bigdecimal")]
@@ -787,6 +789,10 @@ pub trait QueryBuilder: QuotedBuilder {
             #[cfg(feature = "with-chrono")]
             Value::DateTimeWithTimeZone(Some(v)) => {
                 write!(s, "\'{}\'", v.format("%Y-%m-%d %H:%M:%S %:z").to_string()).unwrap()
+            }
+            #[cfg(feature = "with-chrono")]
+            Value::DateTimeUtc(Some(v)) => {
+                write!(s, "\'{}\'", v.format("%Y-%m-%d %H:%M:%S %z").to_string()).unwrap()
             }
             #[cfg(feature = "with-rust_decimal")]
             Value::Decimal(Some(v)) => write!(s, "{}", v).unwrap(),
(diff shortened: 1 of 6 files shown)


charleschege:
@billy1624 @tyt2y3 Any thoughts on this are welcome

src/backend/query_builder.rs
```diff
@@ -788,6 +790,10 @@ pub trait QueryBuilder: QuotedBuilder {
             Value::DateTimeWithTimeZone(Some(v)) => {
                 write!(s, "\'{}\'", v.format("%Y-%m-%d %H:%M:%S %:z").to_string()).unwrap()
             }
+            #[cfg(feature = "with-chrono")]
+            Value::DateTimeUtc(Some(v)) => {
+                write!(s, "\'{}\'", v.format("%Y-%m-%d %H:%M:%S %z").to_string()).unwrap()
+            }
```
billy1624, line 796: Wouldn't it be better to state the timezone in the format that same as `Value::DateTimeWithTimeZone`? i.e. `+00:00`

REF: https://docs.rs/chrono/latest/chrono/format/strftime/index.html

charleschege: Sure. I will make the changes

src/value.rs
```diff
@@ -1275,6 +1305,17 @@ mod tests {
...
+    #[test]
+    #[cfg(feature = "with-chrono")]
+    fn test_chrono_utc_value() {
+        use chrono::{DateTime, NaiveDateTime, Utc};
+
+        let timestamp = DateTime::<Utc>::from_utc(NaiveDateTime::from_timestamp(61, 0), Utc);
+        let value: Value = timestamp.into();
+        let out: DateTime<FixedOffset> = value.unwrap();
+        assert_eq!(out, timestamp);
+    }
```
billy1624, line 1317: This is unexpected. `DateTime<Utc>` should now be store as `Value::DateTimeUtc` but instead it's `Value::DateTimeWithTimeZone`

```rust
let out: DateTime<FixedOffset> = value.unwrap();
```

Because the following code convert all `DateTime<T>` into `Value::DateTimeWithTimeZone`

https://github.com/SeaQL/sea-query/blob/67a8ffba1351c754aae2559051930d4fe2fe70ae/src/value.rs#L295-L303

billy1624:
Btw... we have to update `src/driver/*.rs` to bind the new `Value::DateTimeUtc` correctly

src/value.rs
charleschege: Thanks for this observation, I had spent hours trying to find this piece of code

billy1624: Hahaaa :P

charleschege:
@billy1624 I have moved the `DateTime<FixedOffset>` into its own `From` trait. I hope that's okay. Also, I am wondering about the correctness of the `src/drivers` code I have introduced. Comments are welcome

src/value.rs
```diff
@@ -555,6 +593,18 @@ impl Value {
...
         panic!("not Value::DateTimeWithTimeZone")
     }
+
+    #[cfg(feature = "with-chrono")]
+    pub fn as_ref_date_time_utc(&self) -> Option<String> {
+        match self {
+            Self::DateTimeUtc(v) => v.as_ref().map(|v| v.naive_utc().to_string()),
+            _ => panic!("not Value::DateTimeUtc"),
+        }
+    }
```
billy1624, line 603: Here should return `Option<&DateTime<Utc>>` instead of serializing it as string

PR #223, opened 2022-01-06
Insert Default

billy1624:
## PR Info
- Required by SeaQL/sea-orm#432
- Related to SeaQL/sea-orm#420
- REF
  - MySQL: https://dev.mysql.com/doc/refman/8.0/en/insert.html
  - PostgreSQL: https://www.postgresql.org/docs/9.1/dml-insert.html
  - SQLite: https://www.sqlite.org/lang_insert.html

### DIFF
--- a/src/backend/mysql/query.rs
+++ b/src/backend/mysql/query.rs
@@ -8,4 +8,8 @@ impl QueryBuilder for MysqlQueryBuilder {
         _collector: &mut dyn FnMut(Value),
     ) {
     }
+
+    fn insert_default_keyword(&self) -> &str {
+        "VALUES (DEFAULT)"
+    }
 }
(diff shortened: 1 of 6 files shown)


tyt2y3:
I totally think that the above API behaviour is too smart, the following should be better.

```rust
assert_eq!(
    Query::insert()
        .into_table(Glyph::Table)
        .or_default_values()
    "INSERT INTO \"glyph\" DEFAULT VALUES"
);
assert_eq!(
    Query::insert()
        .into_table(Glyph::Table)
        .columns(vec![Glyph::Image])
        .values_panic(vec!["ABC".into()])
    "INSERT INTO \"glyph\" (\"image\") VALUES ('ABC')"
);
```

Issue #201, opened 2021-12-03
No Left Join Lateral


rex-remind101:
Hi, I've been looking for a way to perform left join lateral for Postgres but I haven't found a way to do this. Is it available?

billy1624:
Hey @rex-remind101, thanks for the question! Currently, it's not supported :(

rex-remind101:
Is there a way to apply raw SQL? I've tried `.expr(Expr::cust())` but it seems incorrectly place the text for the join in the select statement and appends a `,` as if it was a column.

billy1624:
> Is there a way to apply raw SQL? I've tried `.expr(Expr::cust())` but it seems incorrectly place the text for the join in the select statement and appends a `,` as if it was a column.

Yes, I think we need to introduce a new join operator to support join lateral. Would you be interested in creating a PR? I could help if you want :P

rex-remind101:
Sure thing 👍

rex-remind101:
think i'll actually tackle this now. i saw you got my other issue fixed (thank you). do you have any pointers on where to start?

billy1624:
I have never use the `LATERAL` keyword. Can you explain what select statement you want to construct?

From my understanding, both MySQL and PostgreSQL support `LATERAL` but not SQLite.
- MySQL: https://dev.mysql.com/doc/refman/8.0/en/lateral-derived-tables.html
- PostgreSQL: https://www.postgresql.org/docs/14/sql-select.html

Some sample select statements with `LATERAL`:

```sql
SELECT *
FROM table_1,
     LATERAL (SELECT * FROM tbale_2) as t2
         LEFT JOIN LATERAL (SELECT * FROM tbale_3) as t3 on 1 = 1
         LEFT JOIN LATERAL (SELECT * FROM tbale_4) as t4 on 1 = 1
WHERE 1 = 1
```

rex-remind101:
i use `[left] join lateral` specifically to pull columns from an outer query into an inner subquery and then join. it's simpler than a strict subquery

e.g.

```sql
select
  t1.name,
  descendants.count IS NULL as is_leaf,
from t1
left join lateral (
  select ancestor, count(1) as count from t2
  where ancestor = t1.uuid group by 1
) as descendants on descendants.ancestor = t1.uuid
```

billy1624:
So, basically we need to write the `LATERAL` keyword to the join expression.

[...]

PR #224, opened 2022-01-07
[join-lateral] 

rex-remind101:
added support for `JOIN LATERAL` as specified in MySQL and Postgres https://github.com/SeaQL/sea-query/issues/201

### DIFF
--- a/src/query/select.rs
+++ b/src/query/select.rs
@@ -74,6 +74,7 @@ pub struct JoinExpr {
     pub join: JoinType,
     pub table: Box<TableRef>,
     pub on: Option<JoinOn>,
+    pub lateral: bool,
 }
 
 /// List of lock types that can be used in select statement
@@ -902,6 +903,7 @@ impl SelectStatement {
             JoinOn::Condition(Box::new(ConditionHolder::new_with_condition(
                 condition.into_condition(),
             ))),
+            false,
         )
     }
 
@@ -973,6 +975,7 @@ impl SelectStatement {
             JoinOn::Condition(Box::new(ConditionHolder::new_with_condition(
                 condition.into_condition(),
             ))),
+            false,
         )
     }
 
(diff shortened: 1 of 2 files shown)


src/query/select.rs
```diff
@@ -1061,14 +1064,86 @@ impl SelectStatement {
...
+    ///     r#"SELECT `name` FROM `font` LEFT JOIN LATERAL (SELECT `id` FROM `glyph`) AS `sub_glyph` ON `font`.`id` = `sub_glyph`.`id`"#
+    /// );
+    /// assert_eq!(
+    ///     query.to_string(PostgresQueryBuilder),
+    ///     r#"SELECT "name" FROM "font" LEFT JOIN LATERAL (SELECT "id" FROM "glyph") AS "sub_glyph" ON "font"."id" = "sub_glyph"."id""#
+    /// );
+    /// assert_eq!(
+    ///     query.to_string(SqliteQueryBuilder),
+    ///     r#"SELECT `name` FROM `font` LEFT JOIN LATERAL (SELECT `id` FROM `glyph`) AS `sub_glyph` ON `font`.`id` = `sub_glyph`.`id`"#
+    /// );
```
billy1624, line 1101: JOIN `LATERAL` seems not supported by SQLite. We shouldn't generate invalid select statement

src/query/select.rs
```diff
@@ -1061,14 +1064,86 @@ impl SelectStatement {
...
+    pub fn join_lateral_subquery<T, C>(
+        &mut self,
+        join: JoinType,
+        query: SelectStatement,
+        alias: T,
+        condition: C,
+    ) -> &mut Self
+    where
+        T: IntoIden,
+        C: IntoCondition,
```
billy1624, line 1129: I'm thinking this method could simply be named `join_lateral`. As it's assumed to take a (sub)query as argument. What do you think?

src/query/select.rs
rex-remind101: how do i error on sqlite within the builder?

src/query/select.rs
rex-remind101: yup, that makes sense 👍

src/query/select.rs
billy1624: We could `panic!` when building the select statement for SQLite. Is this be reasonable? @tyt2y3 It's better to let the user know explicitly something went wrong early.

tyt2y3:
I think the `lateral` should be made part of `JoinType` though.

rex-remind101:
@billy1624 hello, any thoughts?

billy1624:
> I noticed this was merged into it's own branch. Has any further work been done on it? Should I make another pr with those changes or is there a different direction we're heading in?

It was later cherry picked into master, see https://github.com/SeaQL/sea-query/commit/12c1292c2e890c06527d9a2a3fd7b3a4d8a24508

rex-remind101:
thank you, do i need to follow up with the above adjustments you suggested? if so, i'm still not sure where to insert the `panic!` for slqite, thanks

billy1624:
> thank you, do i need to follow up with the above adjustments you suggested? if so, i'm still not sure where to insert the `panic!` for slqite, thanks

I can open a PR for that :P

billy1624:
> thank you, do i need to follow up with the above adjustments you suggested? if so, i'm still not sure where to insert the `panic!` for slqite, thanks

Hey @rex-remind101, I have open an PR (#243) for this

PR #225, opened 2022-01-09
Bind array values in SQLx PostgreSQL driver

billy1624:
## PR Info
- Continue #205
- Fixing SeaQL/sea-query@67a8ffba1351c754aae2559051930d4fe2fe70ae

### DIFF
--- a/examples/sqlx_postgres/Cargo.toml
+++ b/examples/sqlx_postgres/Cargo.toml
@@ -17,9 +17,17 @@ sea-query = { path = "../../", features = [
     "with-rust_decimal",
     "with-bigdecimal",
     "with-uuid",
+    "postgres-array",
 ] }
 # NOTE: if you are copying this example into your own project, use the following line instead:
-# sea-query = { version = "^0", features = [...] }
+# sea-query = { version = "^0", features = [
+#     "sqlx-postgres",
+#     "with-chrono",
+#     "with-json",
+#     "with-rust_decimal",
+#     "with-bigdecimal",
+#     "with-uuid",
+# ] }
 
 [dependencies.sqlx]
 version = "^0.5"
--- a/src/driver/sqlx_postgres.rs
+++ b/src/driver/sqlx_postgres.rs
@@ -51,7 +51,7 @@ macro_rules! bind_params_sqlx_postgres {
                     } else if value.is_uuid() {
                         query.bind(value.as_ref_uuid())
                     } else if value.is_array() {
-                        unimplemented!("SQLx array is not supported");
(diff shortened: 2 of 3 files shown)


tyt2y3:
Ah, so we still have to clone the values :(
But yea, it's so elaborate

billy1624:
> Ah, so we still have to clone the values :( But yea, it's so elaborate

I can't think of a better way of doing it... given the defined `Value::array`

tyt2y3:
It would be slightly if we introduce a separate function that takes ownership of Values

billy1624:
> It would be slightly better if we introduce a separate function that takes ownership of Values

When binding the `value` onto SQLx query? Currently, it's taking `value` as reference.

tyt2y3:
I think this will need to updated after merging https://github.com/SeaQL/sea-query/pull/267
It seems like a considerable maintenance cost in one sense

itsbalamurali:
@billy1624 @tyt2y3 since #267 is merged. Can this be rebased and merged?

ammongit:
Exciting! Glad to see array work progress.

ikrivosheev:
@billy1624 Do you need help with this PR?

ammongit:
Is there anything blocking this PR from being merged?

tyt2y3:
Apart from the merge conflict, I'd appreciate if we can figure out a less verbose approach at supporting this.
The problem is when we add a new type variant to `Value`, it seems like that a lot has to be changed, and we have no compile-time check / CI that can catch those.
Basically a maintainability problem.

negezor:
@billy1624 Is change needed here?
https://github.com/SeaQL/sea-query/blob/master/sea-query-driver/src/sqlx_postgres.rs#L59-L63

negezor:
@billy1624 Is change needed here?
https://github.com/SeaQL/sea-query/blob/master/sea-query-driver/src/sqlx_postgres.rs#L59-L63

kudlatyamroth:
maybe it is better to implement this in that form, and in future refactor it?
i think that many ppl needs this functionality

Sytten:
Indeed I am in need of it so anything I can do to help let me know.

billy1624:
Hey @Sytten @tyt2y3, I think we could dump this PR and start working on this feature from master, i.e. based on `sea-query-binder`.

billy1624:
Would you like to take on this? @Sytten

Sytten:
Yeah will do

billy1624:
Thanks! @Sytten
Do let me know if you need help :D

PR #237, opened 2022-01-14
Support for (Pg)Interval with chrono::Duration

Rudi3:
Preparation for
https://github.com/SeaQL/sea-orm/issues/426

**Support for**
- [x] `std::time::Duration`
- [x] `chrono::Duration`

I'm not sure if it makes sense to use `std::time::Duration` without `chrono`

`postgres-types` doesn't seem to support `ToSql` for `chrono::Duration` - at least it wasn't listed in the docs. So I'm not sure if this is a big deal in `postgres.rs`

`sqlx` will panic if a conversion fails; if:

- Duration is longer than what `PgInterval` can store
- Duration contains nanoseconds - too small for `PgInterval`


Rudi3:
For the error handling: what happens when `as_ref_interval` returns None? Will it continue silently?
Should it `panic!` when a conversion fails? That is likely too harsh...

Rudi3:
It seems that `sqlx` actually supports `chrono::Duration` already https://github.com/launchbadge/sqlx/blob/12d9f54c7d37b0f23ec164bad257447e8797ad41/sqlx-core/src/postgres/types/interval.rs#L147
`postgres-types` however doesn't seem to (`ToSql` for `chrono::Duration`) - at least it wasn't listed in the docs. So I'm not sure if this is a big deal in `postgres.rs`
So I might have to rework this to use the `sqlx` conversion instead to remove the redundant checks (and for better error handling).

ikrivosheev:
@Rudi3 hello, are there any plans to finish?)

Rudi3:
@ikrivosheev hello, i have to admit that I forgot about it.
But this was pretty much done except for the uncertainty about `ToSql`.
Do you think there might be any other problems?

ikrivosheev:
@Rudi3 which uncertainty with `ToSql`? I dont understand

Rudi3:
@ikrivosheev [src/driver/postgres.rs](https://github.com/Rudi3/sea-query/blob/3a21d56591dfa7ef8ac8e0e3163d4401e9c6dca2/src/driver/postgres.rs#L65) I decided to throw unimplemented if `to_sql` is called in the postgres driver.rs because postgres-types doesn't implement `ToSql` - as far as I could tell. I'm not sure if that would lead to any problems. I'll have to check if support was added in the meantime, however.
At a glance, it doesn't seem like it: https://docs.rs/postgres-types/latest/postgres_types/trait.ToSql.html
If this is needed, it might block this for now.

ikrivosheev:
Ah... I understood! We can create issue in postgres-types and wait for implantation or implement it yourself. 

I will create issue.

ikrivosheev:
Find: https://github.com/sfackler/rust-postgres/issues/60

billy1624:
Hey @Rudi3, welcome back!

After second thought, I think we can support `time::Duration` & `chrono::Duration`. Just like what we did to all other datetime values.

- https://github.com/SeaQL/sea-query/blob/70cd3dcf4ed26334cc349c1ba419e3f22c5f66d8/src/value.rs#L51-L53
- https://github.com/SeaQL/sea-query/blob/70cd3dcf4ed26334cc349c1ba419e3f22c5f66d8/src/value.rs#L75-L77

Rudi3:
Hey, @billy1624!
Unfortunately `ToSql` isn't implemented for `time::Duration` in `postgres-types` either.
But I'll see if I can add `time::Duration` at least in here when I get to it.

etsea117:
@Rudi3 any update on this?

billy1624:
Hey @Rudi3, I'm sorry for the delay. 

> Hey, @billy1624! Unfortunately `ToSql` isn't implemented for `time::Duration` in `postgres-types` either. But I'll see if I can add `time::Duration` at least in here when I get to it.

I think [`postgres`](https://crates.io/crates/postgres) driver implements non of `chrono::Duration`, `time::Duration` and `std::time::Duration`?

We could get this PR merge before moving on to https://github.com/SeaQL/sea-orm/pull/456

hongquan:
There is a crate, [chrono-intervals](https://crates.io/crates/chrono-intervals) that provide `Interval` type.

Rudi3:
Hello, I revised this PR and noticed that `to_sql` seems be run on a Json value rather than on e.g. `chrono::DateTime` or `chrono::Duration`. So it's converted to a string in a previous step. If I got that right, we don't actually need the `ToSql` Trait for `chrono::Duration` or `time::Duration` - but just a conversion to a precise string that is accepted by postgres.

For now, I opted for the ISO8601 duration format that's provided by `chrono::Duration`. Negative durations are expressed by a minus before the statement like `-P3DT4H5M6S`. But postgres docs mention that each [field should be negative to avoid misinterpretation](https://www.postgresql.org/docs/current/datatype-datetime.html#DATATYPE-INTERVAL-INPUT).

So maybe I should rewrite the formatter to do exactly this. Furthermore, the maximum allowed precision is 6. So should any other digits be rounded/cut off/etc? Requesting some feedback on this and maybe confirmation of my assumption about the `to_sql` conversion, please. Thank you!

DevSlashRichie:
Is there any advancement on this? Is there anything I can help with?

ccqpein:
Hi guys, is there any update about this PR?

firebladed:
is this going anywhere?

Issue #231, opened 2022-01-13
Support Insert from Select


Sytten:
This is a very common syntax that is currently not supported:
```sql
INSERT INTO table SELECT ...;
```
It don't expect it is very hard to support, willing to help on it.

billy1624:
I think this would be easy to implement.

Just want to put some references down here:
- SQLite: https://www.sqlite.org/lang_insert.html
- MySQL: https://dev.mysql.com/doc/refman/8.0/en/insert-select.html
- PostgreSQL: https://www.postgresql.org/docs/current/sql-insert.html

05storm26:
I am working on this.

PR #238, opened 2022-01-14
Insert from select

05storm26:
Fixes: #231

### DIFF
--- a/src/query/insert.rs
+++ b/src/query/insert.rs
@@ -3,6 +3,16 @@ use crate::{
     QueryStatementBuilder, SelectExpr, SelectStatement, SimpleExpr,
 };
 
+/// Represents a value source that can be used in an insert query.
+///
+/// [`InsertValueSource`] is a node in the expression tree and can represent a raw value set
+/// ('VALUES') or a select query.
+#[derive(Debug, Clone)]
+pub enum InsertValueSource {
+    Values(Vec<Vec<SimpleExpr>>),
+    Select(Box<SelectStatement>),
+}
+
 /// Insert any new rows into an existing table
 ///
 /// # Examples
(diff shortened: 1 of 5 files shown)


src/query/insert.rs
```diff
@@ -3,6 +3,16 @@ use crate::{
...
 
+/// Represents a value source that can be used in an insert query.
+///
+/// [`InsertValueSource`] is a node in the expression tree and can represent a raw value set
+/// ('VALUES') or a select query.
+#[derive(Debug, Clone)]
+pub enum InsertValueSource {
+    VALUES(Vec<Vec<SimpleExpr>>),
+    QUERY(SelectStatement),
+}
```
billy1624, line 14: I think it's better to `Box` the `SelectStatement`, i.e. `Box<SelectStatement>`.
CC @tyt2y3

05storm26: Done

src/query/insert.rs
```diff
@@ -3,6 +3,16 @@ use crate::{
...
 
+/// Represents a value source that can be used in an insert query.
+///
+/// [`InsertValueSource`] is a node in the expression tree and can represent a raw value set
+/// ('VALUES') or a select query.
+#[derive(Debug, Clone)]
+pub enum InsertValueSource {
+    VALUES(Vec<Vec<SimpleExpr>>),
+    QUERY(Box<SelectStatement>),
+}
```
billy1624, line 14: Could we rename variants as `Values` and `Select`?

05storm26: Done

billy1624 (approved):
Ty!

tyt2y3 (approved):
Wow. I absolutely love the API

tyt2y3:
okay. one question. why use case does pub-ing `value_source` allows?

05storm26:
@tyt2y3 I don't know. Should we make it non public? Should I open a new PR about it?

Issue #228, opened 2022-01-11
Common table expressions and `WITH RECURSIVE`


05storm26:
Basically support for this: https://www.postgresql.org/docs/9.1/queries-with.html

Especially useful would be the `WITH RECURSIVE` query support. This basically allows you to do all kinds of crazy stuffs like traversing a flattened tree using standard SQL syntax, etc.

billy1624:
Hey @05storm26, I think we can support this. The syntax of common table expressions is quite similar across all three databases.
- PostgreSQL: https://www.postgresql.org/docs/14/queries-with.html
- MySQL: https://dev.mysql.com/doc/refman/8.0/en/with.html
- SQLite: https://www.sqlite.org/lang_with.html

05storm26:
I am working on this.

tyt2y3:
Thank you @05storm26 actually I have used CTE in raw SQL and I think it is quite powerful!
Can you share a bit how you imagine the final API looks like?

billy1624:
Thanks!! @05storm26

PR #239, opened 2022-01-15
With clause and with queries

05storm26:
Fixes: #228 

Example:
```rust
use sea_query::{*, IntoCondition, IntoIden, tests_cfg::*};

let base_query = SelectStatement::new()
                    .column(Alias::new("id"))
                    .expr(Expr::val(1i32))
                    .column(Alias::new("next"))
                    .column(Alias::new("value"))
                    .from(Alias::new("table"))
                    .to_owned();

let cte_referencing = SelectStatement::new()
                            .column(Alias::new("id"))
                            .expr(Expr::col(Alias::new("depth")).add(1i32))
                            .column(Alias::new("next"))
                            .column(Alias::new("value"))
                            .from(Alias::new("table"))
                            .join(
                                JoinType::InnerJoin,
                                Alias::new("cte_traversal"),
                                Expr::tbl(Alias::new("cte_traversal"), Alias::new("next")).equals(Alias::new("table"), Alias::new("id")).into_condition()
                            )
                            .to_owned();

let common_table_expression = CommonTableExpression::new()
            .query(
                base_query.clone().union(UnionType::All, cte_referencing).to_owned()
            )
            .column(Alias::new("id"))
            .column(Alias::new("depth"))
            .column(Alias::new("next"))
            .column(Alias::new("value"))
            .table_name(Alias::new("cte_traversal"))
            .to_owned();

let select = SelectStatement::new()
        .column(ColumnRef::Asterisk)
        .from(Alias::new("cte_traversal"))
        .to_owned();

let with_clause = WithClause::new()
        .recursive(true)
        .cte(common_table_expression)
        .cycle(Cycle::new_from_expr_set_using(SimpleExpr::Column(ColumnRef::Column(Alias::new("id").into_iden())), Alias::new("looped"), Alias::new("traversal_path")))
        .to_owned();

let query = select.with(with_clause).to_owned();

assert_eq!(
    query.to_string(MysqlQueryBuilder),
    r#"WITH RECURSIVE `cte_traversal` (`id`, `depth`, `next`, `value`) AS (SELECT `id`, 1, `next`, `value` FROM `table` UNION ALL SELECT `id`, `depth` + 1, `next`, `value` FROM `table` INNER JOIN `cte_traversal` ON `cte_traversal`.`next` = `table`.`id`) SELECT * FROM `cte_traversal`"#
);
assert_eq!(
    query.to_string(PostgresQueryBuilder),
    r#"WITH RECURSIVE "cte_traversal" ("id", "depth", "next", "value") AS (SELECT "id", 1, "next", "value" FROM "table" UNION ALL SELECT "id", "depth" + 1, "next", "value" FROM "table" INNER JOIN "cte_traversal" ON "cte_traversal"."next" = "table"."id") CYCLE "id" SET "looped" USING "traversal_path" SELECT * FROM "cte_traversal""#
);
assert_eq!(
    query.to_string(SqliteQueryBuilder),
    r#"WITH RECURSIVE "cte_traversal" ("id", "depth", "next", "value") AS (SELECT "id", 1, "next", "value" FROM "table" UNION ALL SELECT "id", "depth" + 1, "next", "value" FROM "table" INNER JOIN "cte_traversal" ON "cte_traversal"."next" = "table"."id") SELECT * FROM "cte_traversal""#
);
```

### DIFF
--- a/src/backend/mysql/query.rs
+++ b/src/backend/mysql/query.rs
@@ -30,4 +30,30 @@ impl QueryBuilder for MysqlQueryBuilder {
         write!(sql, " ").unwrap();
         self.prepare_order(&order_expr.order, sql, collector);
     }
+
+    fn prepare_query_statement(
+        &self,
+        query: &dyn QueryStatementBuilder,
+        sql: &mut SqlWriter,
+        collector: &mut dyn FnMut(Value),
+    ) {
+        query.build_collect_any_into(self, sql, collector);
+    }
+
+    fn prepare_with_clause_recursive_options(
+        &self,
+        _: &WithClause,
+        _: &mut SqlWriter,
+        _: &mut dyn FnMut(Value),
+    ) {
+        // MySQL doesn't support sql recursive with query 'SEARCH' and 'CYCLE' options.
+    }
+
+    fn prepare_with_query_clause_materialization(
+        &self,
+        _: &CommonTableExpression,
+        _: &mut SqlWriter,
+    ) {
+        // MySQL doesn't support declaring materialization in SQL for with query.
+    }
 }
--- a/src/backend/postgres/query.rs
+++ b/src/backend/postgres/query.rs
@@ -90,4 +90,13 @@ impl QueryBuilder for PostgresQueryBuilder {
             Some(NullOrdering::First) => write!(sql, " NULLS FIRST").unwrap(),
         }
     }
+
+    fn prepare_query_statement(
+        &self,
+        query: &dyn QueryStatementBuilder,
+        sql: &mut SqlWriter,
+        collector: &mut dyn FnMut(Value),
+    ) {
+        query.build_collect_any_into(self, sql, collector);
+    }
 }
--- a/src/backend/query_builder.rs
+++ b/src/backend/query_builder.rs
@@ -635,6 +635,173 @@ pub trait QueryBuilder: QuotedBuilder {
         }
     }
 
+    /// Translate [`QueryStatement`] into SQL statement.
+    fn prepare_query_statement(
+        &self,
+        query: &dyn QueryStatementBuilder,
+        sql: &mut SqlWriter,
+        collector: &mut dyn FnMut(Value),
+    );
+
+    fn prepare_with_query(
+        &self,
+        query: &WithQuery,
+        sql: &mut SqlWriter,
+        collector: &mut dyn FnMut(Value),
+    ) {
+        self.prepare_with_clause(&query.with_clause, sql, collector);
+        self.prepare_query_statement(query.query.as_ref().unwrap().deref(), sql, collector);
+    }
+
+    fn prepare_with_clause(
+        &self,
+        with_clause: &WithClause,
+        sql: &mut SqlWriter,
+        collector: &mut dyn FnMut(Value),
+    ) {
+        self.prepare_with_clause_start(with_clause, sql);
+        self.prepare_with_clause_common_tables(with_clause, sql, collector);
+        if with_clause.recursive {
+            self.prepare_with_clause_recursive_options(with_clause, sql, collector);
+        }
+    }
+
+    fn prepare_with_clause_recursive_options(
+        &self,
+        with_clause: &WithClause,
+        sql: &mut SqlWriter,
+        collector: &mut dyn FnMut(Value),
+    ) {
+        if with_clause.recursive {
+            if let Some(search) = &with_clause.search {
+                write!(
+                    sql,
+                    "SEARCH {} FIRST BY ",
+                    match &search.order.as_ref().unwrap() {
+                        SearchOrder::BREADTH => "BREADTH",
+                        SearchOrder::DEPTH => "DEPTH",
+                    }
+                )
+                .unwrap();
+
+                self.prepare_simple_expr(&search.expr.as_ref().unwrap().expr, sql, collector);
+
+                write!(sql, " SET ").unwrap();
+
+                search
+                    .expr
+                    .as_ref()
+                    .unwrap()
+                    .alias
+                    .as_ref()
+                    .unwrap()
+                    .prepare(sql, self.quote());
+                write!(sql, " ").unwrap();
+            }
+            if let Some(cycle) = &with_clause.cycle {
+                write!(sql, "CYCLE ").unwrap();
+
+                self.prepare_simple_expr(cycle.expr.as_ref().unwrap(), sql, collector);
+
+                write!(sql, " SET ").unwrap();
+
+                cycle.set_as.as_ref().unwrap().prepare(sql, self.quote());
+                write!(sql, " USING ").unwrap();
+                cycle.using.as_ref().unwrap().prepare(sql, self.quote());
+                write!(sql, " ").unwrap();
+            }
+        }
+    }
+
+    fn prepare_with_clause_common_tables(
+        &self,
+        with_clause: &WithClause,
+        sql: &mut SqlWriter,
+        collector: &mut dyn FnMut(Value),
+    ) {
+        let mut cte_first = true;
+        assert_ne!(
+            with_clause.cte_expressions.len(),
+            0,
+            "Cannot build a with query that has no common table expression!"
+        );
+
+        if with_clause.recursive {
+            assert_eq!(
+                with_clause.cte_expressions.len(),
+                1,
+                "Cannot build a recursive query with more than one common table! \
+                A recursive with query must have a single cte inside it that has a union query of \
+                two queries!"
+            );
+        }
+        for cte in &with_clause.cte_expressions {
+            if !cte_first {
+                write!(sql, ", ").unwrap();
+            }
+            cte_first = false;
+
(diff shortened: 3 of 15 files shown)


05storm26:
I had to change `QueryStatementBuilder` to be able to create `&dyn QueryStatementBuilder` trait objects.

Please also note that with queries can be embedded into each other with no limit. In order to support that I am using trait objects:

```rust
#[derive(Debug, Clone, Default)]
pub struct WithClause {
    pub(crate) recursive: bool,
    pub(crate) search: Option<Search>,
    pub(crate) cycle: Option<Cycle>,
    pub(crate) cte_expressions: Vec<CommonTableExpression>,
}

#[derive(Debug, Clone, Default)]
pub struct WithQuery {
    pub(crate) with_clause: WithClause,
    pub(crate) query: Option<Box<dyn QueryStatementBuilder>>,
}
```

Then because I am manually implementing `QueryStatementBuilder` for `WithQuery` I can embed a `WithQuery` into another `WithQuery` as a `Box<dyn QueryStatementBuilder>`

05storm26:
@billy1624 I fixed the unused variable warning that caused the test to fail.

billy1624 (commented):
Hey @05storm26, I'm not sure how do we use the `with` method inside insert, update and delete

billy1624:
I can imagine INSERT with CTE but this requires the new API `select_from` (from PR#238)

```rust
#[test]
fn insert_6() -> sea_query::error::Result<()> {
    let select = SelectStatement::new()
        .columns([Glyph::Id, Glyph::Image, Glyph::Aspect])
        .from(Glyph::Table)
        .to_owned();
    let cte = CommonTableExpression::new()
        .query(select)
        .column(Glyph::Id)
        .column(Glyph::Image)
        .column(Glyph::Aspect)
        .table_name(Alias::new("cte"))
        .to_owned();
    let with_clause = WithClause::new().cte(cte).to_owned();
    let select = SelectStatement::new()
        .columns([Glyph::Id, Glyph::Image, Glyph::Aspect])
        .from(Alias::new("cte"))
        .to_owned();
    let mut insert = Query::insert();
    insert
        .into_table(Glyph::Table)
        .columns([Glyph::Id, Glyph::Image, Glyph::Aspect])
        .select_from(select)?;
    let sql = insert.with(with_clause).to_string(PostgresQueryBuilder);
    assert_eq!(
        sql.as_str(),
        [
            r#"WITH "cte" ("id", "image", "aspect") AS (SELECT "id", "image", "aspect" FROM "glyph")"#,
            r#"INSERT INTO "glyph" ("id", "image", "aspect") SELECT "id", "image", "aspect" FROM "cte""#,
        ].join(" ")
    );
    Ok(())
}
```

billy1624:
Some nitpicks
- https://github.com/billy1624/sea-query/commit/5eee3fed44aa0b901ec01d9e59e80f9678d6a91a
- https://github.com/billy1624/sea-query/commit/2d439de70805b3457c899dce29518499f425a2e7

The INSERT with CRE demo
- https://github.com/billy1624/sea-query/commit/1d2e4fe354f22e68300306ca2be165c8a813ad6d

05storm26:
> I can imagine INSERT with CTE but this requires the new API select_from (from #238)

Yeah actually :D I didn't reallize that it was dependent on that PR for inserts since I didn't write tests for insert with CTE but its fine since it's merged now.

05storm26:
Thanks for the fixes I have merged the small ones into my changeset. I left the larger one that creates a test for postgres alone as a separate commit because that can kind of stand on itself. (Except I have edited the commit message a little bit)

billy1624:
I'm still not sure about the `with` method inside update and delete. How are we going to use it?

05storm26:
I don't think you can use it with delete and update right now. Postgres and sqlite supports FROM in UPDATE queries. If you want to use CTE in an UPDATE query we will need to add support FROM in UPDATE. 

You can also use the common tables created in the with clause in subqueries or in the query as a table like this:

https://stackoverflow.com/a/21823546

I am not sure about  mysql.

In delete you can use the common table in subqueries or as a table name like: 
```
WITH IDsToDelete AS (...)
DELETE FROM MyTable
WHERE ID IN IDsToDelete;
```

Currently you cannot have an sea-query query expression contain a WithQuery. The `SimpleExpr::SubQuery` can only contain `SelectStatement`. I have opened an issue about that: https://github.com/SeaQL/sea-query/issues/240

In general we should allow `Box<&dyn QueryStatementBuilder>` in most expression places (especially in `SimpleExpr::SubQuery`) and that would allow more usablility of all kinds of composite queries. (Including UPDATE/DELETE ... RETURNING queries, which also cannot be used currently with the current  `SimpleExpr::SubQuery` design.

05storm26:
In general I think queries should be accessed using an interface to allow composability, that is why this PR moves towards using queries through: `Box<&dyn QueryStatementBuilder>`.

But more work needed beyond this PR like implementing:  #240  And we should add support for the FROM clause in UPDATE

billy1624:
> I don't think you can use it with delete and update right now.

Should we revert the `with` method inside delete & update for now? We can try and make it work in another PR.

05storm26:
> Should we revert the with method inside delete & update for now? We can try and make it work in another PR.

Yes I will do that shortly. Supporting SELECT queryies is more important anyway.

billy1624:
Thanks! @05storm26

05storm26:
I have done 2 things in this push:

 1. I have fixed a bug in the `QueryBuilder::prepare_query_statement` that caused the positional arguments to have a incorrect index.
 2. I have added examples to update and delete about the use of with query.
 
I have added a 3rd commit that fixes: #240.

tyt2y3:
> that caused the positional arguments to have a incorrect index

Can you point out where? Is it a pre-existing bug or just something introduced in this PR?

05storm26:
> Is it a pre-existing bug or just something introduced in this PR?

It was caused by this PR. It is fixed in this force push: https://github.com/SeaQL/sea-query/compare/736f7b0c7e22f0b4c48baeed321a8fd3095be55d..e4d43a136f9731e36349a97d16c8d222605df13a

It was caused by creating a new `SqlWriter` when calling `prepare_query_statement` that is preparing the actual query of the with query or in the case of subqueries the subquery. This was caused because `prepare_query_statement` was initially implemented to use `build_collect_any` but that creates a new `SqlWriter`.

By creating a new `SqlWriter` the positinal indexes were restarted and conflicted in the whole query.

To fix it the new `prepare_query_statement` now doesn't call `build_collect_any` but a new `build_collect_any_into` that is exactly like `build_collect_any` except it doesn't create a new `SqlWriter` but receives it as an argument, so by using that inside of `prepare_query_statement`, it continues to work into the same `SqlWriter` and thus the parameter indexes are not restarted.

Because numbered parameters are used only by `PostgresQueryBuilder` this issue was only affecting that generator.

But now this is fixed using the `build_collect_any_into`.

tyt2y3:
Thanks for the explanation! Sounds intricate. So our test cases covered that now?

05storm26:
I actually caught it when I rewrote the `SimpleExpr::SubQuery` to use `prepare_query_statement` as well. There was a single example in rustdoc about a subquery expression that brought this out: 

src/lib.rs:207 : 
```rust
 use sea_query::{*, tests_cfg::*};
 assert_eq!(
     Query::select()
         .column(Char::Character)
         .from(Char::Table)
         .and_where(
             Expr::expr(Expr::col(Char::SizeW).add(1))
                 .mul(2)
                 .equals(Expr::expr(Expr::col(Char::SizeH).div(2)).sub(1))
         )
         .and_where(
             Expr::col(Char::SizeW).in_subquery(
                 Query::select()
                     .expr(Expr::cust_with_values("ln(? ^ ?)", vec![2.4, 1.2]))
                     .take()
             )
         )
         .and_where(
             Expr::col(Char::Character)
                 .like("D")
                 .and(Expr::col(Char::Character).like("E"))
         )
     [
         r#"SELECT "character" FROM "character""#,
         r#"WHERE ("size_w" + 1) * 2 = ("size_h" / 2) - 1"#,
         r#"AND "size_w" IN (SELECT ln(2.4 ^ 1.2))"#,
         r#"AND (("character" LIKE 'D') AND ("character" LIKE 'E'))"#,
     ]
     .join(" ")
```

It triggered this issue because it had multiple query parameter in the original query (before the appearance of the subquery ) and more than one in the subquery itself, that all had different values, (2 and 1, and then in the subquery 2.4 and 1.2). Also the `PostgresQueryBuilder` is used.

I don't know if the examples count as "test". They are run when you run `cargo test` although I think they are more prone to change since people just perceive them as non important documentation for users. Do you think we should add this query as a separate test function?

tyt2y3:
awesome. I think it's good enough

rybertm:
> And we should add support for the FROM clause in UPDATE

I couldn't find an issue regarding this, is it being tracked anywhere?

Sorry for commenting on this closed issue.

ikrivosheev:
@rybertm hello! I am not in context. Can you open new issues and explain the problem?

PR #243, opened 2022-01-21
SQLite does not support "LATERAL join"

billy1624:
## PR Info
- Will panic when building SQLite select with "LATERAL join"
- Continue of #224

### DIFF
--- a/src/backend/sqlite/query.rs
+++ b/src/backend/sqlite/query.rs
@@ -14,6 +14,18 @@ impl QueryBuilder for SqliteQueryBuilder {
         // SQLite doesn't supports row locking
     }
 
+    fn prepare_join_table_ref(
+        &self,
+        join_expr: &JoinExpr,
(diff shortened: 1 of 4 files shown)


rex-remind101 (approved):
excellent, thank you

tyt2y3:
I don't know how we should deal with serialization errors. Seems like we need a `Result` interface instead of panics

rex-remind101:
Wondering if there was any follow up here?

billy1624:
> I don't know how we should deal with serialization errors. Seems like we need a `Result` based API instead of panics

This would be a major breaking change across SeaQuery.

> Wondering if there was any follow up here?

I think we can leave it as is for now? We already added a line in the docs saying this is not supported by SQLite.

https://docs.rs/sea-query/0.22.0/sea_query/query/struct.SelectStatement.html#method.join_lateral

Issue #246, opened 2022-01-24
Add support for ORDER BY FIELD/CASE


PauMAVA:
In MySQL there is a `FIELD` function that can be used with `ORDER BY` to order the result in the same order as a list of elements. For example:
```sql
SELECT * FROM my_table
ORDER BY FIELD(my_col, ['USD', 'CAD', 'AUD', 'BBD', 'EUR', 'GBP'])
```
Would order the result according to the value of `my_col` being `USD` rows before `CAD` rows before `AUD`...
In Postgres and SQLite this can be achieved by using the following syntax:
```sql
ORDER BY
  CASE
    WHEN code='USD' THEN 1
    WHEN code='CAD' THEN 2
    WHEN code='AUD' THEN 3
    WHEN code='BBD' THEN 4
    WHEN code='EUR' THEN 5
    WHEN code='GBP' THEN 6
    ELSE 7
```
As a side note, this is needed in order to accomplish foreign-key marker support in `sea-orm`.

I propose that we add a new variant to the `Order` enum such as:
```rust
pub Order {
    Asc,
    Desc,
    Field((ColumnRef, Vec<DynIden>)),
}
```
Where `ColumnRef` is the column where the values belong and `Vec<DynIden>` the values to be used to order the result.

billy1624:
Hey @PauMAVA, thanks for the help :)

I think the `ColumnRef` is not necessary, as we can get store "the column" inside `OrderExpr::expr`.

https://github.com/SeaQL/sea-query/blob/6c11e64dd7b88b994fd778a0763161a9801a7769/src/types.rs#L161-L167

Also, perhaps we can change `Field((ColumnRef, Vec<DynIden>))` into `Field(Values)`, where `Values` is a vector of values.

https://github.com/SeaQL/sea-query/blob/6c11e64dd7b88b994fd778a0763161a9801a7769/src/value.rs#L108-L109

Does this make sense?

PauMAVA:
Yes @billy1624 , I have only one doubt.

In `OrderExpr.expr` there will be only the column name, right? I will implement this today and open a PR😄.

billy1624:
For `OrderExpr.expr` it's of type `SimpleExpr`. Column name is one of the possible variant of it, i.e. `SimpleExpr::Column(ColumnRef)`.

https://github.com/SeaQL/sea-query/blob/6c11e64dd7b88b994fd778a0763161a9801a7769/src/expr.rs#L20-L38

[...]

PR #248, opened 2022-01-25
Add support for ORDER BY FIELD/CASE

PauMAVA:
Implemented #246 and tested.

The correct syntax differs a little from the one in #246:

PostgreSQL and SQLite:
```sql
ORDER BY
  CASE
    WHEN code='USD' THEN 1
    WHEN code='CAD' THEN 2
    WHEN code='AUD' THEN 3
    WHEN code='BBD' THEN 4
    WHEN code='EUR' THEN 5
    WHEN code='GBP' THEN 6
    ELSE 7
    END
```

MySQL:
```sql
SELECT * FROM my_table
ORDER BY FIELD(my_col, 'USD', 'CAD', 'AUD', 'BBD', 'EUR', 'GBP')
```

"Major" changes:
* `prepare_order` takes the `OrderExpr` instead of only `Order` as it needs access to the column name if the order is `Order::Field`
* `Order` does not derive `Eq, Copy` anymore as `Values` does not implement those traits.

Implementation notes:
* Added a new function called `prepare_field_order` in `QueryBuilder` trait.
* Default implementation inside the trait of `prepare_field_order` produces the `CASE-WHEN-THEN-ELSE-END` syntax.
* Reimplementation inside `MySQLQueryBuilder` produces the `FIELD` syntax.
*  Inside `prepare_order_expr` if `matches!(order_expr.order, Order::Field(_))` the column name is not written after the `ORDER BY`.

### DIFF
--- a/tests/mysql/query.rs
+++ b/tests/mysql/query.rs
@@ -900,6 +900,54 @@ fn select_53() {
 }
 
 #[test]
+fn select_54() {
+    assert_eq!(
+        Query::select()
+            .columns(vec![Glyph::Aspect,])
+            .from(Glyph::Table)
+            .and_where(Expr::expr(Expr::col(Glyph::Aspect).if_null(0)).gt(2))
+            .order_by(
+                Glyph::Id,
+                Order::Field(Values(vec![4.into(), 5.into(), 1.into(), 3.into(),]))
+            )
+            .order_by((Glyph::Table, Glyph::Aspect), Order::Asc)
+            .to_string(MysqlQueryBuilder),
+        [
+            r#"SELECT `aspect`"#,
+            r#"FROM `glyph`"#,
+            r#"WHERE IFNULL(`aspect`, 0) > 2"#,
+            r#"ORDER BY FIELD(`id`, 4, 5, 1, 3),"#,
+            r#"`glyph`.`aspect` ASC"#,
+        ]
+        .join(" ")
+    );
+}
+
+#[test]
+fn select_55() {
+    assert_eq!(
+        Query::select()
+            .columns(vec![Glyph::Aspect,])
+            .from(Glyph::Table)
+            .and_where(Expr::expr(Expr::col(Glyph::Aspect).if_null(0)).gt(2))
+            .order_by((Glyph::Table, Glyph::Aspect), Order::Asc)
+            .order_by(
+                Glyph::Id,
+                Order::Field(Values(vec![4.into(), 5.into(), 1.into(), 3.into(),]))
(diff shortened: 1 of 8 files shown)


tests/mysql/query.rs
```diff
@@ -899,6 +899,54 @@ fn select_53() {
...
+                Glyph::Id,
+                Order::Field(Values(vec![4.into(), 5.into(), 1.into(), 3.into(),]))
+            )
+            .to_string(MysqlQueryBuilder),
+        [
+            r#"SELECT `aspect`"#,
+            r#"FROM `glyph`"#,
+            r#"WHERE IFNULL(`aspect`, 0) > 2"#,
+            r#"ORDER BY `glyph`.`aspect` ASC,"#,
+            r#"FIELD(`id`, 4, 5, 1, 3)"#,
```
billy1624, line 944: According to the [MySQL docs of `FIELD` function](https://dev.mysql.com/doc/refman/8.0/en/string-functions.html#function_field)

> If str is NULL, the return value is 0 because NULL fails equality comparison with any value. FIELD() is the complement of ELT().

That mean if the value of `id` column is not exists in array `[4, 5, 1, 3]`, then those rows will be sorted and placed in front of those with `id` exists in the array.

This behaviour is different from PostgreSQL & SQLite. Which both of them placed those rows behind instead of in the front.

PauMAVA: You are right. I will modify the `MySQL` query builder to produce:
```sql
SELECT `aspect` FROM `glyph`
ORDER BY IF(`id` NOT IN (4, 5, 1, 3),4,0), FIELD(`id`, 4, 5, 1, 3)
```
This causes all the ids not in the set passed to the `FIELD` function to have an order or `4` (in this case as the length of the set is 4) and thus being placed after `4,5,1,3`.

billy1624: Or we use [case when](https://dev.mysql.com/doc/refman/8.0/en/case.html) syntax on all three databases?

PauMAVA: Oh, I see. Didn't know that it was supported by MySQL. Let me change it.

billy1624 (commented):
Hey @PauMAVA, just a nitpick. See if it make sense :P
https://github.com/PauMAVA/sea-query/compare/oder-by-field-case...billy1624:pr/248

PauMAVA:
Makes sense. Thanks! I think this would be ready for merging.

billy1624 (approved):
Thanks!! @PauMAVA

PR #256, opened 2022-02-08
Add "with-time" feature

cemoktra:
`with-time` and `with-chrono` are mutually exclusive

### DIFF
--- a/src/value.rs
+++ b/src/value.rs
@@ -10,7 +10,7 @@ use std::str::from_utf8;
 use chrono::{DateTime, FixedOffset, Local, NaiveDate, NaiveDateTime, NaiveTime, Utc};
 
 #[cfg(feature = "with-time")]
-use time::{PrimitiveDateTime, OffsetDateTime, macros::offset};
+use time::{macros::offset, OffsetDateTime, PrimitiveDateTime};
 
 #[cfg(feature = "with-rust_decimal")]
 use rust_decimal::Decimal;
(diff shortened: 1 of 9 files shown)


src/value.rs
```diff
@@ -583,30 +675,37 @@ impl Value {
...
 
 impl Value {
     pub fn is_date_time_utc(&self) -> bool {
-        #[cfg(feature = "with-chrono")]
+        #[cfg(any(feature = "with-chrono", feature = "with-time"))]
         return matches!(self, Self::DateTimeUtc(_));
-        #[cfg(not(feature = "with-chrono"))]
+        #[cfg(not(any(feature = "with-chrono", feature = "with-time")))]
         return false;
     }
```
billy1624, line 697: I think we could refactor all these feature guards by adding a marker feature, perhaps named `_with-chrono-or-time`. This feature will be turned on if either `with-chrono` or `with-time` was enabled.

Then, we could simple check
```rust
#[cfg(feature = "_with-chrono-or-time")]
#[cfg(not(feature = "_with-chrono-or-time"))]
```

billy1624 (commented):
Btw... we should also tests if `time` can be binded on various db drivers or not.
https://github.com/SeaQL/sea-query/tree/master/src/driver

src/value.rs
tyt2y3: I think it's fine either way

tyt2y3:
> Btw... we should also tests if `time` can be binded on various db drivers or not. https://github.com/SeaQL/sea-query/tree/master/src/driver

what is the support matrix?

tyt2y3:
Thank you so much for the contribution. I have a clearer thought by now:

SO the problem is this creates a confusion. Cargo features are not supposed to be mutually exclusive, although quite a number of crates does so. Also, all-features does not compile.
But inevitably if some crate turned on the `time` feature there might be another crate not expecting that and result in compilation failure.

It will be hard to maintain in the long run if we support the two using mutually exclusive feature flags.

What I have in mind is may be we can have a separate "facade" crate that abstracts the two libraries and provide a uniform interface for SeaQuery (and other downstream crates) to consume.

tyt2y3:
Actually, it might makes more sense to support both to coexist at the same time.

tyt2y3:
> Actually, it might makes more sense to support both to coexist at the same time.

This should be easier and a once off breaking change right? @billy1624

billy1624:
> > Actually, it might makes more sense to support both to coexist at the same time.
> 
> This should be easier and a once off breaking change right? @billy1624

Yeah, this would be easier to just add `time` types as new `Value`'s variants, alongside with `chrono`'s variants. It will be a breaking change but I think it's fine as long as we keep all old `Value` variants untouched. This is easy to fix for anyone matching the `Value` Enum.

tyt2y3:
Sounds good

tyt2y3:
Closed in favour of #267

PR #257, opened 2022-02-08
support postgresql distinct on

### LINKED ISSUE #250: add postgres distinct on please
I tried this :
Query::select()
            .expr(Expr::cust("distinct on (id)"))

 but got a comma here:
SELECT distinct on (id), "id" FROM
____________________^

https://www.postgresql.org/docs/12/sql-select.html#SQL-DISTINCT

oonxt:
fix #250

### DIFF
--- a/src/backend/query_builder.rs
+++ b/src/backend/query_builder.rs
@@ -374,16 +374,23 @@ pub trait QueryBuilder: QuotedBuilder {
         sql: &mut SqlWriter,
         _collector: &mut dyn FnMut(Value),
     ) {
-        write!(
-            sql,
-            "{}",
-            match select_distinct {
-                SelectDistinct::All => "ALL",
-                SelectDistinct::Distinct => "DISTINCT",
-                SelectDistinct::DistinctRow => "DISTINCTROW",
+        match select_distinct {
+            SelectDistinct::All => write!(sql, "{}", "ALL").unwrap(),
+            SelectDistinct::Distinct => write!(sql, "{}", "DISTINCT").unwrap(),
+            SelectDistinct::DistinctRow => write!(sql, "{}", "DISTINCTROW").unwrap(),
+            #[cfg(feature = "backend-postgres")]
+            SelectDistinct::DistinctOn(cols) => {
+                write!(sql, "{}", "DISTINCT ON (").unwrap();
+                cols.into_iter().fold(true, |first, c| {
+                    if !first {
+                        write!(sql, "{}", ", ").unwrap();
+                    }
+                    c.prepare(sql, self.quote());
+                    false
+                });
+                write!(sql, "{}", ")").unwrap();
             }
-        )
-        .unwrap();
+        };
     }
 
     /// Translate [`LockType`] into SQL statement.
--- a/src/query/select.rs
+++ b/src/query/select.rs
@@ -54,11 +54,13 @@ pub struct SelectStatement {
 }
 
 /// List of distinct keywords that can be used in select statement
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[derive(Debug, Clone)]
 pub enum SelectDistinct {
     All,
     Distinct,
     DistinctRow,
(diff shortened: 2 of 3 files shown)


billy1624 (commented):
Hey @oonxt, Thank you for the contributions! Sorry for the delay. Is there any reason for you to use `DISTINCT ON` over [`group_by_col`](https://docs.rs/sea-query/0.21.0/sea_query/query/struct.SelectStatement.html#method.group_by_col)?

oonxt:
Basically distinct on can be rewritten with group by, but in some cases it may be necessary to use subqueries or window functions, while distinct on is much simpler.

like this:

1. the group by version:
with data(user_id, product_id, paid_at) as (
    values (1, 1, '2022-01-01 13:23:21'::timestamptz),
           (1, 2, '2022-01-02 22:30:55'::timestamptz),
           (2, 1, '2022-01-01 09:05:28'::timestamptz),
           (2, 2, '2022-01-02 20:11:14'::timestamptz)
)
select a.user_id, a.product_id
From data a,
     (
         select user_id, max(paid_at) paid_at
         from data
         group by user_id
     ) b
where a.user_id = b.user_id
  and a.paid_at = b.paid_at;


2. the window function version:
with data(user_id, product_id, paid_at) as (
    values (1, 1, '2022-01-01 13:23:21'::timestamptz),
           (1, 2, '2022-01-02 22:30:55'::timestamptz),
           (2, 1, '2022-01-01 09:05:28'::timestamptz),
           (2, 2, '2022-01-02 20:11:14'::timestamptz)
)
select user_id, product_id
From (
         select user_id, product_id, row_number() over (partition by user_id order by paid_at desc nulls last) rn
         from data
     ) a
where a.rn = 1

3. the distinct on version:
with data(user_id, product_id, paid_at) as (
    values (1, 1, '2022-01-01 13:23:21'::timestamptz),
           (1, 2, '2022-01-02 22:30:55'::timestamptz),
           (2, 1, '2022-01-01 09:05:28'::timestamptz),
           (2, 2, '2022-01-02 20:11:14'::timestamptz)
)
select distinct on (user_id) user_id, product_id
from data
order by user_id, paid_at desc nulls last;

src/query/select.rs
```diff
@@ -283,6 +285,22 @@ impl SelectStatement {
         self
     }
 
+    /// Select distinct on
```
billy1624, line 288: Hey @oonxt, all in all the PR looks good to me. Just a nitpick, can we add some example here? Also, should we emphasize on the docs that the `distinct_on` method only works for PostgreSQL?

billy1624:
Thanks!! @oonxt So, in other words, `DISTINCT ON` is a shortcut of `GROUP BY`

src/query/select.rs
oonxt: Sorry for late. I'll make a new pr later.

src/backend/query_builder.rs
```diff
@@ -376,16 +376,23 @@ pub trait QueryBuilder: QuotedBuilder {
...
-            "{}",
-            match select_distinct {
-                SelectDistinct::All => "ALL",
-                SelectDistinct::Distinct => "DISTINCT",
-                SelectDistinct::DistinctRow => "DISTINCTROW",
+        match select_distinct {
+            SelectDistinct::All => write!(sql, "{}", "ALL").unwrap(),
+            SelectDistinct::Distinct => write!(sql, "{}", "DISTINCT").unwrap(),
+            SelectDistinct::DistinctRow => write!(sql, "{}", "DISTINCTROW").unwrap(),
+            #[cfg(feature = "backend-postgres")]
```
tyt2y3, line 383: Sorry if I have not documented it somewhere. As a rule-of-thumb, backend does not make use of feature guard. Instead, we should add the relevant feature in the `postgres` backend itself. This file is only for features common across all backends.

tyt2y3 (changes requested):

ikrivosheev:
@oonxt thank you for the contribution! Do you want to complete PR?)

PR #261, opened 2022-02-09
Support multiple tables in the select from

### LINKED ISSUE #258: SelectStatement from should be a Vec
Currently I think it is possible to select from multiple tables and do an "old fashion" join in the where clauses because the `.from` method overrides the from on each call. I believe it should be a `Vec` though unsure how to do it without doing a breaking change.

Sytten:
Fixes #258 

This is a breaking change since the previous behaviour would replace the `from` table when you called it multiple times.
I think this is correct, just need to be marked as such in the changelog.

### DIFF
--- a/src/backend/query_builder.rs
+++ b/src/backend/query_builder.rs
@@ -89,9 +89,15 @@ pub trait QueryBuilder: QuotedBuilder {
             false
         });
 
-        if let Some(from) = &select.from {
+        if !select.from.is_empty() {
             write!(sql, " FROM ").unwrap();
-            self.prepare_table_ref(from, sql, collector);
+            select.from.iter().fold(true, |first, table_ref| {
+                if !first {
+                    write!(sql, ", ").unwrap()
+                }
+                self.prepare_table_ref(table_ref, sql, collector);
+                false
+            });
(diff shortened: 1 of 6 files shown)


billy1624 (commented):
Thanks!! @Sytten

Should we add a method take multiple `FROM` at once?

```rust
pub fn froms<R, I>(&mut self, tbl_refs: I) -> &mut Self
    where
        R: IntoTableRef,
        I: IntoIterator<Item = R>;
```

Sytten:
The name is a little weird, but yeah this is a good idea. Let me code it real quick

billy1624:
> The name is a little weird

agree loll, is there any better name for it?

Sytten:
`from_many`?

billy1624:
> `from_many`?

Yeah, much better :)

Sytten:
I just realized, I don't think we can do that because the Iden will never be from the same enum so you can't easily build a vec of them :/

billy1624:
Oh, right. Sorry, I forgot that. Then we could leave it as is

billy1624 (approved):
Ty!

PR #262, opened 2022-02-09
Add support for replace insert

Sytten:
Adds support for the `REPLACE` directive. Eventually a better implementation (at least for Sqlite) would add all the `OR REPLACE | ABORT | FAIL | IGNORE | ROLLBACK` or we could direct people to use the `ON CONFLICT` syntax (when #167 is merged)


billy1624:
> We could direct people to use the `ON CONFLICT` syntax (when #167 is merged)

This make sense.

billy1624 (approved):
Looks good to me! Thanks! @Sytten

PR #264, opened 2022-02-18
postgres ltree support

cemoktra:
## PR Info

- Dependencies:
  - https://github.com/SeaQL/sea-schema/pull/46
  - https://github.com/SeaQL/sea-orm/pull/545

Need to wait for next sqlx release before merging this.

## Adds

This PR adds sqlx ltree support to sea-query


cemoktra:
sqlx has been released, in order to make it work i also had to update rusqlite

Got one question, should i have an ltree feature or rather hide it behind sqlx/postgres? Cause on sqlx side ltree is not a feature

cemoktra:
Ok i removed the `with-tree` feature and put it to `backend-postgres`, it simplifies things and as its not a feature on `sqlx` side it makes perfect sense

tyt2y3:
Okay this seems impressive. Although the problem is we leaked the `lquery` type from SQLx, which breaks the design philosophy of sea-query. sea-query should be a pure Rust library, and so to properly support this API, we have to expose our own lquery type and only translate to SQLx type for the SQLx driver.

cemoktra:
Ok. Just tried to do it like with UUID for example

cemoktra:
sea-query could map to a string internally i guess,if you want to more you can use sqlx structures and convert them into strings, what do you think? @tyt2y3

cemoktra:
I'm still not sure if i understood how you would like to have this

cemoktra:
Basically you would like to have copies of `LTree`, `LTreeLabel` and `LQuery` of SQLx in `extension/postgres/types.rs` pr `backend/postgres/types.rs`?

tyt2y3:
sea-query's frontend is (and should be) backend-agnostic. for example, one might be using the `postgres` driver. we cannot leak the types from SQLx

cemoktra:
Should have added this via external crate into SQLx...

billy1624 (commented):
Hey @cemoktra, sorry for the delay and thanks for the great contributions!!

I notice you're the contributor who added `ltree` and `lquery` support to SQLx! 
- https://github.com/launchbadge/sqlx/pull/1696
- https://github.com/launchbadge/sqlx/pull/1710

> Should have added this via external crate into SQLx...

I second putting `PgLQuery` and `PgLTree` into a separate crate then use it in SQLx and SeaQuery.

We could get this merge before moving on to SeaSchema and SeaORM
- https://github.com/SeaQL/sea-schema/pull/46
- https://github.com/SeaQL/sea-orm/pull/545

Thanks!!

PR #266, opened 2022-03-08
Insert Default

### LINKED ISSUE #253: Insert default values behaviour
I totally think that the above API behaviour is too smart, the following should be better.
There is no basis that the query builder just assumes that `DEFAULT VALUES` is valid for the table either.

```rust
assert_eq!(
    Query::insert()
        .into_table(Glyph::Table)
        .or_default_values()
    "INSERT INTO \"glyph\" DEFAULT VALUES"
);
assert_eq!(
    Query::insert()
        .into_table(Glyph::Table)
        .columns(vec![Glyph::Image])
        .values_panic(vec!["ABC".into()])
        .or_default_values()
    "INSERT INTO \"glyph\" (\"image\") VALUES ('ABC')"
);
```

_Originally posted by @tyt2y3 in https://github.com/SeaQL/sea-query/pull/223#issuecomment-1030835987_

billy1624:
## PR Info

- Closes #253
- Continues #223
- Dependents:
  - 

## Adds

- Insert default
- `InsertStatement::or_default_values` method

## References

- MySQL: https://dev.mysql.com/doc/refman/8.0/en/insert.html
- PostgreSQL: https://www.postgresql.org/docs/9.1/dml-insert.html
- SQLite: https://www.sqlite.org/lang_insert.html


tyt2y3:
Okay I am just being curious, does the SQL syntax actually allow multiple DEFAULT VALUES?

src/backend/mysql/query.rs
```diff
@@ -58,4 +58,8 @@ impl QueryBuilder for MysqlQueryBuilder {
     ) {
         // MySQL doesn't support declaring materialization in SQL for with query.
     }
+
+    fn insert_default_keyword(&self) -> &str {
+        "VALUES (DEFAULT)"
```
tyt2y3, line 63: it does seem like it's only for inserting one row

ikrivosheev:
@billy1624 can you resolve the merge conflict?

src/backend/mysql/query.rs
billy1624: It will insert only one row with default values

tyt2y3:
> MySQL/Pg/SQLite actually allow multiple DEFAULT VALUES?

I think we need to investigate this and if it's valid SQL we should implement accordingly

billy1624:
Yes, it's possible to insert multiple default rows in single insert statement on MySQL & PostgreSQL but not SQLite

Demo
- PostgreSQL: https://dbfiddle.uk/?rdbms=postgres_14&fiddle=028aaaef2ba22652f3d647ab758d39a4
- MySQL: https://dbfiddle.uk/?rdbms=mysql_8.0&fiddle=c30eb9bd39a8e82c053d463f14ac3df3
- MariaDB: https://dbfiddle.uk/?rdbms=mariadb_10.6&fiddle=c30eb9bd39a8e82c053d463f14ac3df3
- SQLite: https://dbfiddle.uk/?rdbms=sqlite_3.27&fiddle=0f22bc8c9b09c0e6d295771239292e1c

billy1624:
> Well then may be we should further work on it

I have added a method, `or_default_values_many`, for inserting multiple rows with default values

tests/postgres/query.rs
```diff
@@ -1125,6 +1126,29 @@ fn insert_6() -> sea_query::error::Result<()> {
...
+
+#[test]
+fn insert_8() {
+    assert_eq!(
+        Query::insert()
+            .into_table(Glyph::Table)
+            .or_default_values()
+            .returning_col(Glyph::Id)
+            .to_string(PostgresQueryBuilder),
+        "INSERT INTO \"glyph\" VALUES (DEFAULT) RETURNING \"id\""
```
ikrivosheev, line 1148: `r#"..."#`

tests/postgres/query.rs
```diff
@@ -1125,6 +1126,29 @@ fn insert_6() -> sea_query::error::Result<()> {
...
 }
 
+#[test]
+fn insert_7() {
+    assert_eq!(
+        Query::insert()
+            .into_table(Glyph::Table)
+            .or_default_values()
+            .to_string(PostgresQueryBuilder),
+        "INSERT INTO \"glyph\" VALUES (DEFAULT)"
```
ikrivosheev, line 1136: r#"..."#

ikrivosheev (commented):
@billy1624 thank you! Some comments.

ikrivosheev (approved):
@billy1624 LGTM!

tyt2y3:
Cool. I think the important piece is how we are implementing the backend

PR #267, opened 2022-03-12
Supports `time` along with with `chrono`

billy1624:
## PR Info

- Continues #256
- Dependents:
  - SeaQL/sea-orm#602

## Adds

- Add `time` crate related features: `with-time`, `postgres-time`
- Add `time` variants to `Value`
- Supports binding of `time` `Value` in all drivers

## Breaking Changes

- Prefix the name of `chrono` `Value` variants with `Chrono`


tyt2y3:
awesome. would appreciate at least one example would make use of this new type

tyt2y3 (changes requested):
Example needed

billy1624:
> Example needed

Added

tyt2y3 (approved):
Awesome

PR #270, opened 2022-03-13
Allow for trailing commas in any and all macros

### LINKED ISSUE #268: Allow trailing commas in all and any macros
## Summary
Basically, it'd be nice to have trailing commas in `any!` and `all!`. The following code won't compile due to that trailing comma

```rust
sea_query::all![
    Expr::col(Node::Uid).eq(uid),
    Expr::col(Node::TenantId).eq(tenant_id),
//                                         ^ that little angel right there
]
```

I'm a novice with macros but I google'd how to do it and this looks right.

```rust
macro_rules! all {
    ( $( $x:expr ),* $(,)?) => {
        {
            let mut tmp = sea_query::Condition::all();
            $(
                tmp = tmp.add($x);
            )*
            tmp
        }
    };
}
```

All I did was add ` $(,)?` which I guess makes the last comma optional.

## Motivation

It's a really tiny, minor thing. But trailing commas make life better. I can comment out the last line, or copy/paste things around, or re-order lines, without fear.

## Additional Information

Note that the suggested approach I've provided does require rust 1.32, which I think is edition 2018. Given that this crate is 2021, there should be no risk of this change breaking anything.

insanitybit:
## PR Info

- Closes #268

I didn't look around for other cases where this might be nice. These are just the two I happened to run into.

## Adds

Support for trailing commas for the `any` and `all` macros.

## Breaking Changes

Backwards compatible.

## Changes

Tiny couple-character change to the macro definitions in `src/query/condition.rs`.

### DIFF
--- a/src/query/condition.rs
+++ b/src/query/condition.rs
@@ -272,7 +272,7 @@ impl std::convert::From<SimpleExpr> for ConditionExpression {
 /// ```
 #[macro_export]
 macro_rules! any {
-    ( $( $x:expr ),* ) => {
+    ( $( $x:expr ),* $(,)?) => {
         {
             let mut tmp = sea_query::Condition::any();
             $(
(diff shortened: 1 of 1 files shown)


billy1624 (approved):
Looks good! Thanks!! @insanitybit

tyt2y3 (approved):

Issue #174, opened 2021-10-28
Support window functions


ikrivosheev:
Hello, thank you very match for very useful library! 

I find issue and read the comment: https://github.com/SeaQL/sea-query/issues/15#issuecomment-825363312. 
Please, can you add support for window functions.

ikrivosheev:
I'll try to prepare PR in a couple of days

billy1624:
Hi @ikrivosheev, thanks in advance!!

Sytten:
I guess there was not any progress on that?
I happen to need `COUNT(*) OVER () AS count`

ikrivosheev:
@Sytten I apologize for the long delay! I'll make PR to the end of the week

Sytten:
No worries at all! I was just checking, we all have busy lives 😃
Looking forward to the PR.

billy1624:
Hey @ikrivosheev, no worries! Thank you so much! :)

PR #271, opened 2022-03-20
issues-174 Add support for WINDOWS statement

ikrivosheev:
## PR Info

- Closes https://github.com/SeaQL/sea-query/issues/174

## Adds

Add support for WINDOW statement

**TODO:**

- [x] support window statement
- [x] add more helpers functions
- [x] add more tests
- [x] add more docs

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,6 @@ name = "sea_query"
 path = "src/lib.rs"
 
 [dependencies]
-either = "1.6.1"
 sea-query-derive = { version = "0.2.0", path = "sea-query-derive", optional = true }
 serde_json = { version = "^1", optional = true }
 bytes = { version = "^1", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,7 @@ name = "sea_query"
 path = "src/lib.rs"
 
 [dependencies]
+either = "1.6.1"
 sea-query-derive = { version = "0.2.0", path = "sea-query-derive", optional = true }
 serde_json = { version = "^1", optional = true }
 bytes = { version = "^1", optional = true }
--- a/src/backend/query_builder.rs
+++ b/src/backend/query_builder.rs
@@ -1,5 +1,4 @@
 use crate::*;
-use either::Either;
 use std::ops::Deref;
 
 pub trait QueryBuilder: QuotedBuilder {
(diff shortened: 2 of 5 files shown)


src/backend/query_builder.rs
```diff
@@ -1177,6 +1198,70 @@ pub trait QueryBuilder: QuotedBuilder {
...
+            });
+        }
+
+        if let Some(frame) = &window.frame {
+            match frame.r#type {
+                FrameType::Range => write!(sql, " RANGE ").unwrap(),
+                FrameType::Rows => write!(sql, " ROWS ").unwrap(),
+            };
+            if let Some(end) = &frame.end {
+                write!(sql, " BETWEEN ").unwrap();
```
billy1624, line 1255:
```suggestion
                write!(sql, "BETWEEN ").unwrap();
```

I guess we have an excessive space before `BETWEEN`?

src/backend/query_builder.rs
```diff
@@ -1177,6 +1198,70 @@ pub trait QueryBuilder: QuotedBuilder {
...
+    /// Translate [`WindowStatement`] into SQL statement.
+    fn prepare_window_statement(
+        &self,
+        window: &WindowStatement,
+        sql: &mut SqlWriter,
+        collector: &mut dyn FnMut(Value),
+    ) {
+        let (first, exprs) = &window.partition_by;
+        write!(sql, " PARTITION BY ").unwrap();
+        self.prepare_simple_expr(first, sql, collector);
```
billy1624, line 1229: From what I understand the `PARTITION BY` is not mandatory?
For instance, we can write
```sql
SELECT salary, sum(salary) OVER (ORDER BY salary) FROM empsalary;
```
Please correct me if I'm wrong

billy1624:
Hey @ikrivosheev, thank you so much for the contributions!

REF:
- https://dev.mysql.com/doc/refman/8.0/en/window-function-descriptions.html
- https://www.sqlite.org/windowfunctions.html
- https://www.postgresql.org/docs/current/tutorial-window.html

src/backend/query_builder.rs
ikrivosheev: Yes, you are right. Done

ikrivosheev:
@billy1624 I have done work with WINDOW functions)

Cargo.toml
```diff
@@ -36,6 +36,7 @@ name = "sea_query"
 path = "src/lib.rs"
 
 [dependencies]
+either = "1.6.1"
```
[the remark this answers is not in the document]
ikrivosheev: Thank you for review. Done

src/query/select.rs
```diff
@@ -64,12 +63,20 @@ pub enum SelectDistinct {
...
 }
 
+/// Window type in [`SelectExpr`]
+pub enum WindowSelectType {
+    /// Name in [`SelectStatement`]
+    Name(DynIden),
+    /// Inline query in [`SelectExpr`]
+    Query(WindowStatement),
+}
+
```
tyt2y3, line 74: I like this. The type explains iteself

tyt2y3 (approved):

tyt2y3:
I think at a minimum sea-orm only needs to update the sea-query dependency (after we released a new sea-query version) and these features will be exposed in sea-orm.

PR #275, opened 2022-03-22
Implement sqlx::IntoArguments for Values

### LINKED ISSUE #273: Implement sqlx::IntoArguments for Values
## Summary
Implement the [`sqlx::IntoArguments`](https://docs.rs/sqlx/latest/sqlx/trait.IntoArguments.html) trait for `sea_query::Values`, for the various DB types (sqlite, posgres, ...)

## Motivation

When writing injection-free code with sqlx, the best would be to write in this style:

```
let (query, values) = Query::select()
    ...
    .cond_where(Expr::col(Users::UserId).eq(user_id)
    .build(SqliteQueryBuilder);
sqlx::query_with(query.as_str(), values).execute()
```

However this fails because `Values` doesn't implement `IntoArguments<'q, Sqlite>`. `Values` is just a `Vec<Value>`, and `sqlx::SqliteArguments<'q>` is just a `Vec<SqliteArgumentValue<'q>>`. The conversion is just mapping each value to an sqlite value, with the convenient `.add` that takes just about any argument.

To make it copy-free, we would also need a `.into_iter()` for `Values`.

Naturally, this would require all the correct compiler guards.

nitnelave:
Fixes #273

## PR Info

- Closes #273 

## Adds

Implements sqlx::IntoArguments for Values.
Implement IntoIterator for Values.

## Breaking Changes

It will break for users who have enabled an optional feature (e.g. with-chrono) and an sqlx backend.
To avoid that, we could add yet another feature to guard the conversion, but it would make the API less discoverable. On the other hand, the errors you get in the scenario above are very clear (it tells you which feature you need to enable).


nitnelave:
Small question: when writing a u64 to Postgres or Sqlite, since they don't support that format, I try to convert it to a i64 and unwrap if it fails. Would it perhaps be better to fail-fast by panicking for all inputs, forcing the user to do the conversion themselves?
Similarly, for Postgres, I promote small unsigned to bigger signed integers (because PG doesn't support unsigned). That will always work, but might not be ideal? Not sure about that one, if we end up serializing that to a string it shouldn't matter. It would be a bit annoying on the user side to be forced to convert everything to signed just because, even though it works like that.

nitnelave:
Example usage in a real project: https://github.com/nitnelave/lldap/pull/127

Sytten:
This is a big shift from the macro (https://github.com/SeaQL/sea-query/blob/master/src/driver/sqlx_sqlite.rs) that was used previously to generate that. 

Like I suggested a while ago I think those integrations should be in different crates then you can greatly simplify the number of features and stay consistent between this and the bind macro.

Different crates would also allow me to write the diesel integration.

nitnelave:
One problem with a different crate is that you can't condition on the sea query features that are enabled (or can you? I didn't find out how to do that for sqlx here). That means that you'll have to repeat the same feature flags for the other crate.

Here we can remove all the feature flags that I added with their sqlx versions, you'll just get a not very clear error message if you enable a feature for sea query but not sqlx.

How would the conversion look like in the other crate? Adding a `.to_sqlx` method to Values, via a trait?

Anyway, we should at least keep the `IntoIterator` part of this PR, it can even be used in the existing macros to avoid some copies.

Sytten:
If we do a separate crate I think its fair to have duplicate features and that a feature X can also enable it on sea-query (like you already did) so you dont have to specify it twice.

I mean yeah that could work too. I am just trying to think on what is more confusing for the user. I think that if you enable sqlx-mysql and with-chrono on sea-query its fair to assume that you need to enable chrono on the sqlx dependency too. Documentation can help on that.

Hum yeah this is more annoying and one major point in favor of not splitting in another crate. Maybe then its a matter of splitting it in modules so they are not all in one giant file and its easy to enable/disable each flavour. I do plan on bringing my diesel implementation in the same format so lets try to make that work.

tyt2y3:
> Anyway, we should at least keep the IntoIterator part of this PR, it can even be used in the existing macros to avoid some copies.

Agree.

The reason I don't want to include sqlx as a dependency is because yeah, it forces us to select a runtime and that's not where sea-query should do. Ideally, sqlx (like postgres) would split out a `types` crate so that we can only depend on the pure type definitions.

I also can't think of a better idea than moving this `IntoArguments` feature to another sub-crate.

nitnelave:
> The reason I don't want to include sqlx as a dependency is because yeah, it forces us to select a runtime 
Only for tests. Otherwise, the only requirement is that _in the final binary_, a runtime is selected. It's still up to the user to choose what runtime they want. If we have a non-default feature like I added there, it's not a problem (and I only added it to make --all-features work, otherwise it's really not needed).

Yeah, ideally sqlx would split out the core types, but that's non-trivial work.

Moving `IntoArguments` into a sub-crate makes the interface less friendly, in my opinion.

nitnelave:
Another advantage of `IntoArguments` vs the macros is that it supports dynamic DB backend, the `Any` type of sqlx. It delays the conversion until sqlx calls it, at which point they know what backend is actually used.

nitnelave:
Hmm, no, actually this won't help with the argument resolution, we'd still have to implement it for Any as well.

Anyway, what's the decision on this? Do we want to cut this back to only IntoIterator, or keep it like this?

Sytten:
I think this needs to wait on the merge of https://github.com/SeaQL/sea-query/pull/292

ikrivosheev:
@Sytten I agree. @nitnelave If you need help adapting your PR, I'm ready to help.

nitnelave:
Thanks for the offer @ikrivosheev ! I should be able to handle it myself, once I understand what I'm adapting to :) What do we want to implement? Are we still on board for the intoargument trait?

Sytten:
With rust 1.60 we can now do `crate?/feature` which is ideal for the use case we have here. Though requiring 1.60 would be a bit brutal. @tyt2y3 @billy1624

ikrivosheev:
@nitnelave I like idea @tyt2y3 about: don't bring sqlx into sea-query. Two variants:
1) move into other subcrate
2) move into drivers

But I don't understand what is different between this and current driver implants?

nitnelave:
The problem is that we can't implement a trait on `Values` in another crate. The only possibility would be that we somehow extract the entire `Values` struct (and dependencies) to another crate, and implement the trait there, but that's still ugly (sea-query depends on that type definition, which depends on sqlx for the trait).

Note that the sqlx dependency is there only if you enable the sqlx feature.

The reason why I consider this approach superior to drivers is:
 - The syntax is nicer, it's just "normal" sqlx syntax, with no macro involved.
 - It works with Sqlx's expected types and traits directly, instead of working around them with multiple calls to `bind`
 - It works with generics: if the function called works on a `Pool<DB>`, then implementing the trait will make sure that the correct `DB` is picked, whereas with macros you have to specify directly which `DB` you're using.

nitnelave:
One idea I can come up with is to have a sub crate that implements a new trait on the various queries, adding a `.build_sqlx` method instead of build, that returns a wrapper type around `Values` with `IntoArguments` implemented.

It's kinda hard to discover though, in that case we should probably add some comments here around the sqlx feature referencing that crate.

Sytten:
I would be in favour of moving values to the driver crate, maybe it could be renamed to better express that.

nitnelave:
That doesn't really solve the issue of depending on sqlx: sea-query depends on the driver crate, which depends on sqlx. Plus you'll have to move all the dependencies of Value to that crate as well. I don't think it would bring much clarity.

ikrivosheev:
@nitnelave, no, sea-query-driver doesn't depend on sqlx...

ikrivosheev:
But I like your idea and the interface is convenient.

Sytten:
Agreed but it is a solved problem on 1.60 though as you can mix optional dependencies with features so its just a matter of waiting a couple of weeks for people to upgrade IMO.

nitnelave:
> @nitnelave, no, sea-query-driver doesn't depend on sqlx...

Well, not now, but if we implement `IntoArguments` for `Values` in sea-query-driver, then it will (to the same extent that we were talking about sea-query depending on sqlx). The dependency doesn't really change, it's just hidden in a subcrate that we depend on.

billy1624:
We could implement `sqlx::IntoArguments` inside `sea-query-driver` with macros, which means `sea-query-driver` doesn't depends on `sqlx`. The macros will define its own `Values` which wrap the `sea_query::Values`.

```rs
pub struct Values(pub sea_query::Values);
```

Then, `IntoIterator` and `sqlx::IntoArguments` can be implemented as follows

```rs
impl IntoIterator for Values {
    type Item = sea_query::Value;
    type IntoIter = std::vec::IntoIter<Self::Item>;

    fn into_iter(self) -> Self::IntoIter {
        self.0.into_iter()
    }
}

impl<'q> sqlx::IntoArguments<'q, sqlx::postgres::Postgres> for Values {
    ...
}
```

nitnelave:
@billy1624 but you still need the macros to wrap the values. Can we merge this with the idea I had above, implementing a new method (via a trait) on the queries that pre-wraps the values?

tyt2y3:
> @billy1624 but you still need the macros to wrap the values. Can we merge this with the idea I had above, implementing a new method (via a trait) on the queries that pre-wraps the values?

I don't 100% understand this approach.

> We could implement sqlx::IntoArguments inside sea-query-driver with macros, which means sea-query-driver doesn't depends on sqlx. The macros will define its own Values which wrap the sea_query::Values.

I guess this is the same as what we are doing with Rusqlite now.

nitnelave:
@tyt2y3 it would be a mix of my external crate proposal and @billy1624 's proposal: the driver macros introduce a new wrapping Values type and implement IntoArgument for it. And in addition, they create a new trait and implement it for all queries, that adds a `.build-sqlx` method which calls `.build` and wraps the returned Values in the new type. That way from outside you just have to use the new trait, call build_sqlx and everything works fine.

tyt2y3:
> build_sqlx and everything works fine

Sounds like an intricate plan. But that's our end goal yeah. As long as 1) it does not affect existing users 2) is opt-in we will merge it and see how it works in action, and might actually replace our current driving mechanism (which is a bit hacky IMO)

Sytten:
That way it should also be possible to create a build_diesel too from my existing code. I will align with your PR once its merged.

nitnelave:
Okay, I had a closer look at how the drivers were implemented and used, and in the end I don't think it's necessary to use macros.
What I propose:
 - Add a new crate, that depends on both sea-query and (feature guarded) sqlx.
 - For this crate, require Rust 1.60 to allow checking for dependencies' features.
 - This create creates the new `Values` type and the trait adding the `.build_sqlx` method, and implements `IntoIterator` for `Values`.
 - That way, users of sea-query only have to import a trait from this crate and use the new function rather than invoke a macro that injects a module into their code.

WDYT?

nitnelave:
I pushed a proof of concept, tell me what you think and I'll finish implementing it.

billy1624 (commented):
The `binders` approach looks good!
I think it's better than injecting code with macros in user's project

Sytten:
Is there any reason not to move values in this binders crate directly? Otherwise I would name the wrapper BindValue or something to differentiate.

nitnelave:
I'm not sure how it would work to put `Values` in this crate: the various `Queries` need to return it, so they need to see it, and we would end up pulling a sizeable chunk of code into the crate.

I'll rename them to `SqlxValues` & co, then.

nitnelave:
One difference between this and the driver approach (more of a difference of choice): the drivers "support" the time arguments for sqlite by calling `time_as_naive_utc_in_string`. But that's going to fail on the way back, decoding the time argument. If the support is not built into `sqlx`, I'm of the opinion that it's up to the user to make a well-behaved wrapper or find an alternative.

Sytten:
Agreed that is what I was suggesting, a `BindValues` and modifying the select trait implementation so its always the same function name to convert to this type.

nitnelave:
Any suggestion for the function name? `build_to_bind`? `build_bind_values`?

Sytten:
Hum no particular idea, just throwing a couple more in the hat: `build_binding`, `build_bind`, `build_binder`.
Heck even just `bind` could be good maybe?

nitnelave:
@billy1624 @tyt2y3 any opinion on the `Values`/`BindValues`/`build_bindable` issue? Or the PR as a whole?

billy1624 (commented):
Thanks! @nitnelave

sea-query-binders/src/sqlx_sqlite.rs
```diff
@@ -0,0 +1,104 @@
...
+                    args.add(uuid.map(|uuid| *uuid));
+                }
+                #[cfg(feature = "with-rust_decimal")]
+                Value::Decimal(_) => {
+                    panic!("Sqlite doesn't support decimal arguments");
+                }
+                #[cfg(feature = "with-bigdecimal")]
+                Value::BigDecimal(_) => {
+                    panic!("Sqlite doesn't support bigdecimal arguments");
+                }
```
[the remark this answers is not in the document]
nitnelave: Done

sea-query-binders/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+categories = [ "database" ]
+keywords = [ "database", "sql", "mysql", "postgres", "sqlite" ]
+
+[lib]
+
+[dependencies]
+sea-query = { version = "^0.24", path = "../" }
+sqlx = { version = "^0.5.6", optional = true }
+
+[features]
```
ikrivosheev, line 19: Please, add support for: `with-ipnetwork` and `with-mac_address`

ikrivosheev (commented):
@nitnelave thank you for contribution! Some comments

sea-query-binders/Cargo.toml
nitnelave: Done.

examples/sqlx_any/Cargo.toml
```diff
@@ -0,0 +1,54 @@
...
+    "sqlx-sqlite",
+    "sqlx-mysql",
+    "sqlx-any",
+    "with-chrono",
+    "with-json",
+    "with-rust_decimal",
+    "with-bigdecimal",
+    "with-uuid",
+    "with-time",
+] }
```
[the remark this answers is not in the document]
Sytten: Shouldnt this be the inverse? Reexport the binder from sea query based on features?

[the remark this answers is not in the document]
nitnelave: Sea-query cannot re-export the binders because it cannot depend on this crate, the dependency is the opposite.

nitnelave:
@ikrivosheev @billy1624 Any other comments?

ikrivosheev (approved):
@nitnelave, thank you! Sory for delay. LGTM!

nitnelave:
Thanks for the review! Can we merge this, then?

sea-query-binders/Cargo.toml
```diff
@@ -0,0 +1,34 @@
...
+sqlx-mysql = ["sqlx/mysql", "sea-query/sqlx-mysql"]
+sqlx-postgres = ["sqlx/postgres", "sea-query/sqlx-postgres"]
+sqlx-sqlite = ["sqlx/sqlite", "sea-query/sqlx-sqlite"]
+sqlx-any = ["sqlx/any"]
+with-chrono = ["sea-query/with-chrono"]
+with-json = ["sea-query/with-json"]
+with-rust_decimal = ["sea-query/with-rust_decimal"]
+with-bigdecimal = ["sea-query/with-bigdecimal"]
+with-uuid = ["sea-query/with-uuid"]
+with-time = ["sea-query/with-time"]
```
billy1624, line 29: Hey @nitnelave, sorry for the delay. All in all it looks good! :D

Should we enable the corresponding SQlx features as well?

```suggestion
with-chrono = ["sqlx/chrono", "sea-query/with-chrono"]
with-json = ["sqlx/json", "sea-query/with-json"]
with-rust_decimal = ["sqlx/decimal", "sea-query/with-rust_decimal"]
with-bigdecimal = ["sqlx/bigdecimal", "sea-query/with-bigdecimal"]
with-uuid = ["sqlx/uuid", "sea-query/with-uuid"]
with-time = ["sqlx/time", "sea-query/with-time"]
```

nitnelave: Good call. I added that, rebased and squashed to clean up the history.

sea-query-binders/src/sqlx_postgres.rs
```diff
@@ -0,0 +1,113 @@
...
+                Value::BigDecimal(d) => {
+                    args.add(d.as_deref());
+                }
+                #[cfg(feature = "with-json")]
+                Value::Json(j) => {
+                    args.add(j.as_deref());
+                }
+                #[cfg(feature = "postgres-array")]
+                Value::Array(_) => {
+                    panic!("SeaQuery doesn't support array arguments for Postgresql");
```
ikrivosheev, line 107: Why `SeaQuery`?) Maybe `sqlx`?

nitnelave: So, `sqlx`  supports arrays for postgres, but the conversion is non-trivial.
Here, we would have to make sure that all the arguments are the same type, and convert them all to a static type before calling `args.add`.

Let's do that in a follow-up PR, maybe?

ikrivosheev: Hmmm, I understand. Can you convert this comment to issues?

nitnelave: Created https://github.com/SeaQL/sea-query/issues/336

nitnelave:
@tyt2y3 when you have time for a review :)

nitnelave:
Any chance of a review?

tyt2y3:
I am about to make some changes and merge into master

Issue #158, opened 2021-10-07
lower(col) and upper(col) support


ayyEve:
I'm looking for support for LOWER(col) for queries, I'm not sure if this already exists but I was unable to find it.

I'm thinking the usage should be as follows:
```rs
sea_query::lower(table::Column::Name).eq(some_string);
// or this, which might be easier to code
table::Column::Name.lower().eq(some_string);
```

I personally prefer the 2nd option, and I believe it would be easier to add.
Let me know if you need anything, or if anything is unclear'

Thanks

PR #276, opened 2022-03-22
issues-158 Add LOWER and UPPER func

ikrivosheev:
## PR Info

- Closes <https://github.com/SeaQL/sea-query/issues/158>

## Adds

Add support for `LOWER` and `UPPER` functions

For add support expression, like: `table::Column::Name.lower().eq(some_string);` i does not understand: the expression must be `Expr::tbl` or `Expr::column`? I think it's better to make a more explicit use for functions.


tyt2y3 (approved):
Awesome, thank you!

PR #277, opened 2022-03-22
issues-213 Add support for multiple ALTER table

ikrivosheev:
## PR Info

- Closes <https://github.com/SeaQL/sea-query/issues/213>

## Adds

Add support:

```
let table = Table::alter()
    .table(Alias::new("table"))
    .add_column(ColumnDef::new(Alias::new("new_col")).integer())
    .add_column(ColumnDef::new(Alias::new("new_col2")).integer())
    .to_owned();
```

For MySQL and PostgreSQL generate:
```
ALTER TABLE tblName
    ADD COLUMN new_col INTEGER,
    ADD COLUMN new_col2 INTEGER
```

For SQLLite should panic.

Issue #229, opened 2022-01-12
support ifnotexist(column) add column


conanchen:
does sea-query support alter table by checking the column if exists or not before add_column?

let table = Table::alter()
    .table(Font::Table)
. ifexists("column")
    .add_column(

PR #278, opened 2022-03-22
issues-229 add column if not exists

ikrivosheev:
## PR Info

- Closes <https://github.com/SeaQL/sea-query/issues/229>

- Dependencies:
  - https://github.com/SeaQL/sea-query/pull/277

## Adds

Add support for query like: 
```
ALTER TABLE name
    ADD COLUMN IF NOT EXISTS column_name data_type
```

for PostgreSQL and for MySQL. For SQLLite this is ignored

### DIFF
--- a/src/backend/mysql/table.rs
+++ b/src/backend/mysql/table.rs
@@ -134,8 +134,11 @@ impl TableBuilder for MysqlQueryBuilder {
                 write!(sql, ", ").unwrap();
             };
             match option {
-                TableAlterOption::AddColumn(column_def) => {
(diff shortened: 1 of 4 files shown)


tyt2y3:
I don't think this works for MySQL, may be just MariaDB?

tyt2y3 (changes requested):
I like the addition of this API. But a `bool` does not seem forword looking, we might want to create a new type `AddColumnOption` (or some reasonable name)?

ikrivosheev:
@tyt2y3 I have done. Can you review it please?)

tyt2y3 (approved):

Issue #285, opened 2022-03-30
ColumnDef `timestamp_with_time_zone_len` generates invalid SQL for PostgreSQL


scootermon:
## Description

As indicated in the title, the generated SQL for `timestamp_with_time_zone_len` columns isn't valid as the precision "argument" is placed after 'with time zone` rather than after 'timestamp'.
For reference, see: <https://www.postgresql.org/docs/current/datatype-datetime.html#DATATYPE-DATETIME-TABLE>

## Steps to Reproduce

Run the following migration:

```rust
use sea_schema::migration::prelude::*;

#[derive(Iden)]
enum Foo {
    Table,
    CreatedAt,
}

pub struct Migration;

impl MigrationName for Migration {
    fn name(&self) -> &str {
        "bug-report"
    }
}

#[async_trait::async_trait]
impl MigrationTrait for Migration {
    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
        manager
            .create_table(
                Table::create()
                    .table(Foo::Table)
                    .col(ColumnDef::new(Foo::CreatedAt).timestamp_with_time_zone_len(0))
                    .to_owned(),
            )
            .await
    }

    async fn down(&self, manager: &SchemaManager) -> Result<(), DbErr> {
        // not relevant
        Ok(())
    }
}
```

### Expected Behavior

Should generate and execute the following SQL:

```sql
CREATE TABLE "foo" (
  [...]
  "created_at" timestamp(0) with time zone,
)
```

### Actual Behavior

```sql
CREATE TABLE "foo" (
  [...]
  "created_at" timestamp with time zone(0),
)

Execution Error: error returned from database: syntax error at or near "("
```

## Versions

```
│   └── sea-orm v0.7.1
│       ├── sea-orm-macros v0.7.0 (proc-macro)
│       ├── sea-query v0.23.0
│       │   ├── sea-query-derive v0.2.0 (proc-macro)
│       ├── sea-strum v0.23.0
│       │   └── sea-strum_macros v0.23.0 (proc-macro)
│   └── sea-schema v0.7.1
│       ├── sea-orm v0.7.1 (*)
│       ├── sea-query v0.22.0
│       │   └── sea-query-derive v0.2.0 (proc-macro) (*)
│       ├── sea-schema-derive v0.1.0 (proc-macro)
```

[...]

PR #286, opened 2022-03-30
issues-285 Fix timestamp_with_time_zone_len

ikrivosheev:
## PR Info

- Closes <https://github.com/SeaQL/sea-query/issues/285>

### Befor fix

```sql
CREATE TABLE "foo" (
  [...]
  "created_at" timestamp with time zone(0),
)
```
Execution Error: error returned from database: syntax error at or near "("

### After fix

```sql
CREATE TABLE "foo" (
  [...]
  "created_at" timestamp(0) with time zone,
)
```


tyt2y3:
Great work. Thanks alot!

Issue #142, opened 2021-09-26
Support `SKIP LOCKED` for `SelectStatement`


Stumblinbear:
I'm working on implementing a job queue using Postgres, and I need to be able to add `SKIP LOCKED` to my select queries

tyt2y3:
Is it an extension to https://github.com/SeaQL/sea-query/pull/117

Stumblinbear:
Actually, no, I don't think so. With Postgres, all you need to do is add SKIP LOCKED onto a SELECT query, and it'll skip any row currently locked by a transaction. So all I have to do is: `DELETE FROM "jobs" WHERE "jobs"."id" = (SELECT "id" FROM "jobs" WHERE "execute_at" < $1 ORDER BY "id" ASC FOR UPDATE SKIP LOCKED LIMIT 1) RETURNING "type_of", "data"` inside of a transaction, do the work, then commit the transaction. Any other server looking for jobs will skip over any currently running jobs, and if the transaction fails or the client gets disconnected, the row gets unlocked.

I'm currently using the transaction branch (even with the lifetime issue (which I haven't run into), it's working great).

PR #289, opened 2022-03-30
issues-142 full support lock in select

ikrivosheev:
## PR Info

- Closes <https://github.com/SeaQL/sea-query/issues/142>

## Adds

Add full support for `LOCK` in `SELECT` statement.

References:

* https://www.postgresql.org/docs/current/sql-select.html#SQL-FOR-UPDATE-SHARE
* https://dev.mysql.com/doc/refman/8.0/en/select.html

### DIFF
--- a/src/backend/query_builder.rs
+++ b/src/backend/query_builder.rs
@@ -391,19 +391,37 @@ pub trait QueryBuilder: QuotedBuilder {
     /// Translate [`LockType`] into SQL statement.
     fn prepare_select_lock(
         &self,
-        select_lock: &LockType,
+        lock: &Lock,
         sql: &mut SqlWriter,
-        _collector: &mut dyn FnMut(Value),
+        collector: &mut dyn FnMut(Value),
     ) {
         write!(
             sql,
-            "{}",
-            match select_lock {
-                LockType::Shared => "FOR SHARE",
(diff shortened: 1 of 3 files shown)


tyt2y3:

So old enum corresponds to the new enum by:

`Exclusive` -> `Update`
`Shared` -> `Share`

am I right?

tyt2y3:
https://github.com/SeaQL/sea-query/pull/117

ikrivosheev:
@tyt2y3 yes, you are right, but i save old name for functions in `SelectStatement`

nappa85:
> #117 @nappa85 request for your review :) we are proposing a (breaking) name change here what was the reasoning behind the `Exclusive` name?

can't remember, probably some docs called it that way, in facts mysql calls it that way https://dev.mysql.com/doc/refman/8.0/en/innodb-locking.html#innodb-shared-exclusive-locks

ikrivosheev:
@tyt2y3 @nappa85 I tried to follow the logic of the library: library names  ~ SQL names and therefore changed a little names in `Enum`. If i can make backwards incompatible changes, I can rename functions in `SelectStatement`.

tyt2y3:
Thanks. I think I edited some docs and added to release notes

Issue #280, opened 2022-03-27
The trait is not implemented fo


Marekuss:
## Description
Hello guys, im not able to compile minimal example (below) with rusqlite backend, where im getting follow error:
```
error[E0277]: the trait bound `ValueRef<'_>: From<&bool>` is not satisfied
   --> src/main.rs:4:1
    |
4   | sea_query::sea_query_driver_rusqlite!();
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `From<&bool>` is not implemented for `ValueRef<'_>`
    |
    = help: the following implementations were found:
              <ValueRef<'a> as From<&'a [u8]>>
              <ValueRef<'a> as From<&'a rusqlite::types::Value>>
              <ValueRef<'a> as From<&'a str>>
              <ValueRef<'a> as From<Option<T>>>
    = note: required because of the requirements on the impl of `Into<ValueRef<'_>>` for `&bool`
    = note: required because of the requirements on the impl of `From<&bool>` for `ToSqlOutput<'_>`
note: required by `from`
    = note: this error originates in the macro `opt_string_to_sql` (in Nightly builds, run with -Z macro-backtrace for more info)
```

## Steps to Reproduce

1. run `cargo init --bin`
2. modify Cargo.toml
```
[package]
name = "sea_query_test"
version = "0.1.0"
edition = "2021"

[dependencies]
sea-query = { version = "^0", features = [
    "rusqlite"
]}

[dependencies.rusqlite]
version = "^0.27"
features = [
    "bundled"
]
```

3. modify main.rs
```
use rusqlite::{Connection, Result};
use sea_query::*;

sea_query::sea_query_driver_rusqlite!();
use sea_query_driver_rusqlite::RusqliteValues;

fn main() -> Result<()> {
    println!("test");
    Ok(())
}
```
4. run `cargo update && cargo build`

## Versions

Linux: Linux user-server 5.4.0-99-generic #112-Ubuntu SMP Thu Feb 3 13:50:55 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
rustc: 1.57.0
cargo: 1.57.0

## Additional Information
cargo tree | grep sea-
```
└── sea-query v0.23.0
    └── sea-query-derive v0.2.0 (proc-macro)
```

cargo tree
```
sea_query_test v0.1.0 (/home/user/Temp/sea_query_test)
├── rusqlite v0.27.0
[...]

PR #292, opened 2022-04-01
issues-280 Rewrite drivers using proc-macro

ikrivosheev:
## PR Info

- Closes <https://github.com/SeaQL/sea-query/issues/280>

## Fixes

Rewrite drivers for: `rusqlite`, `sqlx_*` from `macro_rules` to `proc-macro`. In `proc-macro` we can use `cfg` instead of blank method for types, which enable by crate features and fix: https://github.com/SeaQL/sea-query/issues/280

I add `cfg` for implementation methods which enable by feature like: `with-*`.

## Breaking Changes

1. Remove methods `Value::is_json` and `Value::as_ref_json` when feature: **with-json** is disabled
2. Remove methods `Value::is_time_*` and `Value::as_ref_time_*` when feature: **with-time** is disabled
3. Remove methods `Value::is_chrono_*` and `Value::as_ref_chrono*` when feature: **with-chrono** is disabled
4. Remove methods `Value::is_decimal`, `Value::as_ref_decimal` and `Value::decimal_to_f64` when feature: **with-rust_decimal** is disabled
5. Remove methods `Value::is_big_decimal`, `Value::as_ref_big_decimal` and `Value::big_decimal_to_f64` when feature: **with-bigdecimal** is disabled
6. Remove methods `Value::is_uuid` and `Value::as_ref_uuid` when feature: **with-uuid** is disabled
7. Remove methods `Value::is_array` and `Value::as_ref_array` when feature: **postgres-array** is disabled

### DIFF
--- a/src/table/alter.rs
+++ b/src/table/alter.rs
@@ -134,7 +134,7 @@ impl TableAlterStatement {
     ///
     /// assert_eq!(
     ///     table.to_string(MysqlQueryBuilder),
-    ///     r#"ALTER TABLE `font` ADD COLUMN `new_col` int NOT NULL DEFAULT 100"#
+    ///     r#"ALTER TABLE `font` ADD COLUMN IF NOT EXISTS `new_col` int NOT NULL DEFAULT 100"#
     /// );
     /// assert_eq!(
     ///     table.to_string(PostgresQueryBuilder),


tyt2y3:
Thank you. This looked extremely brilliant!
Does that mean the examples stay the same with no change?
Btw we might want to add it to the workspace? https://github.com/SeaQL/sea-query/blob/master/Cargo.toml#L2
The only nitpick might be.. can we name the crate `sea-query-drivers` ?

ikrivosheev:
@tyt2y3 thank you for review. Done, renamed!

Yes, all examples work as before. But, for example, if feature: `with-json` does not enable, `Value` does not have methods: `is_json`, `as_json_ref`.

ikrivosheev:
@tyt2y3 i rebase branch with master.

tyt2y3:
Thank you for the explanation. I think it also obeys to the rule of "feature addition"

ikrivosheev:
@tyt2y3 what should i do?

tyt2y3:
Oh, nothing more has to be done. I was thinking the scenario where two downstream creates each with a different feature set.

ikrivosheev:
I have seen similar approach in other libraries (for example [syn](https://docs.rs/syn/latest/syn/), when features add new functions and methods into crate and types. It is very comfortable and works well in rust.

ikrivosheev:
@billy1624 @tyt2y3 I have corrected the description

PR #300, opened 2022-04-09
Type definition auto-generation

### LINKED ISSUE #296: Procedural macros for generating a Iden enum from structs
## Summary

I recently made a crate in which it auto-generates an enum-based type definitions from structs, and I think it would be useful to have a similar feature in the official crate. I think it would be useful to also have a similar feature for creating tables and don't needing to provide all columns manually (such as a `AllColumn` helper).

I can implement this feature by myself by opening a pull request, if everyone agrees to it.

https://github.com/eexsty/squtils
**Example**
```rust
#[generate_type_def]
struct Foo {
    bar: String,
    baz: i32,
}

/// This is the same as doing:
/// struct Foo {
///    bar: String,
///    baz: i32,
/// }
/// #[derive(Iden)]
/// enum FooTypeDef {
///   Table,
///   Bar,
///   Baz
/// }
```

## Motivation

Avoid having to declare the same fields in a enum and struct, this would generate it automatically for you, this would avoid people forgetting to add field in both enum and struct.

eexsty:
## PR Info

- Closes #296

## Adds

A crate in which it auto-generates an enum-based, Iden-deriving type definitions from structs.

### DIFF
--- a/sea-query-attr/src/lib.rs
+++ b/sea-query-attr/src/lib.rs
@@ -45,10 +45,13 @@ pub fn gen_type_def(_args: TokenStream, input: TokenStream) -> TokenStream {
         input.ident.span(),
     );
     
+    let original_name = &input.ident;
     let struct_name = quote::format_ident!("{}TypeDef", &input.ident);
     let pascal_def_names = field_names.iter().map(|field| &field.pascal);
     let pascal_def_names2 = pascal_def_names.clone(); // we can't repeat the same ident twice in a quote!, so we need to clone the first one
+    let pascal_def_names3 = pascal_def_names.clone(); 
     let default_names = field_names.iter().map(|field| &field.default);
+    let default_names2 = default_names.clone();
 
     TokenStream::from(quote! {
         #input
@@ -65,6 +68,15 @@ pub fn gen_type_def(_args: TokenStream, input: TokenStream) -> TokenStream {
                     #struct_name::Table => stringify!(#table_name),
                     #(#struct_name::#pascal_def_names2 => stringify!(#default_names)),*
                 }).unwrap();
+            }
+        }
+
+        impl #struct_name {
+            pub fn value_for<I>(&self, entity: &#original_name) -> I {
+                match self {
+                    #(#struct_name::#pascal_def_names3 => entity.#default_names2 as I),*
+                    #struct_name::Table => unreachable!()
+                }
             }
         }
     })
--- a/sea-query-attr/src/lib.rs
+++ b/sea-query-attr/src/lib.rs
@@ -6,9 +6,6 @@ use syn::{
     Ident,
 };
 
-#[macro_use]
-extern crate quote;
-
 struct NamingHolder {
     pub default: Ident,
     pub pascal: Ident,
@@ -79,7 +76,7 @@ pub fn gen_type_def(args: TokenStream, input: TokenStream) -> TokenStream {
     let pascal_def_names2 = pascal_def_names.clone();
     let default_names = field_names.iter().map(|field| &field.default);
 
-    TokenStream::from(quote! {
+    TokenStream::from(quote::quote! {
         #input
 
         #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
--- a/sea-query-attr/src/lib.rs
+++ b/sea-query-attr/src/lib.rs
@@ -1,21 +1,45 @@
+use darling::FromMeta;
 use heck::{ToPascalCase, ToSnakeCase};
 use proc_macro::TokenStream;
-use syn::{parse_macro_input, spanned::Spanned, Data, DataStruct, DeriveInput, Fields, Ident};
-
-extern crate proc_macro;
+use syn::{
+    parse_macro_input, spanned::Spanned, AttributeArgs, Data, DataStruct, DeriveInput, Fields,
+    Ident,
+};
 
 #[macro_use]
 extern crate quote;
-extern crate syn;
 
 struct NamingHolder {
     pub default: Ident,
     pub pascal: Ident,
 }
 
+#[derive(Debug, FromMeta)]
+struct GenTypeDefArgs {
+    #[darling(default)]
+    pub prefix: Option<String>,
+    #[darling(default)]
+    pub suffix: Option<String>,
+}
+
+const DEFAULT_PREFIX: &'static str = "";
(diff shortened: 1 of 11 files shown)


tyt2y3:
Thank you for the contribution. Is this the final API?
```rust
#[generate_type_def]
struct Foo {
    bar: String,
    baz: i32,
}
```

```rust
// This is the same as doing:
struct Foo {
   bar: String,
   baz: i32,
}
#[derive(Iden)]
enum FooTypeDef {
  Table,
  Bar,
  Baz
}
```

If so we'd appreciate some tests under `sea-query-derive/tests` and probably a bit of docs

sea-query-attr/src/lib.rs
```diff
@@ -0,0 +1,74 @@
+use heck::{ToPascalCase, ToSnakeCase};
+use proc_macro::TokenStream;
+use syn::{parse_macro_input, spanned::Spanned, Data, DataStruct, DeriveInput, Fields, Ident};
+
+extern crate proc_macro;
```
ikrivosheev, line 5: You can remove `extern crate` after Rust 2018. in most cases [you won't need to use extern crate anymore](https://doc.rust-lang.org/edition-guide/rust-2018/module-system/path-clarity.html).

sea-query-attr/src/lib.rs
```diff
@@ -0,0 +1,74 @@
...
+#[macro_use]
+extern crate quote;
+extern crate syn;
+
+struct NamingHolder {
+    pub default: Ident,
+    pub pascal: Ident,
+}
+
+#[proc_macro_attribute]
```
ikrivosheev, line 16: Can you add two attributes: `prefix` and `suffix` for result `enum`? I think this is very useful. @tyt2y3 what do you think?

sea-query-attr/src/lib.rs
```diff
@@ -0,0 +1,74 @@
...
+    TokenStream::from(quote! {
+        #input
+
+        #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+        enum #struct_name {
+            Table,
+            #(#pascal_def_names,)*
+        }
+
+        impl sea_query::Iden for #struct_name {
```
ikrivosheev, line 65: In the driver, we have added the possibility of a prefix for import? I think it would be very usefull here too.

src/lib.rs
```diff
@@ -752,4 +752,7 @@ pub use value::*;
 #[cfg(feature = "derive")]
 pub use sea_query_derive::Iden;
 
+#[cfg(feature = "attr")]
```
ikrivosheev, line 755: I don't see you adding this future to `Cargo.toml`, can you add?

ikrivosheev (commented):
@eexsty thank you for contribution! Can you add tests and documentation?

sea-query-attr/src/lib.rs
ikrivosheev: Or you can generate submodule, but how it will work with multiple struct...

eexsty: > @tyt2y3 what do you think?

Sure

eexsty:
> Thank you for the contribution. Is this the final API?
> 
> ```rust
> #[generate_type_def]
> struct Foo {
>     bar: String,
>     baz: i32,
> }
> ```
> 
> This generates:
> 
> ```rust
> struct Foo {
>    bar: String,
>    baz: i32,
> }
> #[derive(Iden)]
> enum FooTypeDef {
>   Table,
>   Bar,
>   Baz
> }
> ```
> 
> If so we'd appreciate some tests under `sea-query-derive/tests` and probably a bit of inline docs

Yes

src/lib.rs
eexsty: > 

image
Already added

ikrivosheev: Oh sorry, I missed.

eexsty:
I think it is done.

sea-query-attr/src/lib.rs
```diff
@@ -0,0 +1,100 @@
+use darling::FromMeta;
+use heck::{ToPascalCase, ToSnakeCase};
+use proc_macro::TokenStream;
+use syn::{
+    parse_macro_input, spanned::Spanned, AttributeArgs, Data, DataStruct, DeriveInput, Fields,
+    Ident,
+};
+
+#[macro_use]
+extern crate quote;
```
ikrivosheev, line 10: Why needed this?

sea-query-attr/src/lib.rs
ikrivosheev: @eexsty, please look at the moment

sea-query-attr/src/lib.rs
eexsty: For using the quote macro.

ikrivosheev: You can write: `use qoute::qoute`

eexsty:
I think everything is alright now

billy1624 (commented):
Hi there, thanks for the contributions! @eexsty

sea-query-attr/src/lib.rs
```diff
@@ -0,0 +1,97 @@
...
+    TokenStream::from(quote::quote! {
+        #input
+
+        #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+        enum #enum_name {
+            Table,
+            #(#pascal_def_names,)*
+        }
+
+        impl sea_query::Iden for #enum_name {
```
ikrivosheev, line 88: @eexsty https://github.com/SeaQL/sea-query/pull/300#discussion_r851996712

eexsty: I don't understand what you mean by 'prefix for imports', could you explain?

tyt2y3: oh, here you see that sea_query::Iden has no prefix. And we might want to customize the prefix with something like `$another_crate::sea_query::Iden`

eexsty: Interesting, I'll add it right now

ikrivosheev (approved):
@tyt2y3 LGTM! 

@eexsty Thank you!

tyt2y3:
I released https://crates.io/crates/sea-query-attr
I renamed a few things hope you wouldn't mind. Thank you again!

eexsty:
@tyt2y3 The enum seems to be not exported as `pub`, could you please fix it?

tyt2y3:
Done

PR #301, opened 2022-04-09
CTE optional columns

npapapietro:
## PR Info

The definition of Common Table Expressions (CTE) allow for optional column definition, allowing for using the table
name as the identifier for columns.:

```sql
-- Should be equal
-- Case 1
WITH "cte" ("id") AS (SELECT "tbl"."id" FROM "tbl")
SELECT "cte"."id" FROM "cte";
-- Case 2
WITH "cte" AS (SELECT "tbl"."id" FROM "tbl")
SELECT "cte"."id" FROM "cte";
```

## Fixes

Currently, the query building will squish the cte name against the `AS` keyword like:
```sql
WITH "cte"AS (Select ...
```
so I added a simple fix to add a space:
```sql
WITH "cte" AS (Select ...
```

Additionally I added simple unit test to illustrate.

### DIFF
--- a/src/backend/query_builder.rs
+++ b/src/backend/query_builder.rs
@@ -804,7 +804,9 @@ pub trait QueryBuilder: QuotedBuilder {
     ) {
         cte.table_name.as_ref().unwrap().prepare(sql, self.quote());
 
-        if !cte.cols.is_empty() {
+        if cte.cols.is_empty() {
+            write!(sql," ").unwrap();
+        } else {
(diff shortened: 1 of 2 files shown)


ikrivosheev:
@npapapietro thank you for the PR! LGTM!

tyt2y3:
Thank you all

tyt2y3:
[0.24.1](https://github.com/SeaQL/sea-query/releases/tag/0.24.1)

PR #304, opened 2022-04-17
Simple Expression support for Case Statements

npapapietro:
## PR Info

Adding support for the use of SQL case statements. This is a first attempt. Because CASE statements are more of an expression that can be used as columns, group bys, order bys, joins etc, I built a new option into `SimpleExpr` to keep them very portable. Pending approval for moving along with this PR, I will add further unit tests, but I have put two doc string tests to illustrate the usage.

## Adds

A new struct called `CaseStatement` and a new option, `SimpleExpr::Case`, and a new method `Expr::case`. Since case statements essentially are a series of `IF/ELSE` statements, the basic logic is to build a vector of a `Option<Condition>` and a resulting `Expr`.  We would use `None` for the condition if it is the `ELSE` clause. It is not required as any conditions met that aren't explicit in the `WHEN ... THEN..` are null.

### DIFF
--- a/src/backend/query_builder.rs
+++ b/src/backend/query_builder.rs
@@ -375,7 +375,40 @@ pub trait QueryBuilder: QuotedBuilder {
             SimpleExpr::AsEnum(_, expr) => {
                 self.prepare_simple_expr(expr, sql, collector);
             }
+            SimpleExpr::Case(case_stmt) => {
+                self.prepare_case_statement(case_stmt, sql, collector);
+            }
         }
+    }
+
+    fn prepare_case_statement(
+        &self,
+        stmts: &Vec<(Option<SimpleExpr>, SimpleExpr)>,
+        sql: &mut SqlWriter,
+        collector: &mut dyn FnMut(Value),
+    ) {
+        assert!(
+            stmts.len() >= 2,
+            "Case statements need at least two results"
+        );
+        write!(sql, "( CASE ").unwrap();
+        for (cond, res) in stmts.into_iter() {
+            
+            if let Some(when) = cond.clone() {
+                write!(sql, "WHEN ( ").unwrap();
+                self.prepare_condition_where(&when.into_condition(), sql, collector);
+                write!(sql, " ) ").unwrap();
+                write!(sql, "THEN ").unwrap();
+            }
+            else{
+                write!(sql, "ELSE ").unwrap();
+            }
+
+            self.prepare_simple_expr(res, sql, collector);
+            
+        }
+
+        write!(sql, "END ) ").unwrap();
     }
 
     /// Translate [`SelectDistinct`] into SQL statement.
--- a/src/backend/query_builder.rs
+++ b/src/backend/query_builder.rs
@@ -395,11 +395,7 @@ pub trait QueryBuilder: QuotedBuilder {
         for case in when.iter() {
             write!(sql, " WHEN (").unwrap();
             self.prepare_condition_where(&case.condition, sql, collector);
-            write!(sql, ") ").unwrap();
-            write!(sql, "THEN ").unwrap();
-            // } else {
-            //     write!(sql, " ELSE ").unwrap();
-            // }
+            write!(sql, ") THEN ").unwrap();
 
             self.prepare_simple_expr(&case.result.clone().into(), sql, collector);
         }
--- a/src/backend/query_builder.rs
+++ b/src/backend/query_builder.rs
@@ -381,34 +381,32 @@ pub trait QueryBuilder: QuotedBuilder {
         }
     }
 
+    /// Translate [`CaseStatement`] into SQL statement.
     fn prepare_case_statement(
         &self,
-        stmts: &Vec<(Option<SimpleExpr>, SimpleExpr)>,
+        stmts: &CaseStatement,
         sql: &mut SqlWriter,
         collector: &mut dyn FnMut(Value),
     ) {
         assert!(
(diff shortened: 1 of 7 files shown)


tyt2y3 (commented):
Thank you. I like the API in general. 
However is there a way to construct a simple case without `Condition::all().add()`?
We can probably do it by introducing a new trait `IntoCaseStatement` that accepts either a `SimpleExpr` or `Condition` (very much like how `IntoTableRef` works)

src/query/case.rs
```diff
@@ -0,0 +1,113 @@
+use crate::{Condition, Expr, SimpleExpr};
+
+#[derive(Debug, Clone)]
+pub(crate) struct CaseStatementCondition {
```
tyt2y3, line 4: In such case, we'd better use a enum:
```rust
enum CaseStatementCondition {
    When({ condition: Condition, result: Expr }),
    Else(Expr),
}
```

tyt2y3 (changes requested):
Thank you for the contribution and welcome to the community!

ikrivosheev:
> Thank you. I like the API in general. 
> However is there a way to construct a simple case without `Condition::all().add()`?
> We can probably do it by introducing a new trait `IntoCaseStatement` that accepts either a `SimpleExpr` or `Condition` (very much like how `IntoTableRef` or `ConditionExpression` works)

I like this idea)

src/query/case.rs
```diff
@@ -0,0 +1,113 @@
...
+    /// `ELSE` clause. Note that `ELSE` is optional and will return NULL
+    /// if `ELSE` is not present but the condition is met.
+    pub(crate) condition: Option<Condition>,
+    /// The result if the condition is met. This is the `THEN` clause.
+    pub(crate) result: Expr,
+}
+
+#[derive(Debug, Clone)]
+pub struct CaseStatement {
+    pub(crate) conditions: Vec<CaseStatementCondition>,
```
ikrivosheev, line 16: I think better struct is:
```
struct CaseStatement {
    conditions: Vec<...>,
    else: Option<...>,
}
```
Because grammar is:
```
CASE WHEN condition THEN result
     [WHEN ...]
     [ELSE result]
END
```
and `WHEN` `ELSE` are different parts

npapapietro:
> Thank you. I like the API in general. However is there a way to construct a simple case without `Condition::all().add()`? We can probably do it by introducing a new trait `IntoCaseStatement` that accepts either a `SimpleExpr` or `Condition` (very much like how `IntoTableRef` or `ConditionExpression` works)

Yes! I'll have something mocked up in the next couple days.

src/backend/query_builder.rs
```diff
@@ -375,7 +375,40 @@ pub trait QueryBuilder: QuotedBuilder {
...
+        write!(sql, "(CASE").unwrap();
+
+        let CaseStatement { when, r#else } = stmts;
+
+        for case in when.iter() {
+            write!(sql, " WHEN (").unwrap();
+            self.prepare_condition_where(&case.condition, sql, collector);
+            write!(sql, ") ").unwrap();
+            write!(sql, "THEN ").unwrap();
+            // } else {
```
ikrivosheev, line 400: maybe it's redundant?

ikrivosheev:
@tyt2y3 LGTM, @npapapietro thank you for contribution.

ikrivosheev (approved):

billy1624 (commented):
Hey @npapapietro, thanks for the contributions!
Looks good! It'd be better if we have test cases for MySQL and SQLite :)

npapapietro:
Added a test case for sqlite, mysql, and postgres.

billy1624 (approved):
Thanks!! @npapapietro

src/query/case.rs
```diff
@@ -0,0 +1,125 @@
...
+    ///     .to_owned();
+    ///
+    /// assert_eq!(
+    ///     query.to_string(PostgresQueryBuilder),
+    ///     r#"SELECT (CASE WHEN ("glyph"."aspect" > 0) THEN 'positive' WHEN ("glyph"."aspect" < 0) THEN 'negative' ELSE 'zero' END) AS "polarity" FROM "glyph""#
+    /// );    
+    /// ```
+    pub fn case<C>(mut self, case: C) -> Self
+    where
+        C: IntoCaseStatement,
```
tyt2y3, line 72: Sorry for the confusion but seems `IntoCaseStatement` is not needed. `IntoCondition` already did the trick.

here it can simply be:

```rust
pub fn case<C>(mut self, cond: C, then: T) -> Self
    where
        C: IntoCondition,
        T: Into<Expr>
```

then we don't need the extra braces when using calling it.

src/query/case.rs
```diff
@@ -0,0 +1,125 @@
...
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use sea_query::{*, tests_cfg::*};
+    ///
+    /// let query = Query::select()
+    ///     .expr_as(
+    ///             Expr::case((
+    ///                 Expr::tbl(Glyph::Table, Glyph::Aspect).gt(0),
```
tyt2y3, line 52: Finally, I'd appreciate at least one example is a complex condition, may be `Cond::all().add(...)` while the other is a `SimpleExpr` to illustrate the API

tyt2y3 (approved):

tyt2y3:
Thank you for the revisions. Such a nice API now! I believe people are going to appreciate it too.

Issue #294, opened 2022-04-06
String escape doesn't work for SQLite


kylestang:
## Description

The [escape_string](https://github.com/SeaQL/sea-query/blob/master/src/value.rs#L1044) function replaces single quotes `'` with `\'` to insert strings into the database, which doesn't work for SQLite. Here's the relevant part of the [documentation](https://www.sqlite.org/lang_expr.html):

>  A string constant is formed by enclosing the string in single quotes ('). A single quote within the string can be encoded by putting two single quotes in a row - as in Pascal. C-style escapes using the backslash character are not supported because they are not standard SQL. 

Using `''` to escape single quotes works in MySQL, Postgres, and SQLite so it should be a pretty easy fix. I'm happy to submit a PR if people are fine with that solution.

## Steps to Reproduce

1. Using rusqlite, create a db and table with a text field.
2. Create an insert query that contains a single quote using SqliteQueryBuilder.
3. Execute the query with rusqlite and unwrap the result.

### Expected Behavior

The entry is added to the table.

### Actual Behavior

An SqliteFailure error is returned.

### Reproduces How Often

Every time.

## Versions

sea-query v0.24.0
sea-query-derive v0.2.0
sea-query-driver v0.1.0
rusqlite v0.27.0

## Additional Information

Reproducible example:

```rust
use rusqlite::Connection;
use sea_query::{ColumnDef, Iden, Query, SqliteQueryBuilder, Table};

#[derive(Iden)]
enum Foo {
    Table,
    Text,
}

fn main() {
    let db = Connection::open("my_db.sqlite").unwrap();
    
    let table = Table::create()
        .table(Foo::Table)
        .if_not_exists()
        .col(ColumnDef::new(Foo::Text).text())
        .build(SqliteQueryBuilder);

    let insert = Query::insert()
        .into_table(Foo::Table)
        .columns(vec![Foo::Text])
        .values_panic(vec!["Can't insert".into()])
        .to_string(SqliteQueryBuilder);

    db.execute(&table, []).unwrap();
    db.execute(&insert, []).unwrap(); // returns error
}
```

ikrivosheev:
@kylestang thank you for the report! To me this looks like a good solution.

@tyt2y3 @billy1624?

tyt2y3:
I am not sure how bad it might be, but I think we need dedicated escape and unescape functions for each backend with logic properly implemented from respective docs.

p.s. the current implementation follows `mysql_real_escape_string`.

[...]

PR #306, opened 2022-04-21
issues-294 Move escape and unescape string to backend

ikrivosheev:
## PR Info

- Closes <https://github.com/SeaQL/sea-query/issues/294>

## Adds

Add new trait: `EscapeBuilder`

## Breaking Changes

Remove `escape_string` and `unescape_string` functions.

### DIFF
--- a/src/value.rs
+++ b/src/value.rs
@@ -1,5 +1,4 @@
 //! Container for all SQL value types.
-use std::fmt::Write;
 
 #[cfg(feature = "with-json")]
 use serde_json::Value as Json;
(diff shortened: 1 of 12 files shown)


src/value.rs
```diff
@@ -1177,34 +1132,6 @@ impl Values {
...
-        assert_eq!(unescape_string(escape_string(test).as_str()), test);
-    }
-
-    #[test]
-    fn test_escape_4() {
-        let test = "a\"b";
-        assert_eq!(escape_string(test), "a\\\"b".to_owned());
-        assert_eq!(unescape_string(escape_string(test).as_str()), test);
-    }
-
```
billy1624, line 1207: Hey @ikrivosheev, could we add back these test cases in query test?

ikrivosheev: @billy1624 I cannot find query test.

tyt2y3: Agree with Billy we'd better not delete the test cases (actually we might want more)

ikrivosheev: @tyt2y3 @billy1624, done, I divided the tests for each backend.

billy1624: Thanks! @ikrivosheev

billy1624 (approved):

PR #307, opened 2022-04-22
Add length check on condition array

SebastienGllmt:
## PR Info

I had a function in my codebase that takes a list and constructs a `Condition` object from it. I had a bug in my code though where after the filtering, the Condition objects end up not having any condition added to it which causes issues.

To solve this, I want to be able to check if a `Condition` object actually has any condition associated with it before I execute a query

## Adds

- Ability to check if a `Condition` is empty (and how many conditions it has)

### DIFF
--- a/src/query/condition.rs
+++ b/src/query/condition.rs
@@ -233,6 +233,28 @@ impl Condition {
         self.negate = !self.negate;
         self
     }
+
+    /// Whether or not any condition has been added
+    ///
+    /// # Examples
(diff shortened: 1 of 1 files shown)


billy1624 (approved):
Thanks!! @SebastienGllmt

ikrivosheev (approved):
LGTM!

tyt2y3:
Thank you all

PR #308, opened 2022-04-24
fix index create statement

hunjixin:
## PR Info

fix too old create index statement

https://github.com/SeaQL/sea-orm/issues/682

## Fixes

move index columns before index type

index statement: [https://dev.mysql.com/doc/refman/5.6/en/create-index.html](https://dev.mysql.com/doc/refman/5.6/en/create-index.html)

### DIFF
--- a/src/backend/mysql/index.rs
+++ b/src/backend/mysql/index.rs
@@ -24,9 +24,9 @@ impl IndexBuilder for MysqlQueryBuilder {
             table.prepare(sql, self.quote());
         }
 
-        self.prepare_index_type(&create.index_type, sql);
-
         self.prepare_index_columns(&create.index.columns, sql);
+
+        self.prepare_index_type(&create.index_type, sql);
     }
 
     fn prepare_index_drop_statement(&self, drop: &IndexDropStatement, sql: &mut SqlWriter) {


tyt2y3:
Thank you for the report.
Can you confirm that it's a long standing bug or it actually only broke in 8.0?

hunjixin:
> Thank you for the report. Can you confirm that it's a long standing bug or it actually only broke in 8.0?
> 
> My concern is whether this change is compatible with MySQL 5.5

also test fail on 5.7,  and check on mysql document in 5.6   column should be ahead of USING

```
CREATE [ONLINE | OFFLINE] [UNIQUE | FULLTEXT | SPATIAL] INDEX index_name
    [index_type]
    ON tbl_name (key_part,...)
    [index_option]
    [algorithm_option | lock_option] ...

key_part:
    col_name [(length)] [ASC | DESC]

index_option: {
    KEY_BLOCK_SIZE [=] value
  | index_type
  | WITH PARSER parser_name
  | COMMENT 'string'
}

index_type:
    USING {BTREE | HASH}

algorithm_option:
    ALGORITHM [=] {DEFAULT | INPLACE | COPY}

lock_option:
    LOCK [=] {DEFAULT | NONE | SHARED | EXCLUSIVE}
```

hunjixin:
i find 5.5 syntax here. http://download.nust.na/pub6/mysql/doc/refman/5.5/en/create-index.html. this maybe broken in mysql 5.5. maybe move index type before ON should be work on 5.5 right? @tyt2y3 
```
CREATE [UNIQUE|FULLTEXT|SPATIAL] INDEX index_name
    [index_type]
    ON tbl_name (index_col_name,...)
    [index_option] ...

index_col_name:
    col_name [(length)] [ASC | DESC]

index_type:
    USING {BTREE | HASH | RTREE}

index_option:
    KEY_BLOCK_SIZE [=] value
  | index_type
  | WITH PARSER parser_name
```

ikrivosheev (approved):
@hunjixin Thank you for contribution! LGTM.

billy1624 (changes requested):
Hey @hunjixin, thanks for the contributions!

What I found however, was that `USING ...` should be in between index name and table name.

```sql
`CREATE INDEX `idx-node-name` USING HASH ON `node` (`name`)`
```

Demo: https://dbfiddle.uk/?rdbms=mysql_8.0&fiddle=d38cd2bc14e7ff859e83136201fc5236

Docs:
- https://dev.mysql.com/doc/refman/5.6/en/create-index.html
- https://dev.mysql.com/doc/refman/5.7/en/create-index.html
- https://dev.mysql.com/doc/refman/8.0/en/create-index.html
- https://mariadb.com/kb/en/create-index/

tyt2y3:
oh this is interesting. So `USING` can actually comes before or after `ON SOMTHING`

PR #311, opened 2022-04-25
issues-294 Move escape and unescape string to backend (#306)

tyt2y3:
continues #306

### DIFF
--- a/src/backend/sqlite/mod.rs
+++ b/src/backend/sqlite/mod.rs
@@ -25,4 +25,46 @@ impl QuotedBuilder for SqliteQueryBuilder {
     }
 }
 
-impl EscapeBuilder for SqliteQueryBuilder {}
+impl EscapeBuilder for SqliteQueryBuilder {
+    fn escape_string(&self, string: &str) -> String {
+        string
+            .replace('\\', "\\\\")
+            .replace('"', "\\\"")
+            .replace('\'', "''")
+            .replace('\0', "\\0")
+            .replace('\x08', "\\b")
+            .replace('\x09', "\\t")
+            .replace('\x1a', "\\z")
+            .replace('\n', "\\n")
+            .replace('\r', "\\r")
+    }
+
+    fn unescape_string(&self, string: &str) -> String {
+        let mut escape = false;
+        let mut output = String::new();
+        for c in string.chars() {
+            if !escape && (c == '\\' || c == '\'') {
+                escape = true;
+            } else if escape {
+                write!(
+                    output,
+                    "{}",
+                    match c {
+                        '0' => '\0',
+                        'b' => '\x08',
+                        't' => '\x09',
+                        'z' => '\x1a',
+                        'n' => '\n',
+                        'r' => '\r',
+                        c => c,
+                    }
+                )
+                .unwrap();
+                escape = false;
+            } else {
+                write!(output, "{}", c).unwrap();
+            }
+        }
+        output
+    }
+}
--- a/src/backend/sqlite/mod.rs
+++ b/src/backend/sqlite/mod.rs
@@ -24,3 +24,5 @@ impl QuotedBuilder for SqliteQueryBuilder {
         '"'
     }
 }
+
+impl EscapeBuilder for SqliteQueryBuilder {}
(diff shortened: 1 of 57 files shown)


tyt2y3:
Great that we split that now. That say, the current implementation does not fully work for SQLite, @kylestang would you mind contributing a bit

ikrivosheev:
@tyt2y3 done. Сan you look at this PR?

ikrivosheev:
Link with issues: https://github.com/SeaQL/sea-query/issues/294

billy1624 (approved):
Thanks! @ikrivosheev LGTM

src/backend/sqlite/mod.rs
```diff
@@ -24,3 +24,5 @@ impl QuotedBuilder for SqliteQueryBuilder {
         '"'
     }
 }
+
+impl EscapeBuilder for SqliteQueryBuilder {}
```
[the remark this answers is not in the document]
ikrivosheev: Done

tyt2y3:
I don' think it's the the correct implementation for SQLite. It confuses everybody but the backslash `\` absolutely does not work

ikrivosheev:
@tyt2y3 why do you think is not correct implementation for SQLite?

src/backend/sqlite/mod.rs
```diff
@@ -24,3 +24,47 @@ impl QuotedBuilder for SqliteQueryBuilder {
...
+        string
+            .replace('\\', "\\\\")
+            .replace('"', "\\\"")
+            .replace('\'', "''")
+            .replace('\0', "\\0")
+            .replace('\x08', "\\b")
+            .replace('\x09', "\\t")
+            .replace('\x1a', "\\z")
+            .replace('\n', "\\n")
+            .replace('\r', "\\r")
```
tyt2y3, line 39: It seems like that we should not escape those with a `\` but then I don't know what should be the correct way of handling those. May be just ignoring them?

ikrivosheev: I'm a little confused... I thought was one problem with: `\``, I cannot find alternative implementation for `escape` string for sqlite and cannot find about this in documentation. 
@tyt2y3 can you explain to me please)

ikrivosheev:
> Thank you so much for the effort. I am confused too. I think may be we wait until a SQLite expert to tell us the correct way.

I found source code sqlite3: https://code.woboq.org/qt5/qtbase/src/3rdparty/sqlite/sqlite3.c.html#sqlite3_str_vappendf

ikrivosheev:
> Thank you so much for the effort. I am confused too. I think may be we wait until a SQLite expert to tell us the correct way.

I find: [SQLite3::escapeString](https://www.php.net/manual/en/sqlite3.escapestring.php), then I write simple code:

```php
<?php
function printAscii($string){      
    for ($n=0;$n<strlen($string);$n++){ 

       echo "<".ord($string[$n]).">";
    }          
} 

echo printAscii(SQLite3::escapeString("'\r\n\""));  // <39><39><13><10><34>
```
We can see, php does not escape `\r\n"`. I think we can do the same.

ikrivosheev:
@tyt2y3 @billy1624 done. I remove escape all character for SqlLite except `'`

tyt2y3:
Thank you so much for the investigation. I think following PHP would be right thing to do.

ikrivosheev:
> Thank you so much for the investigation. I think following PHP would be the right thing to do.

@tyt2y3 I have done with PR. Can you review and merge?

ikrivosheev:
@tyt2y3 some more information. From [php source code](https://github.com/php/php-src/blob/master/ext/sqlite3/sqlite3.c#L484):
```c
/* {{{ Returns a string that has been properly escaped. */
PHP_METHOD(SQLite3, escapeString)
{
	zend_string *sql;
	char *ret;

	if (FAILURE == zend_parse_parameters(ZEND_NUM_ARGS(), "S", &sql)) {
		RETURN_THROWS();
	}

	if (ZSTR_LEN(sql)) {
		ret = sqlite3_mprintf("%q", ZSTR_VAL(sql));  // <-- READ THIS STRING
		if (ret) {
			RETVAL_STRING(ret);
			sqlite3_free(ret);
		}
	} else {
		RETURN_EMPTY_STRING();
	}
}
```

Then got to sqlite documentation: https://www.sqlite.org/printf.html#percentq and read about: `%q`:

> The argument is a zero-terminated string. The string is printed with all single quote (') characters doubled so that the string can safely appear inside an SQL string literal. The %Q substitution type also puts single-quotes on both ends of the substituted string.

tyt2y3:
Thank you for the detailed investigation!

PR #313, opened 2022-04-26
Feature/issues 250 distinct on

ikrivosheev:
## PR Info

- Closes <https://github.com/SeaQL/sea-query/issues/250>
- Continues: https://github.com/SeaQL/sea-query/pull/257

## Adds

Add support for `SELECT DISTINCT ON ...`

### DIFF
--- a/src/backend/sqlite/query.rs
+++ b/src/backend/sqlite/query.rs
@@ -50,6 +50,11 @@ impl QueryBuilder for SqliteQueryBuilder {
         // Sqlite doesn't support sql recursive with query 'SEARCH' and 'CYCLE' options.
     }
 
+    fn insert_default_values(&self, _: u32, sql: &mut SqlWriter) {
+        // SQLite doesn't support inserting multiple rows with default values
+        write!(sql, "DEFAULT VALUES").unwrap()
+    }
+
     fn prepare_select_distinct(
         &self,
         select_distinct: &SelectDistinct,
--- a/src/backend/sqlite/query.rs
+++ b/src/backend/sqlite/query.rs
@@ -49,4 +49,17 @@ impl QueryBuilder for SqliteQueryBuilder {
     ) {
         // Sqlite doesn't support sql recursive with query 'SEARCH' and 'CYCLE' options.
     }
+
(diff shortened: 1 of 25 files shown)


ikrivosheev:
@oonxt @tyt2y3 @billy1624 I'a done, can you look this PR?

src/query/select.rs
```diff
@@ -56,11 +56,12 @@ pub struct SelectStatement {
 }
 
 /// List of distinct keywords that can be used in select statement
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[derive(Debug, Clone)]
 pub enum SelectDistinct {
     All,
```
billy1624, line 61: I think we don't need `SelectDistinct::All`? I actually can't see it's being used and we don't need it anyways because `SELECT ALL` is the default behaviour.

src/backend/sqlite/query.rs
```diff
@@ -49,4 +49,17 @@ impl QueryBuilder for SqliteQueryBuilder {
...
+        select_distinct: &SelectDistinct,
+        sql: &mut SqlWriter,
+        _collector: &mut dyn FnMut(Value),
+    ) {
+        match select_distinct {
+            SelectDistinct::All => write!(sql, "{}", "ALL").unwrap(),
+            SelectDistinct::Distinct => write!(sql, "{}", "DISTINCT").unwrap(),
+            _ => {}
+        }
+    }
```
billy1624, line 64: I guess we can safely remove this and fallback to the default implementation?

ikrivosheev: Default implementation is blank. What do you mean?

tyt2y3:
Thank you

maxclaey:
Great, thanks! I was trying to give it a go already by specifying a git repository dependency, but that doesn't seem to work (my guess is because of the mixed workspace/package Cargo.toml)

wurikiji:
How can I use with `sea-orm`'s `Entity` and `Column` ?

tyt2y3:
We can try bumping sea-query to 0.25 now?

PR #314, opened 2022-04-27
support bigger blob

hunjixin:
## PR Info

support medium blob longblob type

## Changes

this implement is a little track , but in the current implementation there is specific blob type

### DIFF
--- a/src/backend/mysql/table.rs
+++ b/src/backend/mysql/table.rs
@@ -77,9 +77,17 @@ impl TableBuilder for MysqlQueryBuilder {
                 },
                 ColumnType::Date => "date".into(),
                 ColumnType::Interval(_, _) => "unsupported".into(),
-                ColumnType::Binary(length) => match length {
-                    Some(length) => format!("binary({})", length),
-                    None => "blob".into(),
+                ColumnType::Binary(blob_size) => match blob_size {
+                    BlobSize::Default(length) => {
+                        match length {
+                            Some(length) => format!("binary({})", length),
+                            None => "blob".into(),  // todo already need? or keep for compatible
+                        }
+                    },
+                    BlobSize::Tiny => "tinyblob".into(),
+                    BlobSize::Blob => "blob".into(),
+                    BlobSize::Medium => "mediumblob".into(),
+                    BlobSize::Long => "longblob".into(),
                 },
                 ColumnType::Boolean => "bool".into(),
                 ColumnType::Money(precision) => match precision {
--- a/src/backend/mysql/table.rs
+++ b/src/backend/mysql/table.rs
@@ -78,14 +78,13 @@ impl TableBuilder for MysqlQueryBuilder {
                 ColumnType::Date => "date".into(),
                 ColumnType::Interval(_, _) => "unsupported".into(),
                 ColumnType::Binary(blob_size) => match blob_size {
-                    BlobSize::Default(length) => {
+                    BlobSize::Tiny => "tinyblob".into(),
+                    BlobSize::Blob(length) => {
                         match length {
                             Some(length) => format!("binary({})", length),
-                            None => "blob".into(),  // todo already need? or keep for compatible
+                            None => "blob".into(),
                         }
-                    },
-                    BlobSize::Tiny => "tinyblob".into(),
-                    BlobSize::Blob => "blob".into(),
+                    }
                     BlobSize::Medium => "mediumblob".into(),
                     BlobSize::Long => "longblob".into(),
                 },
--- a/src/table/column.rs
+++ b/src/table/column.rs
@@ -33,7 +33,7 @@ pub enum ColumnType {
     Time(Option<u32>),
     Date,
     Interval(Option<PgInterval>, Option<u32>),
-    Binary(Option<u32>),
+    Binary(BlobSize),
     Boolean,
     Money(Option<(u32, u32)>),
     Json,
@@ -72,6 +72,15 @@ pub enum PgInterval {
     HourToMinute,
     HourToSecond,
     MinuteToSecond,
+}
+
+#[derive(Debug, Clone)]
+pub enum BlobSize {
+    Default(Option<u32>), // Backward compatibility: MySQL & SQLite support `binary(length)` column type
+    Tiny,
+    Blob,  //default blob, maybe need a better name
+    Medium,
+    Long,
 }
 
 #[cfg(feature = "postgres-interval")]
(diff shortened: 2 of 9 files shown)


src/backend/mysql/table.rs
```diff
@@ -78,7 +78,19 @@ impl TableBuilder for MysqlQueryBuilder {
...
+                        } else if *length <= 65535 {
+                            //64KiB
+                            "blob".into()
+                        } else if *length <= 16777215 {
+                            //16MiB
+                            "mediumblob".into()
+                        } else {
+                            "longblob".into() //longblob max to 4294967295 4GiB but this limit can be config
+                        }
+                    }
```
billy1624, line 93: Hey @hunjixin, thank you! Just a nitpick, wouldn't it be more readable to express the number as power of 2?

```suggestion
                    Some(length) => {
                        if *length < (2 << 8) {
                            format!("binary({})", length)
                        } else if *length < (2 << 16) {
                            //64KiB
                            "blob".into()
                        } else if *length < (2 << 24) {
                            //16MiB
                            "mediumblob".into()
                        } else {
                            //longblob max to 4294967295 4GiB but this limit can be config
                            "longblob".into()
                        }
                    }
```

References:
- https://dev.mysql.com/doc/refman/8.0/en/storage-requirements.html#data-types-storage-reqs-strings
- https://mariadb.com/kb/en/blob/

tyt2y3: how does that suggest change actually works?

billy1624: Select the lines you want to comment and suggest changes. Then, type the suggested changes in code block of `suggestion` language. See [GitHub Docs](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/commenting-on-a-pull-request#adding-line-comments-to-a-pull-request)

billy1624:
> This does not seem right. We should instead add more variants to ColumnType

But then all these new types, `mediumblob` and `longblob`, are not supported in other db backend. Hence, it will be maps to `blob` datatype in other db. Hope it won't confuse the users.

hunjixin:
> > This does not seem right. We should instead add more variants to ColumnType
> 
> But then all these new types, `mediumblob` and `longblob`, are not supported in other db backend. Hence, it will be maps to `blob` datatype in other db. Hope it won't confuse the users.

best way to support blob is add new column type,

billy1624:
> consider anyway to enable the database to support unique types

Are you suggesting `ColumnType` that are feature guarded by db backend? I don't think this is a good idea. We try to avoid relying on feature guard by db backend.

> But then all these new types, `mediumblob` and `longblob`, are not supported in other db backend. Hence, it will be maps to `blob` datatype in other db.

This will be a good middle ground for me though we have to be explicit these new column types are primarily added for MySQL.

tyt2y3 (commented):
I think a blob type with a enum BlobSize param makes the most sense

billy1624:
> I think a blob type with a enum BlobSize param makes the most sense

You mean blob type with an `Option<BlobSize>` param?

tyt2y3:
Okay might not actually be related.
My rule of thumb is "anything that exists conceptually deserves to be named".

billy1624:
Hey @hunjixin, got any updates? :)

hunjixin:
> Hey @hunjixin, got any updates? :)

add new columntype "BLOB" to support this feature? if all think it okay, i'd like to implement one, but i am not sure whether affect other db types, maybe need help.

billy1624:
My plans are...

1. Add `BlobSize` to `Binary` variant in `ColumnType` enum
    ```rs
    Binary(BlobSize)
    ```
    With `BlobSize` defined as
    ```rs
    enum BlobSize {
        Default(Option<u32>), // Backward compatibility: MySQL & SQLite support `binary(length)` column type
        Medium,
        Long,
    }
    ```
2. Add helper methods to `ColumnDef`, namely `binary_medium` and `binary_long`
3. Update SQL builder in all three backends, if `binary_medium` and `binary_long` are not support in a particular backend then both of them will be mapped to `binary`

CC @tyt2y3 @hunjixin

hunjixin:
> My plans are...
> 
> 1. Add `BlobSize` to `Binary` variant in `ColumnType` enum
>    ```rust
>    Binary(BlobSize)
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    With `BlobSize` defined as
>    ```rust
>    enum BlobSize {
>        Default(Option<u32>), // Backward compatibility: MySQL & SQLite support `binary(length)` column type
>        Medium,
>        Long,
>    }
>    ```
> 2. Add helper methods to `ColumnDef`, namely `binary_medium` and `binary_long`
> 3. Update SQL builder in all three backends, if `binary_medium` and `binary_long` are not support in a particular backend then both of them will be mapped to `binary`
> 
> CC @tyt2y3 @hunjixin

see your idea, but have two suggestions

1. its better to include blob/tinyblob  include blobsize?
  
```rust
#[derive(Debug, Clone)]
enum BlobSize {
    Default(Option<u32>), // Backward compatibility: MySQL & SQLite support `binary(length)` column type
    Tiny,
    Blob,  //default blob, need a better name?
    Medium,
    Long,
}
```

2. maybe more appropriate to add the blob method, so as not to be confused with binary
```rust
    /// Set column type as blob
    pub fn blob(&mut self, size: BlobSize) -> &mut Self {
        //todo need to check set default size or convert to correct blob type?
        self.types = Some(ColumnType::Binary(size));
        self
    }
```

hunjixin:
@billy1624 updated pr

src/table/column.rs
```diff
@@ -74,6 +74,15 @@ pub enum PgInterval {
...
 }
 
+#[derive(Debug, Clone)]
+pub enum BlobSize {
+    Default(Option<u32>), // Backward compatibility: MySQL & SQLite support `binary(length)` column type
+    Tiny,
+    Blob,  //default blob, maybe need a better name
+    Medium,
+    Long,
+}
```
billy1624, line 84: Hey @hunjixin, all in all it looks good! I have one question. If we have `BlobSize::Blob`, should we make the "default" to `BlobSize::Default(u32)`? Removing `Option<T>`. Basically, `BlobSize::Default(None)` will be `BlobSize::Blob`.

tyt2y3: Makes sense. Actually seems like `BlobSize::Default` can be removed, we only need one `BlobSize::Blob(Option<u32>)` (like before)

hunjixin: got

hunjixin: updated, very cool idea.

billy1624: Thanks!

billy1624 (approved):
LGTM! Thanks! @hunjixin

billy1624:
I will updates the downstream accordingly
- https://github.com/SeaQL/sea-schema/pull/61
- https://github.com/SeaQL/sea-orm/pull/685

tyt2y3:
Thank you for the patience @hunjixin

PR #316, opened 2022-05-04
Introduce `JsonValue` wrapper struct

### LINKED ISSUE #185: Add `Json<T>` type to simplify working with Josn types
Hi, I would request to add `Json<T>` that we can use inside struct, this type would serialize and save `T` as json/jsonb data type in the database, and would de-serialize from database type to `Json<T>` so we can work with `T` as regular rust type (not as `serde_json::Value`).

the same idea can be found in sqlx too, Here is `Json<T>` from sqlx:
```rust
#[derive(
    Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Default, Serialize, Deserialize,
)]
#[serde(transparent)]
pub struct Json<T: ?Sized>(pub T);
```

so in sea-query, we would write this:
```rust
#[derive(sqlx::FromRow, Debug)]
struct CharacterStruct {
    id: i32,
    uuid: Uuid,
    character: String,
    font_size: i32,
    meta: Json<Meta>, // <-- use Json<Meta> instead of serde_json::Value
    created: NaiveDateTime,
}
```

and in sea-orm, I would write:
```rust
#[derive(Serialize, Deserialize, Debug, PartialEq, PartialOrd, Eq, Ord, Clone, DeriveEntityModel)]
#[sea_orm(table_name = "roles")]
pub struct Model {
    #[sea_orm(primary_key)]
    id: RoleId,
    name: RoleName,
    creation_info: Json<CreationInfo>, // <-- use Json<CreationInfo> instead of serde_json::Value
}
```

Using `Json<T>` would make working with json types way easier, so when we fetch `roles::Model` we would access `creation_info` as normal Rust struct instead of `serde_json::Value`:

```rust
let role = role::Entity::find().one().await?;
// accessing `creation_info` is easy
println!("{}", role.creation_info.0.created_by);
```

billy1624:
## PR Info

- Closes #185

- Dependents:
  - SeaQL/sea-orm#696

### TODO

- [x] A wrapper struct for `serde_json::Value`
- [x] Conversion traits and methods for the wrapper struct

## Adds

- A wrapper struct for `serde_json::Value`


billy1624:
Close in favor of https://github.com/SeaQL/sea-orm/pull/696#issuecomment-1154676140

PR #318, opened 2022-05-05
Feature/support escape

ikrivosheev:
Continue: https://github.com/SeaQL/sea-query/pull/94


ikrivosheev:
The situation is worse than I thought...

```rust
struct LikeClause {
    pattern: String,
    escape: Option<Char>,
}

struct LikeStatement {
    negative: bool,  // LIKE OR NOT LIKE
    expr: SimpleExpr,
   clause: LikeClause,
}

Into<SimpleExpr> for LikeStatement {...}

impl Expr {
    ...
    fn like(self, v: &str) -> LikeStatement  {...}
```

And we can write queries:

```rust
.and_where(Expr::col(Char::Character).like().escape('\ \'))
```

But I cannot write:
```rust
.and_where(Expr::col(Char::Character).like("").or(..))
```
Solution: create new trait:
```rust

trait Operator {
    fn unary_operator() -> SimpleExpr;
    fn binary_operator() -> SimpleExpr;
}

trait BoolOperator: Operator {
    fn and() -> SimpleExpr;
    fn or() -> SimpleExpr;
    fn not() -> SimpleExpr;

impl Operator for SimpleExpr;
impl BoolOperator for SimpleExpr;


impl Operator for LikeStatement;
impl BoolOperator for LikeStatement;
```

@billy1624 @Sytten @tyt2y3 What do you think? I will split PR:

1. Create new traits
2. Add support it for `LIKE`

ikrivosheev:
@tyt2y3, @billy1624 I push commits with breaking changes:

Cannot write query like:

```
and_where(Expr::col(Char::Character).like("").or(..))
```

What do you think about this?

src/shim.rs
```diff
@@ -57,23 +57,35 @@ macro_rules! impl_conditional_statement {
             use crate::{ConditionalStatement, SimpleExpr, IntoCondition, $struct_name};
 
             impl $struct_name {
-                pub fn and_where(&mut self, other: SimpleExpr) -> &mut Self {
-                    <Self as ConditionalStatement>::and_where(self, other)
```
billy1624, line 61: I think it's better to keep these as is. Then, calling `into` method when constructing filter with `LikeStatement`. E.g.

```rs
Query::select()
    .column(Char::Character)
    .from(Char::Table)
    .and_where(Expr::col(Char::Character).like("A").escape('\\').into())
```

billy1624:
I'm not sure what the `Operator` trait will do.

```rs
trait Operator {
    fn unary_operator() -> SimpleExpr;
    fn binary_operator() -> SimpleExpr;
}
```

For `BoolOperator`, I agree we can introduce this new trait for `SimpleExpr` and `LikeStatement`. I think we can provide default implementation for `and`, `or` and `not` methods, as long as it implemented `Into<SimpleExpr>`.

```rs
trait BoolOperator: Into<SimpleExpr> {
    fn and(self, right: SimpleExpr) -> SimpleExpr { ... }
    fn or(self, right: SimpleExpr) -> SimpleExpr { ... }
    fn not(self) -> SimpleExpr { ... }
}
```

ikrivosheev:
@billy1624 I can make this in another branch and after we can talk about it.

src/shim.rs
ikrivosheev: Api doesn't look good with: `.into()`.

Sytten:
Any progress @ikrivosheev ?

ikrivosheev:
@Sytten thank you for reminding ! I didn't understand:
1. Do we want to make breaking changes?
2. Or add a new trait for support operations: `or`, `and`, ...?

Sytten:
I have no opinion, usually traits are better but probably need to discuss with @billy1624

ikrivosheev:
I don't like api:  `Expr::col(Char::Character).like("").or(Expr::col(Char::Character).like("")))`. This is strange for me... 
In sea-query we have marcos: `all!` and `any!` which helps make conditions for where expressions.

tests/mysql/query.rs
```diff
@@ -997,6 +1003,18 @@ fn select_57() {
...
 
+#[test]
+fn select_58() {
+    assert_eq!(
+        Query::select()
+            .column(Char::Character)
+            .from(Char::Table)
+            .and_where(Expr::col(Char::Character).like("A").escape('\\'))
+            .to_string(MysqlQueryBuilder),
+        r#"SELECT `character` FROM `character` WHERE `character` LIKE 'A' ESCAPE '\'"#
```
tyt2y3, line 1014: Can we double confirm that this is the correct syntax? Because it seemed strange to me at first sight that the `\` in `'\'` does not actually escape the next `'`

tyt2y3:
I think we can simply change the argument into a generic:

```rust
fn like(self, v: IntoLikeExpr)
```

and then so it accepts both a &str (existing behaviour) and an object that provides:

```rust
Expr::col(Character).like(LikeExpr::str("E").escape('\'))
```

PR #319, opened 2022-05-05
Getter methods under "getters" feature for Statements.

sjud:
## Adds

The "getters" feature which adds getter methods on statements via a getters derive attribute configured with the feature flag.

## Changes

This getter methods derived from the getset Getter derive look like
```rust
impl STRUCT_NAME {
    pub fn get_FIELD_NAME(&self) -> &FIELD_TYPE {&self.FIELD_NAME}
}
```
When a structure has pre-existing and thus conflicting getter methods the impl was written manually under a #[cfg(feature="getters")] attribute.
When a structure had fields whose type was a pub(crate) struct, the struct was made pub to allow for getting from within it's parents related getter method. As far as I know there isn't a way to change visibility based on features currently.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,10 +50,12 @@ uuid = { version = "^0", optional = true }
 proc-macro2 = { version = "1", optional = true }
 quote = { version = "^1", optional = true }
 time = { version = "^0.2", optional = true }
+getset = { version = "^0.1.2", optional = true}
 
 [dev-dependencies]
 criterion = { version = "0.3", features = ["html_reports"] }
 pretty_assertions = { version = "^1" }
+
 
 [features]
 backend-mysql = []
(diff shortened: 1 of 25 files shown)


ikrivosheev:
@sjud thank you for contributing! Can you describe the situation where you need this feature?

sjud:
@ikrivosheev 
So I originally thought that this would be part of a solution for writing queries in the client and authenticating them on the server before executing them but I'm looking higher up the abstraction layer currently. But if this ends up being the solution I will supply an example. Right now I'm working with my forked version to see if I actually need this feature or another layer of abstraction outside of the sea-ql libraries. I opened a pull request before I could prove the need for functionality,  sorry, but I'll leave it open for now in case it does become part of the solution. Thank you, for your patience. :)

tyt2y3:
Cool, thanks for the info! Let us what you found.

PR #321, opened 2022-05-06
Partial oracle support

Niedzwiedzw:
## Adds
partial oracle db support (selects only as far as I have been able to test. I have created an automated tool to quickly translate queries postgres -> oracle sql from postgres' testcases, so full support should be possible. It is a bit time consuming though...


billy1624 (commented):
Hey @Niedzwiedzw, welcome and sorry for the delay! Thanks for taking the initiative to introduce Oracle support for SeaQuery!

Currently, we want to focus on bringing support for open-source databases and have no plan to support proprietary databases such as Oracle.

Niedzwiedzw:
I understand, I will try and maintain this fork for the time being then

PR #324, opened 2022-05-11
Make sea-query-driver an optional dependency

### LINKED ISSUE #323: "sea-query-driver" dependency isn't optional, loads proc macro crates
## Description

One of the things that immediately stood out to me about sea-query was: `SeaQuery is very lightweight, all dependencies are optional.`

Awesome! No waiting for long build times on procedural macros 😁

Unfortunately, this appears to not be the case -- `sea-query-driver` is a hard dependency, and it depends on `syn`, `proc_macro2`, and some of the other usual suspects of procedural macros.

It looks like it's not actually needed as part of the library, and it was likely just a mistake that the words `optional = true` and some `cfg` checks were omitted.

I have an incoming PR to remedy the situation

## Steps to Reproduce

Build `sea-query` from another project with default features disabled

### Expected Behavior

`sea-query` should be the only thing built

### Actual Behavior

`sea-query-driver` and several of its dependencies are also pulled from `crates.io` and built

### Reproduces How Often

Always

## Versions

current

marti4d:
Fixes #323

## PR Info

- Closes #323

## Fixes

- [x] sea-query-driver being built even with all default features disabled

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,7 +39,7 @@ path = "src/lib.rs"
 [dependencies]
 sea-query-attr = { version = "^0.1.1", path = "sea-query-attr", optional = true }
 sea-query-derive = { version = "0.2.0", path = "sea-query-derive", optional = true }
-sea-query-driver = { version = "^0.1.1", path = "sea-query-driver" }
+sea-query-driver = { version = "^0.1.1", path = "sea-query-driver", optional = true }
 serde_json = { version = "^1", optional = true }
 bytes = { version = "^1", optional = true }
 chrono = { version = "^0", optional = true }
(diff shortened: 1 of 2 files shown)


tyt2y3:
Nice catch!

tyt2y3:
https://github.com/SeaQL/sea-query/releases/tag/0.24.6

Sytten:
It should have been added to the default features though since this will break people's code...

tyt2y3:
Thanks for pointing that out.
I think I released the correct code, but I accidentally hit the Release on GitHub before successfully pushed and so tagged the wrong version.

> It should have been added to the default features though since this will break people's code...

I think this will be fine. For example, `sqlx-mysql` will enable `sea-query-driver/sqlx-mysql` implicitly.

PR #328, opened 2022-05-11
Don't build driver for chrono/json/uuid/time/etc unless explicitly asked

### LINKED ISSUE #327: Don't unconditionally build driver when chrono/json/time/uuid/etc are enabled
## Description

Currently, if you enable the chrono/json/time/uuid features it adds a dependency on `sea-query-driver/<feature>`, which forces the `sea-query-driver` to be built as well.

If we add a question mark (?), we can change Cargo's behavior to only enable the feature if `sea-query-driver` is already enabled by something else.

marti4d:
Cargo.toml files can use the "?" operator in their sub-dependencies to avoid
building the main dependency unless it is explicitly enabled by another feature

This uses that operator to avoid building the driver crate for
chrono/json/uuid/time/etc unless the driver crate is already enabled
explicitly

Fixes #327

## PR Info

- Closes #327

## Fixes

- [x] The driver will no longer be built unless explicitly needed

## Breaking Changes

- [x] It's possible that it may break downstream code that relied on that behavior


billy1624 (commented):
Hey @marti4d, thanks for the contributions! Good use of the new cargo feature.
- https://blog.rust-lang.org/2022/04/07/Rust-1.60.0.html#new-syntax-for-cargo-features

Btw this means user have to upgrade their Rust to at least 1.60 to compile SeaQuery? But I guess it's not an issue.

tyt2y3:
This is GREAT

PR #331, opened 2022-05-18
Add `VarBinary` column type

billy1624:
## PR Info

- Closes SeaQL/sea-orm#704

- Dependents:
  - SeaQL/sea-schema#67

## Adds

- [x] Add `VarBinary` column type as two out of three db backend support this datatype. Only SQLite doesn't support it

## References
- https://mariadb.com/kb/en/varbinary/
- https://dev.mysql.com/doc/refman/8.0/en/binary-varbinary.html
- https://www.postgresql.org/docs/current/datatype-bit.html


ikrivosheev (approved):
@billy1624 thank you! LGTM!

PR #332, opened 2022-05-21
Support `IF NOT EXISTS` when create index

liberwang1013:
## PR Info
This PR add support `IF NOT EXISTS` when create index

## Adds

- [x] support `IF NOT EXISTS` when create index in postgres & sqlite

### DIFF
--- a/src/index/create.rs
+++ b/src/index/create.rs
@@ -103,6 +103,7 @@ pub struct IndexCreateStatement {
     pub(crate) primary: bool,
     pub(crate) unique: bool,
     pub(crate) index_type: Option<IndexType>,
+    pub(crate) if_not_exists: bool,
 }
 
 /// Specification of a table index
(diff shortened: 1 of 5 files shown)


src/index/create.rs
```diff
@@ -129,9 +130,16 @@ impl IndexCreateStatement {
             primary: false,
             unique: false,
             index_type: None,
+            if_not_exists: false,
         }
     }
 
+    /// Create index if index not exists
```
ikrivosheev, line 137: I think is better add doc tests instead `tests`. It will be like examples

ikrivosheev (commented):
@liberwang1013 thank you for PR! Some comments.

src/index/create.rs
liberwang1013: already add doc tests

ikrivosheev (approved):
@liberwang1013 thank you. LGTM!

tyt2y3 (approved):
Awesome

tyt2y3:
Thank you all

PR #341, opened 2022-05-30
upgrade time dependency version to 0.3

### LINKED ISSUE #340: upgrade time to 0.3.x version
## Description

The [time](https://crates.io/crates/time) crate version 0.3.x as been released for a while now and `sea-orm` is still on version 0.2, blocking the integration with tools like `async-graphql` which relies on 0.3.x.
I'm opening this issue to propose a PR 😉

jdrouet:
The [time](https://crates.io/crates/time) crate version 0.3.x as been released for a while now and sea-orm is still on version 0.2, blocking the integration with tools like async-graphql which relies on 0.3.x.

## PR Info

- Closes #340

## Adds

- [x] create a feature `with-time-0_2` to use `time@^0.2`
- [x] create a feature `with-time-0_3` to use `time@^0.3`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -741,6 +741,7 @@
 )]
 
 pub mod backend;
+pub mod deps;
 pub mod driver;
 pub mod error;
 pub mod expr;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -741,7 +741,7 @@
 )]
 
 pub mod backend;
-pub mod deps;
+mod deps;
 pub mod driver;
 pub mod error;
 pub mod expr;
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -22,13 +22,20 @@ jobs:
           toolchain: stable
           components: clippy
           override: true
-          
+
       - uses: actions-rs/cargo@v1
         with:
           command: clippy
           args: >
             --all-targets
-            --all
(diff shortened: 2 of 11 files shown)


ikrivosheev (commented):
@jdrouet hello! Thank you for contributation. I think we need support both `time = 0.2` and `time = 0.3` Because postgres-types has support both time version: https://github.com/SeaQL/sea-query/blob/79cacdde4ee4950e986de8306069d8bd49e4e3ed/Cargo.toml#L73

src/lib.rs
```diff
@@ -741,6 +741,7 @@
 )]
 
 pub mod backend;
+pub mod deps;
```
ikrivosheev, line 744: I think `deps` should be `pub(crate)`, because this is internal mod.

ikrivosheev (commented):
@jdrouet awesome! Small comment.

ikrivosheev:
@jdrouet some problems with tests, because test run with: `--all-feature``

src/lib.rs
jdrouet: well, we need to expose it to use it in `tests/*`

ikrivosheev: My point was: when we were making some internal mod public it became part of the public API  sea-query and every change is breaking...

jdrouet: sure, I'll try to fix that. For now I'm trying to find a nice way to fix the CI 😉

ikrivosheev:
@jdrouet hello! Thank you for PR. We have done in: https://github.com/SeaQL/sea-query/issues/356

PR #348, opened 2022-06-07
Add is / is_not to Expr

Sytten:
For Sqlite, they are the same as `eq` and `neq` but treat NULL differently.
I put them behind a flag since I don't think other DBMS allow random values with those operators.

## Adds

- `IS` and `IS NOT` operators for `Sqlite`

### DIFF
--- a/src/expr.rs
+++ b/src/expr.rs
@@ -1169,6 +1169,40 @@ impl Expr {
         self.bin_oper(BinOper::Is, SimpleExpr::Keyword(Keyword::Null))
     }
 
+    /// Express a `IS` expression.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use sea_query::{*, tests_cfg::*};
+    ///
+    /// let query = Query::select()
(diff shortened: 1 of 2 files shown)


ikrivosheev:
@Sytten thank you for the PR! 
I don't understand... You add a doc test for Postgres backend and for MySQL backend but hide `is` under the feature flag: `backend-sqlite`.

Sytten:
Right I will remove it, I did a stupid copy paste

ikrivosheev:
@Sytten as I know in postgresql we can use `IS TRUE` for example. Also we didn`t want to add feature flag to `Expr`.

Sytten:
So how do we wish to proceed? Just leave it open for any backend?

tyt2y3:
Yes. We don't actually feature guard the frontend in general.

ikrivosheev (approved):
@Sytten thank you! LGTM!

tyt2y3:
Thank you all

PR #349, opened 2022-06-13
Add `CURRENT_TIMESTAMP` function

billy1624:
## PR Info

- Dependents:
  - https://github.com/SeaQL/sea-orm/pull/790

## Adds

- [x] Add `CURRENT_TIMESTAMP` function

## Breaking Changes

- [x] `SimpleExpr::FunctionCall(Function, Vec<SimpleExpr>)` with empty `Vec<SimpleExpr>` now would not generate the empty tuple, i.e. `()`

### DIFF
--- a/src/backend/query_builder.rs
+++ b/src/backend/query_builder.rs
@@ -280,7 +280,9 @@ pub trait QueryBuilder: QuotedBuilder + EscapeBuilder {
             }
             SimpleExpr::FunctionCall(func, exprs) => {
                 self.prepare_function(func, sql, collector);
-                self.prepare_tuple(exprs, sql, collector);
+                if !exprs.is_empty() {
+                    self.prepare_tuple(exprs, sql, collector);
+                }
             }
             SimpleExpr::Binary(left, op, right) => {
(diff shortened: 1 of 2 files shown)


ikrivosheev (approved):
@billy1624 thank you! LGTM!

src/backend/query_builder.rs
```diff
@@ -280,7 +280,9 @@ pub trait QueryBuilder: QuotedBuilder + EscapeBuilder {
             }
             SimpleExpr::FunctionCall(func, exprs) => {
                 self.prepare_function(func, sql, collector);
-                self.prepare_tuple(exprs, sql, collector);
+                if !exprs.is_empty() {
```
tyt2y3, line 283: Basically I don't think it is a good practice. If it's a function, we should make sure it's a `FUNCTION()`.

tyt2y3:
We should remove `CURRENT_TIMESTAMP` from function and make put it under SystemVariable (or SysVar) instead.

billy1624:
Yeah, that make more sense. Basically, `CURRENT_TIMESTAMP` is a system variable (or constant) and `CURRENT_TIMESTAMP()` is a function.

PR #351, opened 2022-06-16
Add support for VALUES lists

### LINKED ISSUE #350: Support VALUES lists in select queries
## Motivation

I'd like to be able to make constant tables from in-memory values via [VALUES lists](https://www.postgresql.org/docs/current/queries-values.html). For example a CTE like

```sql
with groups_cte(id, name) as (
  select * from (values (1, 'hello'), (2, 'world'))
)
select * from users inner join groups_cte on users.group_id = groups_cte.id
```

## Proposed Solutions

Currently [`SelectStatement::from`](https://docs.rs/sea-query/0.25.0/sea_query/query/struct.SelectStatement.html#method.from) accepts an `IntoTableRef`. Given that Postgres itself refers to such lists as constant tables, it could be reasonable to just add the variant `TableRef::Values(Vec<ValueTuple>)` and an impl
```rust
impl<I, V> IntoTableRef for I
where
    I: IntoIterator<Item = V>,
    V: IntoValueTuple,
{}
```
However I'm unaware of other databases support for this; after a very brief search, it looks like MySQL has some variant of this, and SQLite does not. It could perhaps just be implemented a Postgres or Postgres/MySQL specific option.

## Additional Information

Note that VALUES lists are supported on `insert` and `update` currently, just not on `select`. However, those places only allow a single row e.g. `values: Vec<Value>`, and for this to be more useful in select queries, it should really be an iterator of iterators `values: Vec<Vec<Value>>` with some way to enforce the same size inner iterators. I've suggested an iterator of [`ValueTuple`](https://docs.rs/sea-query/0.25.0/sea_query/value/enum.ValueTuple.html)s above, but that type should probably have more instances for larger tuples. A bit of a separate issue, but it might be nice to use the same approach for the `insert`/`update` values, to allow a multi-insert.

samtay:
## PR Info

As title suggests, add support for VALUES lists. I realized in implementation, this is actually supported by all three databases. Here are the [doc test outputs](https://github.com/SeaQL/sea-query/compare/master...samtay:samtay/support-values-lists?expand=1#diff-a8c210c287ba3601876e953b1aa6476a85a31256a48b47d846740cdc6c00f2c7R875-R886) on dbfiddle:

1. [mysql](https://dbfiddle.uk/?rdbms=mysql_8.0&fiddle=92074c22f017a0840d54c66ae858d9aa)
2. [postgres](https://dbfiddle.uk/?rdbms=postgres_14&fiddle=61088d050a79706d7bda9ca2934a0db5)
3. [sqlite](https://dbfiddle.uk/?rdbms=sqlite_3.27&fiddle=61088d050a79706d7bda9ca2934a0db5)

- Closes #350 

## Adds

- [x] Support for VALUES lists in FROM clauses.

## Breaking Changes

- [x] Technically yes, I think, because the enum `TableRef` has a new variant.

### DIFF
--- a/src/backend/mysql/query.rs
+++ b/src/backend/mysql/query.rs
@@ -108,4 +108,8 @@ impl QueryBuilder for MysqlQueryBuilder {
             _ => {}
         };
     }
+
+    fn values_list_tuple_prefix(&self) -> &str {
+        "ROW"
+    }
 }
--- a/src/backend/query_builder.rs
+++ b/src/backend/query_builder.rs
@@ -7,6 +7,11 @@ pub trait QueryBuilder: QuotedBuilder + EscapeBuilder {
         ("?", false)
     }
 
+    /// Prefix for tuples in VALUES list (e.g. ROW for Mysql)
+    fn values_list_tuple_prefix(&self) -> &str {
+        ""
+    }
+
     /// Translate [`InsertStatement`] into SQL statement.
     fn prepare_insert_statement(
         &self,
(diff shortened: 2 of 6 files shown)


tyt2y3 (approved):
Cool. This is quite idomatic!

tyt2y3:
This is quite neat. I wish we could have something like a 'sea-query-fiddle' someday

billy1624:
Hey @samtay, thanks for the digging!

samtay:
No worries, happy to contribute! And thank you guys for the quick merge :bow:

PR #353, opened 2022-06-18
Support LIKE ESCAPE

ikrivosheev:
## PR Info

- Related:
  - https://github.com/SeaQL/sea-query/pull/94
  - https://github.com/SeaQL/sea-query/pull/318
  - https://github.com/SeaQL/sea-query/pull/352

## Adds

1. Move Like from `BinOp` into `SimpleExpr`
2. Add new struct: `LikeExpr`
3. Add new trait: `IntoLikeExpr`

### DIFF
--- a/src/value.rs
+++ b/src/value.rs
@@ -1177,7 +1177,7 @@ pub fn sea_value_to_json_value(value: &Value) -> Json {
         Value::Float(Some(v)) => (*v).into(),
         Value::Double(Some(v)) => (*v).into(),
         Value::String(Some(s)) => Json::String(s.as_ref().clone()),
-        Value::Char(Some(c)) => Json::String(s.to_string()),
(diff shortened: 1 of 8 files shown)


ikrivosheev:
CC @Sytten

tyt2y3:
Thanks! I am about to make so modifications

src/value.rs
```diff
@@ -48,6 +48,7 @@ pub enum Value {
     Float(Option<f32>),
     Double(Option<f64>),
     String(Option<Box<String>>),
+    Char(Option<char>),
```
tyt2y3, line 51: Actually, I think we missed some impl in various drivers?

ikrivosheev: I think we can remove this. This is my mistake...

PR #354, opened 2022-06-19
Feature/support escape v2

tyt2y3:
Continue #352, #353

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -106,6 +106,8 @@ pub enum UnOper {
 pub enum BinOper {
     And,
     Or,
+    Like,
+    NotLike,
     Is,
     IsNot,
     In,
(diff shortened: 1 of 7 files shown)


Sytten:
Looks good!

src/types.rs
```diff
@@ -125,6 +125,7 @@ pub enum BinOper {
     Mul,
     Div,
     As,
+    Escape,
```
ikrivosheev, line 128: @tyt2y3 

I think `LIKE ... ESCAPE ...` is not `BinOper`. If we locking into sqlite doc, for example, we can see that: Like Expression is special case. And so I move it to `SimpleExpr`. `LIKE` Can take only `String` (I cannot find other example).

Sytten: It should definitely not be value as it doesn't need to be escaped with a `?`. Unless I am mistaken it will be in the current implementation.

tyt2y3: Sorry for that. I just thought maintaining a SimpleExpr variant is significantly more cost than maintaining an additional BinOper.

tyt2y3: And that `BETWEEN A AND B` is also a quirky thing in SQL (that has some special treatment in our code generator)

PR #355, opened 2022-06-19
Remove Value::Char

ikrivosheev:
Close comment: https://github.com/SeaQL/sea-query/pull/353#discussion_r901134034


billy1624:
I think it's okay to keep `Value::Char` but we need driver supports for it
- https://github.com/SeaQL/sea-query/tree/master/sea-query-driver/src

ikrivosheev:
@billy1624 thank you! Ok, I add this to the drivers backend.

PR #360, opened 2022-06-25
Fix network types

ikrivosheev:
## PR Info

Query example: 
```rust

let addr = Ipv4Network::new(Ipv4Addr::new(127, 0, 0, 1), 8);
let query = Query::select()
        .column(Network::Ip)
        .from(Network::Table)
        .and_where(Expr::col(Network::Ip).eq(addr))

println!("{}", query);
// Before: SELECT "ip" FROM "network" WHERE "ip" = 127.0.0.1/8 - ERROR!
// After: SELECT "ip" FROM "network" WHERE "ip" = '127.0.0.1/8' - OK
```

## Fixes

- Fix network types convert to string
- Remove unneeded escape character

### DIFF
--- a/src/backend/query_builder.rs
+++ b/src/backend/query_builder.rs
@@ -1159,54 +1159,54 @@ pub trait QueryBuilder: QuotedBuilder + EscapeBuilder {
             }
             Value::Bytes(Some(v)) => write!(
                 s,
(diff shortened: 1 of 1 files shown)


tyt2y3:
Thank you!

PR #361, opened 2022-06-26
Sea query binder

tyt2y3:
Continue #275


tyt2y3:
I don't want to replace all existing examples for now to use sea-query-binder.
The question is how do we actually support 0.5 & 0.6 together, or only support 0.6?

examples/sqlx_any/Cargo.toml
```diff
@@ -25,6 +25,7 @@ sea-query-binder = { path = "../../sea-query-binder", features = [
     "with-bigdecimal",
     "with-uuid",
     "with-time",
+    "postgres-array",
```
tyt2y3, line 28: I added this feature because another crate in the workspace enabled sea-query/postgres-array, and so this cannot build

tyt2y3:
Merged now. Feel free to create follow up PRs

PR #371, opened 2022-07-01
sea-query-binder: Don't panic on big unsigned, convert to i64

nitnelave:
## Fixes

Any big unsigned values (u64 that doesn't fit in i64) will no longer panic but instead will wrap as i64. They can then be read back as u64 to recover the initial value.

Ran into this when storing hashes in SQlite.


ikrivosheev (approved):
@nitnelave thank you! LGTM!

nitnelave:
Sorry, this broke the build. Fix is in https://github.com/SeaQL/sea-query/pull/372.

PR #374, opened 2022-07-02
fix in_tuples bugs

oonxt:
failed when the tuple items has different type

```rust
Query::select()
    .columns([Char::Character, Char::FontId])
    .from(Char::Table)
    .and_where(
        Expr::tuple([
            Expr::col(Char::Character).into_simple_expr(),
            Expr::col(Char::FontId).into_simple_expr(),
        ])
        .in_tuples(vec![[1, String::from("1")], [2,String::from("2")]])
    )
    .to_owned();
```

got a error: expected integer, found struct `std::string::String`

## Fixes

- [x]  in_tuples bug: failed when the tuple items has different type

## Breaking Changes

- [x] in_tuples accept a vec![[Value,]]


tyt2y3:
Sad, there are some errors in CI

oonxt:
@tyt2y3 
my mistake. a doc error and fixed.

Issue #187, opened 2021-11-16
PostgreSQL ipaddr/cidr/macaddress


ajf:
It would be great to see support for the postgres types for `inet`/`cidr`/`macaddr`/`macaddr8`

https://www.postgresql.org/docs/14/datatype-net-types.html

I was poking at this but I have a few questions:

These data types are only built-in on PostgreSQL and appear to be supported by sqlx features `ipnetwork`, `mac_address`.   I assume I have to add features to `sea-query` and `sea-orm`.  I'm mainly just following the code for `uuid`.

Next, what's the general handling of sqllite / mysql databases that doesn't handle this type? Should we ignore it or try to serialize/deserialize to varchar or 32/48/128-bit integers?

`postgres-types`, a dependency of `sea-query` uses `eui48`, while `sqlx` uses `mac_address` to provide mac addresses.  What relation does sea-query have to sqlx? I don't see any sqlx dependency in the query crate, but the ORM crate has a dependency on both sea-query and sqlx.

billy1624:
> These data types are only built-in on PostgreSQL and appear to be supported by sqlx features `ipnetwork`, `mac_address`. I assume I have to add features to `sea-query` and `sea-orm`. I'm mainly just following the code for `uuid`.
> 
> Next, what's the general handling of sqllite / mysql databases that doesn't handle this type? Should we ignore it or try to serialize/deserialize to varchar or 32/48/128-bit integers?
> 
> `postgres-types`, a dependency of `sea-query` uses `eui48`, while `sqlx` uses `mac_address` to provide mac addresses. What relation does sea-query have to sqlx? I don't see any sqlx dependency in the query crate, but the ORM crate has a dependency on both sea-query and sqlx.

ikrivosheev:
@ajf @billy1624 I want to implement this if no one minds)

ajf:
@ikrivosheev I have not started any work, feel free!

billy1624:
Thanks!! @ikrivosheev @ajf

ikrivosheev:
I am working on PR.

ikrivosheev:
https://github.com/SeaQL/sea-query/pull/309 - add support for sqlx. it remains to add support for `postgres-types`

ikrivosheev:
Done: https://github.com/SeaQL/sea-query/pull/503

PR #377, opened 2022-07-03
Fix postgres driver

ikrivosheev:
## PR Info

Postgres driver tests are failed.

## Fixes

- Postgres driver

PR #378, opened 2022-07-03
Feature/issues 187 postgres types

ikrivosheev:
## PR Info

- Closes https://github.com/SeaQL/sea-query/issues/187

- Dependencies:
  - https://github.com/SeaQL/sea-query/pull/377

## Adds

- Add support for `cidr` 
- Add support for `eui48`


sea-query-driver/Cargo.toml
```diff
@@ -33,6 +33,8 @@ with-time = []
 with-ipnetwork = []
 with-mac_address = []
 postgres-array = []
+postgres-cidr = []
+postgres-eui48 = []
```
billy1624, line 37: Hey @ikrivosheev, I think we also need to match the `Value`s of `cidr` and `eui48`? Just like what we did to `Value::Array`:
- https://github.com/SeaQL/sea-query/blob/190eb5041e44173a560bb252f1dd4dd6d07fc77b/sea-query-driver/src/sqlx_postgres.rs#L59-L63

billy1624 (changes requested):

sea-query-driver/Cargo.toml
ikrivosheev: @billy1624 hello! In each driver we have:

```
_ => unimplemented!(),
```

PR #379, opened 2022-07-04
issues-118 Add SubQuery operator

ikrivosheev:
## PR Info

- Closes https://github.com/SeaQL/sea-query/issues/118
- https://github.com/SeaQL/sea-query/issues/216

## Adds

- SubQuery operators: `EXISTS`, `ALL`, `ANY`, `SOME`

### DIFF
--- a/src/expr.rs
+++ b/src/expr.rs
@@ -531,9 +531,9 @@ impl Expr {
     /// ```
     pub fn eq<V>(self, v: V) -> SimpleExpr
     where
-        V: Into<Value>,
+        V: Into<SimpleExpr>,
     {
-        self.bin_oper(BinOper::Equal, SimpleExpr::Value(v.into()))
+        self.bin_oper(BinOper::Equal, v.into())
     }
 
     /// Express a not equal (`<>`) expression.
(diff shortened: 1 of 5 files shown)


ikrivosheev:
@billy1624 @tyt2y3 can you review this?) I have done.

billy1624 (changes requested):
Hey @ikrivosheev, thanks!!

Based on the docs and experiment result. Seems that SQLite only support `EXISTS`. The other three `ANY`, `SOME` and `ALL` are not supported.
- https://sqlite.org/lang_expr.html#the_exists_operator
- https://dbfiddle.uk/?rdbms=sqlite_3.39&fiddle=201457e4ee750fb8972847041a736e19

Also, it'd be best if we have doc test for subquery operators to be used in condition. Such as...
```sql
SELECT s1 FROM t1 WHERE s1 > ANY (SELECT s1 FROM t2);
SELECT s1 FROM t1 WHERE s1 <> ANY  (SELECT s1 FROM t2);
SELECT s1 FROM t1 WHERE s1 <> SOME (SELECT s1 FROM t2);
```

ikrivosheev:
@billy1624 @tyt2y3 hello! I have small question. [Expr::ne](https://docs.rs/sea-query/latest/sea_query/expr/struct.Expr.html#method.ne) have signature: `pub fn ne<V: Into<Value>>(self, v: V) -> SimpleExpr`. If i rewrite this method to: `pub fn ne<V: Into<SimpleExpr>>(self, v: V) -> SimpleExpr` will not be so convenient to use for simple queries. 

What is the best way to name new methods? For example:

1. `ne_expr`?
2. `expr_ne`?
3. Something else?)

billy1624:
Hey @ikrivosheev, good suggestions!! I do think we need `eq_expr` and `ne_expr` inside `Expr`. That's definitely nice to have.

ikrivosheev:
> Hey @ikrivosheev, good suggestions!! I do think we need `eq_expr` and `ne_expr` inside `Expr`. That's definitely nice to have.

@billy1624 hello! Done

src/expr.rs
```diff
@@ -567,7 +567,80 @@ impl Expr {
...
+    ///     query.to_string(PostgresQueryBuilder),
+    ///     r#"SELECT "id" FROM "character" WHERE "id" <> ANY(SELECT "id" FROM "character")"#
+    /// );
+    /// ```
+
+    pub fn neq_expr<T>(self, expr: T) -> SimpleExpr
+    where
+        T: Into<SimpleExpr>,
+    {
+        self.bin_oper(BinOper::NotEqual, expr.into())
```
billy1624, line 643: Nit: we have a blank line here loll

billy1624 (approved):
Thanks!! @ikrivosheev

ikrivosheev:
Hmm, @billy1624, I think about new method... Maybe is better `impl From<T> for SimpleExpr`, where T is all types for `Value`? And after that we don`t need new method. What do you think?

ikrivosheev:
> Hmm, @billy1624, I think about new method... Maybe is better `impl From<T> for SimpleExpr`, where T is all types for `Value`? And after that we don`t need new method. What do you think?

@tyt2y3 what do you think about it?

ikrivosheev:
@billy1624 @tyt2y3 done! I add `impl<T> From<T> for SimpleExpr` and everything works without the new method!

ikrivosheev:
@billy1624 @tyt2y3 can you review this PR?

src/expr.rs
```diff
@@ -2149,6 +2262,15 @@ impl From<Expr> for SelectExpr {
     }
 }
 
+impl<T> From<T> for SimpleExpr
+where
+    T: Into<Value>,
```
tyt2y3, line 2267: Interesting... I was thinking of the potential drawbacks of this blanket impl but could not think of one.
Should improve ergonomics

tyt2y3 (approved):

src/expr.rs
billy1624: This is amazing!

Issue #366, opened 2022-06-30
support upsert where clause


oonxt:
## Motivation

support upsert where clause.

## Proposed Solutions
the code:
```rust
Query::insert()
    .into_table(Glyph::Table)
    .columns([Glyph::Aspect, Glyph::Image])
    .values_panic(vec![
        2.into(),
        3.into(),
    ])
    .on_conflict(
        OnConflict::column(Glyph::Id)
            .update_values([
                (Glyph::Aspect, "04108048005887010020060000204E0180400400".into()),
                (Glyph::Image, 3.1415.into()),
            ])
            .and_where(Expr::col(Char::Aspect).is_null())       //<-- here
            .to_owned()
    )
    .to_owned();

```

## Additional Information

the postgresql doc :
https://www.postgresql.org/docs/11/sql-insert.html

PR #384, opened 2022-07-15
Add support for conditionals to ON CONFLICT

ryanahall:
## PR Info

I make use of partial unique constraints within postgresql quite often, which necessitates the use of `WHERE` clauses in order to specify the appropriate constraint conditions. This is my first attempt at adding support for this clause, so happy to hear any feedback and make appropriate changes.

- Closes https://github.com/SeaQL/sea-query/issues/366

- Dependencies:
  - 

- Dependents:
  - 

## Adds

- Support for `WHERE` within `ON CONFLICT` clauses for `postgresql` 

## Fixes

- [ ] 

## Breaking Changes

- [ ] 

## Changes

- [ ]


ryanahall:
Just realized that `ON CONFLICT WHERE` requires column names to be qualified by their table name. Will investigate

ryanahall:
In postgresql, it seems that `WHERE` is supported for both the `conflict_action` and `conflict_target`, so this change will not work as-is. Will need to investigate more.

src/query/on_conflict.rs
```diff
@@ -1,9 +1,13 @@
...
+use crate::query::condition::*;
+use crate::{
+    ConditionHolder, DynIden, Expr, IntoCondition, IntoIden, LogicalChainOper, SimpleExpr, Value,
+};
 
 #[derive(Debug, Clone, Default)]
 pub struct OnConflict {
     pub(crate) target: Option<OnConflictTarget>,
     pub(crate) action: Option<OnConflictAction>,
+    pub(crate) wherei: ConditionHolder,
```
ikrivosheev, line 10: `r#where`

ikrivosheev (commented):
@ryanahall thank you for the PR! Small comment.

billy1624 (commented):
Hey @ryanahall, sorry for the delay. Thanks for the contributions! The PR looks good!

Seems that SQLite also support ON CONFLICT with condition?
- https://www.sqlite.org/lang_UPSERT.html

billy1624:
Ping @ryanahall :)

ryanahall:
Hey all, apologies for the delay. My original version only accounted for `WHERE` clauses at the action target level. The updated version has support for both `filter` targets and `action` targets, however, it does not make use of the condition trait since the implementations would collide. This seems to work for the time being, albeit with some duplication.

billy1624 (commented):
Hey @ryanahall, thanks for the updates!! Sorry for the delay. I think the `target_*` and `action_*` methods look okay.

Btw... I think SQLite also support ON CONFLICT with conditions. Please check https://www.sqlite.org/syntax/upsert-clause.html

ryanahall:
Thanks @billy1624 , I've removed the unimplemented method and updated the doctests to reflect the new statement outputs.

tyt2y3:
Very cool. My only nit is to separate the doc tests as 'basic usage' and 'advanced usage' instead of piggy-backing on the existing test case (it would reduce the test coverage).

tyt2y3:
I would like to include this in 0.27
If we can resolve the conflict and add a few test cases.

tyt2y3:
Oh interesting, GitHub recognized that this PR is merged

PR #385, opened 2022-07-16
Handle Postgres schema name for `sea-query` statements

### LINKED ISSUE #338: `sea-query` statements should qualify schema name
## Motivation

Linked to [sea-orm issue 521](https://github.com/SeaQL/sea-orm/issues/521): `sea-orm-cli migrate` should respect the `DATABASE_SCHEMA` for Postgres in the specified in a project's `.env` file. This is currently not respected and the migration is done on the `public` schema.

## Proposed Solutions

Update all statement types in `sea-query` to quantify the schema name for Postgres. Check all `stmt` in the link below and see if any table reference is of `Iden` type, it should be changed to `TableRef`.

Also, some statement support schema scope, for example, type name could have schema prefix [e.g. in Postgres](https://www.postgresql.org/docs/current/sql-createtype.html). We can update it from `Iden` to `IdenList`.

https://github.com/SeaQL/sea-orm/blob/183639dc8c817fbd3211a87c67bcb983db4050ed/sea-orm-migration/src/manager.rs#L37-L89

## Additional Information

[See discussion here](https://github.com/SeaQL/sea-orm/issues/521)

nahuakang:
## PR Info

- Closes #338 

## Fixes

This should enable `SchemaManager` in `sea-orm-migration` to inject table reference & schema information for Postgres.

### DIFF
--- a/src/backend/index_builder.rs
+++ b/src/backend/index_builder.rs
@@ -23,12 +23,38 @@ pub trait IndexBuilder: QuotedBuilder {
 
         write!(sql, " ON ").unwrap();
         if let Some(table) = &create.table {
-            table.prepare(sql, self.quote());
+            self.prepare_table_ref(table, sql);
         }
 
         self.prepare_index_type(&create.index_type, sql);
 
         self.prepare_index_columns(&create.index.columns, sql);
+    }
+
+    /// Translate [`TableRef`] into SQL statement.
+    fn prepare_table_ref(&self, table_ref: &TableRef, sql: &mut SqlWriter) {
+        self.prepare_table_ref_common(table_ref, sql);
+    }
+
+    fn prepare_table_ref_common(&self, table_ref: &TableRef, sql: &mut SqlWriter) {
+        match table_ref {
+            TableRef::Table(table) => {
+                table.prepare(sql, self.quote());
+            }
+            TableRef::SchemaTable(schema, table) => {
+                schema.prepare(sql, self.quote());
+                write!(sql, ".").unwrap();
+                table.prepare(sql, self.quote());
+            }
+            TableRef::DatabaseSchemaTable(database, schema, table) => {
+                database.prepare(sql, self.quote());
+                write!(sql, ".").unwrap();
+                schema.prepare(sql, self.quote());
+                write!(sql, ".").unwrap();
+                table.prepare(sql, self.quote());
+            }
+            _ => panic!("Not supported"),
+        }
     }
 
     /// Translate [`IndexDropStatement`] into SQL statement.
--- a/src/backend/index_builder.rs
+++ b/src/backend/index_builder.rs
@@ -1,6 +1,6 @@
 use crate::*;
 
-pub trait IndexBuilder: QuotedBuilder {
+pub trait IndexBuilder: QuotedBuilder + TableRefBuilder {
     /// Translate [`IndexCreateStatement`] into SQL expression.
     fn prepare_table_index_expression(&self, create: &IndexCreateStatement, sql: &mut SqlWriter) {
         self.prepare_index_prefix(create, sql);
@@ -23,7 +23,7 @@ pub trait IndexBuilder: QuotedBuilder {
 
         write!(sql, " ON ").unwrap();
         if let Some(table) = &create.table {
-            self.prepare_table_ref(table, sql);
+            self.prepare_table_ref_index_stmt(table, sql);
         }
 
         self.prepare_index_type(&create.index_type, sql);
@@ -32,7 +32,7 @@ pub trait IndexBuilder: QuotedBuilder {
     }
 
     /// Translate [`TableRef`] into SQL statement.
-    fn prepare_table_ref(&self, table_ref: &TableRef, sql: &mut SqlWriter);
+    fn prepare_table_ref_index_stmt(&self, table_ref: &TableRef, sql: &mut SqlWriter);
 
     /// Translate [`IndexDropStatement`] into SQL statement.
     fn prepare_index_drop_statement(&self, drop: &IndexDropStatement, sql: &mut SqlWriter);
--- a/src/backend/index_builder.rs
+++ b/src/backend/index_builder.rs
@@ -32,30 +32,7 @@ pub trait IndexBuilder: QuotedBuilder {
     }
 
     /// Translate [`TableRef`] into SQL statement.
-    fn prepare_table_ref(&self, table_ref: &TableRef, sql: &mut SqlWriter) {
-        self.prepare_table_ref_common(table_ref, sql);
-    }
-
-    fn prepare_table_ref_common(&self, table_ref: &TableRef, sql: &mut SqlWriter) {
-        match table_ref {
-            TableRef::Table(table) => {
-                table.prepare(sql, self.quote());
-            }
-            TableRef::SchemaTable(schema, table) => {
-                schema.prepare(sql, self.quote());
-                write!(sql, ".").unwrap();
-                table.prepare(sql, self.quote());
-            }
-            TableRef::DatabaseSchemaTable(database, schema, table) => {
-                database.prepare(sql, self.quote());
-                write!(sql, ".").unwrap();
-                schema.prepare(sql, self.quote());
-                write!(sql, ".").unwrap();
-                table.prepare(sql, self.quote());
-            }
-            _ => panic!("Not supported"),
-        }
-    }
+    fn prepare_table_ref(&self, table_ref: &TableRef, sql: &mut SqlWriter);
 
     /// Translate [`IndexDropStatement`] into SQL statement.
     fn prepare_index_drop_statement(&self, drop: &IndexDropStatement, sql: &mut SqlWriter);
--- a/src/backend/postgres/index.rs
+++ b/src/backend/postgres/index.rs
@@ -86,13 +86,6 @@ impl IndexBuilder for PostgresQueryBuilder {
                 write!(sql, ".").unwrap();
                 table.prepare(sql, self.quote());
             }
-            TableRef::DatabaseSchemaTable(database, schema, table) => {
-                database.prepare(sql, self.quote());
-                write!(sql, ".").unwrap();
-                schema.prepare(sql, self.quote());
-                write!(sql, ".").unwrap();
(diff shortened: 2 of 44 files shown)


nahuakang:
@billy1624 @ikrivosheev 👋 Hey guys, I've made [some adjustments](https://github.com/SeaQL/sea-query/pull/385/commits/086bd3b0d7da9b9df2d6edad867eb8d903fd26da) to all table-related statements used by `SchemaManager`:

- `TableCreateStatement`
- `TableAlterStatement`
- `TableDropStatement`
- `TableRenameStatement`
- `TableTruncateStatement` 

Then, I made [some changes](https://github.com/SeaQL/sea-query/pull/385/commits/016a6fe1291780c1b84632bdde3e887d641bfa7d) for index-related statements for `SchemaManager` (including a copy-paste, which I wonder if it's okay to do so):

- `IndexCreateStatement`
- `IndexDropStatement`

Is this going in the right direction? The remaining are the `ForeignKey` and `Type` related statements. 

@billy1624 You mentioned:
> Also, some statement support schema scope, for example, type name could have schema prefix [e.g. in Postgres](https://www.postgresql.org/docs/current/sql-createtype.html). We can update it from `Iden` to `IdenList`.

I'm still not sure where to find this in the code base :D Can you give some more hints? 🙏

src/backend/sqlite/table.rs
```diff
@@ -144,7 +144,7 @@ impl TableBuilder for SqliteQueryBuilder {
         };
         write!(sql, "ALTER TABLE ").unwrap();
         if let Some(table) = &alter.table {
-            table.prepare(sql, self.quote());
+            table_builder::TableBuilder::prepare_table_ref(self, table, sql);
```
ikrivosheev, line 147: Why not to use `self.prepare_table_ref`? And in other places.

ikrivosheev (commented):
@nahuakang thank you for PR! Some comments...

ikrivosheev:
> I'm still not sure where to find this in the code base :D Can you give some more hints? pray

@nahuakang I think @tyt2y3 mean: https://github.com/SeaQL/sea-query/blob/master/src/extension/postgres/types.rs

src/backend/sqlite/table.rs
nahuakang: I think the compiler would be confused:

```
multiple applicable items in scope
multiple `prepare_table_ref` foundrustc[E0034](https://doc.rust-lang.org/error-index.html#E0034)
index_builder.rs(35, 5): candidate #1 is defined in an impl of the trait `backend::index_builder::IndexBuilder` for the type `sqlite::SqliteQueryBuilder`
query_builder.rs(515, 5): candidate #2 is defined in an impl of the trait `query_builder::QueryBuilder` for the type `sqlite::SqliteQueryBuilder`
```

ikrivosheev: @nahuakang ah, yes, thank you!

nahuakang:
@billy1624 Besides the question on `Type`, do we need to change the table from `Option<DynIden>` to `Option<TableRef>` for both `ForeignKeyDropStatement` and `TableForeignKey` (even if the former contains the latter and thus contains `TableRef`)?

```rust
pub struct ForeignKeyDropStatement {
    pub(crate) foreign_key: TableForeignKey,
    pub(crate) table: Option<DynIden>,
}

#[derive(Debug, Clone)]
pub struct TableForeignKey {
    pub(crate) name: Option<String>,
    pub(crate) table: Option<DynIden>,
    pub(crate) ref_table: Option<DynIden>,
    pub(crate) columns: Vec<DynIden>,
    pub(crate) ref_columns: Vec<DynIden>,
    pub(crate) on_delete: Option<ForeignKeyAction>,
    pub(crate) on_update: Option<ForeignKeyAction>,
}
```

And is it safe to assume that schema can be prepended to the table names in any statement involving creating & dropping foreign keys (both the from and to tables)? Thanks :)

billy1624:
Hey @nahuakang @ikrivosheev, yes, I mean https://github.com/SeaQL/sea-query/blob/master/src/extension/postgres/types.rs

billy1624:
> Do we need to change the table from `Option<DynIden>` to `Option<TableRef>` for both `ForeignKeyDropStatement` and `TableForeignKey` (even if the former contains the latter and thus contains `TableRef`)?

Yes, we should change both to `Option<TableRef>`

> And is it safe to assume that schema can be prepended to the table names in any statement involving creating & dropping foreign keys (both the from and to tables)? Thanks :)

This assumption is valid :)

tyt2y3:
Thank you for the hard work! My question is why do we needed `TypeRef` in addition to `TableRef`?
It seems like `TypeRef` is equivalent to `TableRef::DatabaseSchemaTable`.
And so, we can either reuse `TableRef` or make `TypeRef` more similar to TableRef, I mean having three variants, `Name`, `SchemaName` and `DatabaseSchemaName`?

src/backend/index_builder.rs
```diff
@@ -23,14 +23,40 @@ pub trait IndexBuilder: QuotedBuilder {
...
 
         self.prepare_index_columns(&create.index.columns, sql);
     }
 
+    /// Translate [`TableRef`] into SQL statement.
+    fn prepare_table_ref(&self, table_ref: &TableRef, sql: &mut SqlWriter) {
+        self.prepare_table_ref_common(table_ref, sql);
+    }
+
+    fn prepare_table_ref_common(&self, table_ref: &TableRef, sql: &mut SqlWriter) {
```
ikrivosheev, line 39: We talk with you about it. MySql and SQLite don't support schema, only database. I think implementation this method should be backend specific.

src/types.rs
```diff
@@ -97,6 +97,59 @@ pub trait IntoTableRef {
     fn into_table_ref(self) -> TableRef;
 }
 
+#[derive(Clone, Debug)]
+pub struct TypeRef {
```
ikrivosheev, line 101: I though about this and have two idea:
1. This should be part of `extension/postgres` not here
2. Maybe  is better rewrite it to enum:
```rust
pub enum TypeRef {
    Type(...),
    DatabaseType(...),
    DatabaseSchemaType(...)
}
```

ikrivosheev (commented):
@nahuakang thank you for PR! Some comments

nahuakang:
@tyt2y3 Thanks for your review :) I think we shouldn't reuse `TableRef` since we need a specific way to mark a type name that might be schema-qualified. 

If we must have consistency between `TabelRef` and `TypeRef`, perhaps we could do the following (and adjust the `IntoTypeRef` implementations as well as the `prepare_type_ref` for `TypeBuilder`? (wdyt @ikrivosheev since Ivan came up with `TypeRef`?)

```rust
#[derive(Debug, Clone)]
pub enum TypeRef {
    Type(DynIden),
    SchemaType(DynIden, DynIden),
    DatabaseSchemaType(DynIden, DynIden, DynIden),
}
```

ikrivosheev:
> Thank you for the hard work! My question is why do we needed `TypeRef` in addition to `TableRef`? It seems like `TypeRef` a subset of `TableRef`'s variants. And so, we can either reuse `TableRef` or make `TypeRef` more similar to `TableRef`, I mean having three variants, `Name`, `SchemaName` and `DatabaseSchemaName`?

This was my idea. We don't need all the options from `TableRef` and this solution looks more type-safe then `panic` with invalid option

src/types.rs
nahuakang: Sounds good. I'll move it to `extension/postgres`!

src/backend/postgres/index.rs
```diff
@@ -36,6 +36,16 @@ impl IndexBuilder for PostgresQueryBuilder {
...
+        if let Some(table) = &drop.table {
+            match table {
+                TableRef::Table(_) => {}
+                TableRef::SchemaTable(schema, _) => {
+                    schema.prepare(sql, self.quote());
+                    write!(sql, ".").unwrap();
+                }
+                _ => panic!("Not supported"),
+            }
+        }
```
billy1624, line 48: I think `TableRef::DatabaseSchemaTable` should also being supported?

billy1624: Oh... sorry I don't think cross database index is supported

src/backend/postgres/index.rs
```diff
@@ -65,4 +75,25 @@ impl IndexBuilder for PostgresQueryBuilder {
...
+                write!(sql, ".").unwrap();
+                table.prepare(sql, self.quote());
+            }
+            TableRef::DatabaseSchemaTable(database, schema, table) => {
+                database.prepare(sql, self.quote());
+                write!(sql, ".").unwrap();
+                schema.prepare(sql, self.quote());
+                write!(sql, ".").unwrap();
+                table.prepare(sql, self.quote());
+            }
```
billy1624, line 95: I don't think we can create index across database? i.e. 
```sql
CREATE INDEX idx_aaa 
ON aaa.public.accounts (password); # `public.accounts` is valid but not `aaa.public.accounts`
```
_https://dbfiddle.uk/?rdbms=postgres_14&fiddle=67b77e58c1c524174d1675ff92511e2d_

billy1624 (commented):
Hey @nahuakang, all in all it looks good!
Thanks @nahuakang @ikrivosheev for the contrition and help :D

src/backend/postgres/index.rs
nahuakang: That was removed :) Thanks for catching this. I think I probably missed testing this in my test DB 🙏

ikrivosheev (approved):
@nahuakang thank you! LGMT!

billy1624 (commented):
Hey @nahuakang, sorry for the long delay. I think we can refactor it by adding a `TableRefBuilder` trait. To handle the writing of `TableRef` as SQL string. From my perspective that looks cleaner and we don't need to repeat same piece of code again and again.
- https://github.com/nahuakang/sea-query/compare/postgres-table-ref...billy1624:sea-query:pr/385

Please check if that make any sense :)
Thansk!!

nahuakang:
@billy1624 This refactoring makes perfect sense and is a lot cleaner for the hierarchy of the trait requirements. Would you be okay merging that into my branch? :)

tyt2y3:
Thank you for your patience and diligence!

Issue #336, opened 2022-05-23
Query binders: Add support for postgres arrays


nitnelave:
## Motivation

Postgres arrays are not supported in the query binders (see #275) or drivers.

## Proposed Solutions

We should implement the conversion from `Vec<sea-query::Value>` to some `Vec<sqlx::Encode>` in `sea-query-binders/src/sqlx_postgres.rs`. For this, we'll need to ensure that all values are of the same type, and do a second level of dispatch based on the value types to convert them to the right type.

ikrivosheev:
@nitnelave do you want to create PR for this issue?

nitnelave:
I don't actually have a need for this, I just wanted to file this for posterity and/or completeness sake.

billy1624:
Related PRs:
- https://github.com/SeaQL/sea-query/pull/225
- https://github.com/SeaQL/sea-query/pull/205

Sytten:
So what I am wondering is what is our status for keep drivers and binders? I think keeping both is an unnecessary burden.
We discussed in the PR using the binders for the array but sea-orm still uses the driver so it would need to be converted.

Sytten:
I think I got an idea for the implementation:
```rust
use crate::{SqlxValue, SqlxValues};
use sea_query::Value;

impl<'q> sqlx::Encode<'q, sqlx::postgres::Postgres> for SqlxValue {
    fn encode_by_ref(&self, buf: &mut sqlx::postgres::PgArgumentBuffer) -> sqlx::encode::IsNull {
        match &self.0 {
            Value::Bool(b) => {
                <Option<bool> as sqlx::Encode<'q, sqlx::postgres::Postgres>>::encode_by_ref(&b, buf)
            }
            Value::TinyInt(i) => {
                <Option<i8> as sqlx::Encode<'q, sqlx::postgres::Postgres>>::encode_by_ref(&i, buf)
            }
            Value::SmallInt(i) => {
                todo!()
            }
            Value::Int(i) => {
                todo!()
            }
            Value::BigInt(i) => {
                todo!()
            }
            Value::TinyUnsigned(i) => {
                todo!()
            }
            Value::SmallUnsigned(i) => {
                todo!()
            }
            Value::Unsigned(i) => {
                todo!()
            }
            Value::BigUnsigned(i) => {
                todo!()
            }
            Value::Float(f) => {
                todo!()
            }
            Value::Double(d) => {
                todo!()
            }
            Value::String(s) => {
                todo!()
            }
[...]

PR #387, opened 2022-07-17
issues-383 Rewrite examples using sea-query-binder

ikrivosheev:
## Changes

- Rewrite examples using sea-query-binder instead sea-query-driver

PR #390, opened 2022-07-19
issues-336 Init array support

ikrivosheev:
## PR Info

Depends:
- https://github.com/SeaQL/sea-query/pull/387

## Adds

- [x] Support Array

## Todo:

- [x] Implement for all other types
- [x] Fix examples
- [x] Fix tests
- [ ] Add documentation
- [x] Fix `postgres` driver
- [x] Fix `sea-query-driver`
- [x] Fix `sea-query-binder`
- [ ] Remove todo from `CommonQueryBuilder`
- [x] Add methods for work with new `Value` items

### DIFF
--- a/examples/postgres/src/main.rs
+++ b/examples/postgres/src/main.rs
@@ -1,7 +1,7 @@
 use chrono::{DateTime, FixedOffset, NaiveDate, NaiveDateTime};
 use postgres::{Client, NoTls, Row};
 use rust_decimal::Decimal;
-use sea_query::{ColumnDef, Iden, Order, PostgresQueryBuilder, Query, Table};
+use sea_query::{ColumnDef, ColumnType, Iden, Order, PostgresQueryBuilder, Query, Table};
 use sea_query_postgres::PostgresBinder;
 use time::{
     macros::{date, offset, time},
@@ -34,7 +34,7 @@ fn main() {
             .col(ColumnDef::new(Document::Timestamp).timestamp())
             .col(ColumnDef::new(Document::TimestampWithTimeZone).timestamp_with_time_zone())
             .col(ColumnDef::new(Document::Decimal).decimal())
-            .col(ColumnDef::new(Document::Array).array("integer".into()))
+            .col(ColumnDef::new(Document::Array).array(ColumnType::Integer(None)))
             .build(PostgresQueryBuilder),
     ]
     .join("; ");
--- a/sea-query-binder/src/sqlx_any.rs
+++ b/sea-query-binder/src/sqlx_any.rs
@@ -67,70 +67,104 @@ impl<'q> sqlx::IntoArguments<'q, sqlx::any::Any> for SqlxValues {
                 Value::ChronoDate(d) => {
                     args.add(d.map(|d| *d));
                 }
+                #[cfg(all(feature = "with-chrono", feature = "with-array"))]
+                Value::ChronoDateArray(_) => panic!("Array support not implemented for Any"),
                 #[cfg(feature = "with-chrono")]
                 Value::ChronoTime(t) => {
                     args.add(t.map(|t| *t));
                 }
+                #[cfg(all(feature = "with-chrono", feature = "with-array"))]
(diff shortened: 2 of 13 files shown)


ikrivosheev:
@Sytten @billy1624 @tyt2y3 This is Draft PR for add support Array type

Sytten:
Related to https://github.com/SeaQL/sea-query/issues/336

This is basically proposition #2 and it looks well done (great job @ikrivosheev!).
I think it's a decent way to move forward pending a larger refactor as discussed with proposition #3.

ikrivosheev:
I have one question. How can we work with array of arrays?

billy1624:
@Sytten, have any experience working with array of arrays in PostgreSQL?

For me, I don't like the idea of having array datatype in database, let alone array of arrays lolll

We might want to leave array of arrays for later?

billy1624:
The code looks nice! @ikrivosheev please continue implementing array support :)
Thanks!!

Sytten:
@billy1624 I don't think array of arrays is a thing in postgres. I never saw it supported anywhere in any-case so no worries there.

I would say custom types are much more common (these days it is mostly for enums and composite keys) and arrays of custom types do happen once in a while. It is not as common now that json support is very well done, but it can happen and so it should be supported at some point.

ikrivosheev:
@Sytten I add support for array to **sea-query-binder**.

ikrivosheev:
@Sytten hello! I have some features:

1. Add support for array in `sea-query-binder`
2. Add support for array in `sea-query-driver`
3. Add support for array in `driver/postgres`

Can you review this part?

Sytten:
Will do tomorrow, pretty busy today. Ping me again if I forget

Sytten (commented):
I don't have much to say at the moment, it is headed in the right direction for this support.
That being said I am not sure we really want to do this in the current state of things.
Personally I would move to a trait based system instead of the enum and I would consolidate the driver/binder mess before thinking about array support.

examples/sqlx_mysql/src/main.rs
```diff
@@ -87,11 +86,9 @@ async fn main() {
                 .into(),
             date!(2020 - 8 - 20).with_time(time!(0:0:0)).into(),
         ])
-        .build(MysqlQueryBuilder);
+        .build_sqlx(MysqlQueryBuilder);
```
tyt2y3, line 89: Quick question. Where is this `build_sqlx` defined? Why I don't seem to find it?

billy1624:
> I don't have much to say at the moment, it is headed in the right direction for this support. That being said I am not sure we really want to do this in the current state of things. Personally I would move to a trait based system instead of the enum and I would consolidate the driver/binder mess before thinking about array support.

In the long run, I support adding a Value trait and redesigning the Value API from ground up. However, at the moment I think we can extend the functionality based on (current) Value enum.

Any comments? @tyt2y3

examples/sqlx_mysql/src/main.rs
Sytten: In the binder crate

ikrivosheev:
> > I don't have much to say at the moment, it is headed in the right direction for this support. That being said I am not sure we really want to do this in the current state of things. Personally I would move to a trait based system instead of the enum and I would consolidate the driver/binder mess before thinking about array support.
> 
> In the long run, I support adding a Value trait and redesigning the Value API from ground up. However, at the moment I think we can extend the functionality based on (current) Value enum.
> 
> Any comments? @tyt2y3 

I think about the current implementation and find it bad. A lot of code... Very hard to maintain. 

@billy1624 if you want, I can create PR with my peace of code.

tyt2y3:
It's quite a lot of boilerplate indeed. `ValueTrait` would help, but we should keep the primitive types in a enum, where i64 (and anything smaller than 8 bytes) should reside on the stack. For Object types (which we Boxed), we can probably use a fat pointer `Box<&dyn ValueTrait>` such that it allows users to impl that on their own.

ikrivosheev:
I close this PR. Better solution is: https://github.com/SeaQL/sea-query/pull/467

PR #392, opened 2022-07-22
add `AnyStatement` enum

kirawi:
Taken from https://github.com/SeaQL/sea-orm/pull/863

I think this might be a necessary change for the `QueryLinter`. I wasn't sure whether to use a macro to implement a `From<_> for AnyStatement` for all the statement types.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -432,6 +432,29 @@ impl IntoLikeExpr for String {
     }
 }
 
+#[derive(Debug)]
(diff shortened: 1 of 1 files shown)


billy1624 (commented):
Hey @kirawi, I think you could just keep this `AnyStatement` enum inside `sea-orm-linter`. And, it's perfectly valid to implement `From<_> for AnyStatement` on `sea-orm-linter`.

Issue #393, opened 2022-07-25
PostgreSQL: `modify_column().null()` does not make column nullable


korewaChino:
## Description

When using PostgreSQL, `modify_column()` uses `SET NULL` instead of `DROP NOT NULL` when making an existing column nullable. This causes it to not do anything.
## Steps to Reproduce
```rust
manager
            .alter_table(
                sea_query::Table::alter()
                    .table(Target::Table)
                    .modify_column(ColumnDef::new(Target::Arch).string().null())
                    .to_owned(),
            )
            .await?;
```

### Expected Behavior
Query should use `DROP NOT NULL` when making a not nullable column nullable

### Actual Behavior
Query uses `SET NULL`, causing nothing to happen.

### Reproduces How Often
Always

## Versions
```
├── sea-orm-migration v0.9.1
│   ├── sea-orm v0.9.1
│   │   ├── sea-orm-macros v0.9.1 (proc-macro)
│   │   ├── sea-query v0.26.2
│   │   │   ├── sea-query-derive v0.2.0 (proc-macro)
│   │   │   ├── sea-query-driver v0.2.0 (proc-macro)
│   │   ├── sea-strum v0.23.0
│   │   │   └── sea-strum_macros v0.23.0 (proc-macro)
│   ├── sea-orm-cli v0.9.1
│   │   ├── sea-schema v0.9.3
│   │   │   ├── sea-query v0.26.2 (*)
│   │   │   └── sea-schema-derive v0.1.0 (proc-macro)
│   ├── sea-schema v0.9.3 (*)
├── sea-orm v0.9.1 (*)
├── sea-orm-rocket v0.5.0
│   └── sea-orm-rocket-codegen v0.5.0 (proc-macro)
```

## Additional Information
PostgreSQL, Fedora Linux 36

tyt2y3:
Quick question, what'd happen when the column itself does not have `NOT NULL` and we execute `DROP NOT NULL`? Is it a no-op or an error?

korewaChino:
> Quick question, what'd happen when the column itself does not have `NOT NULL` and we execute `DROP NOT NULL`? Is it a no-op or an error?

No-op.

tyt2y3:
Nice

PR #394, opened 2022-07-25
issues-393 Fix DROP NOT NULL

ikrivosheev:
## PR Info

- Closes https://github.com/SeaQL/sea-query/issues/393

## Fixes

- Fix making an existing column nullable


tyt2y3:
Great!

PR #400, opened 2022-08-01
Update SQLite backend Boolean type from int to bool

### LINKED ISSUE #375: SQLite: Booleans become integer columns
## Description

When using SQLite, boolean data types are converted to integer.

Whilst this is somewhat fair enough (SQLite doesn't have a native boolean type; it expects you to use an integer with 0 and 1), it means that `sea-orm-cli generate entity` gives you a `i32` field rather than a `bool` field, which is ... sad.

This conflicts a little bit with the 'schema-first' goal of this project, as we lose information by encoding it into the schema and then decoding it again.

See https://github.com/SeaQL/sea-query/blob/1467e1d8b7a6da0bc23dfd35fcdfe7a3ac1803a9/src/backend/sqlite/table.rs#L105.

## Steps to Reproduce

1. Create a schema migration including a boolean type, e.g.:
```rust
        manager
            .create_table(
                Table::create()
                    .table(TimelineSegment::Table)
                    .if_not_exists()
                    .col(
                        ColumnDef::new(TimelineSegment::SegmentId)
                            .integer()
                            .not_null()
                            .auto_increment()
                            .primary_key(),
                    )
                    .col(
                        ColumnDef::new(TimelineSegment::PrevGappy)
                            .boolean()
                            .not_null(),
                    )
                    .to_owned(),
            )
            .await?;
```
2. Apply the migration
3. Generate entities using `sea-orm-cli generate entity --database-url sqlite://$(realpath ../../../devdb.sqlite3)`

### Expected Behavior

I should have obtained this entity:
```rust
#[derive(Clone, Debug, PartialEq, DeriveEntityModel)]
#[sea_orm(table_name = "timeline_segment")]
pub struct Model {
    #[sea_orm(primary_key)]
    pub segment_id: i32,
    pub prev_gappy: bool,
}
```

### Actual Behavior

I obtained this entity:
```rust
#[derive(Clone, Debug, PartialEq, DeriveEntityModel)]
#[sea_orm(table_name = "timeline_segment")]
pub struct Model {
    #[sea_orm(primary_key)]
    pub segment_id: i32,
    pub prev_gappy: i32,  // <---- !!!
}
```

SQLite's schema shows the reason why quite clearly:
```sql
sqlite> .schema
CREATE TABLE IF NOT EXISTS "seaql_migrations" ( "version" text NOT NULL PRIMARY KEY, "applied_at" integer NOT NULL );
CREATE TABLE IF NOT EXISTS "timeline_segment" ( "segment_id" integer NOT NULL PRIMARY KEY AUTOINCREMENT, "prev_gappy" integer NOT NULL );
```

### Reproduces How Often

Always

## Versions

```
│   └── sea-orm v0.8.0
│       ├── sea-orm-macros v0.8.0 (proc-macro)
│       ├── sea-query v0.24.6
│       │   ├── sea-query-derive v0.2.0 (proc-macro)
│       │   ├── sea-query-driver v0.1.1 (proc-macro)
│       ├── sea-strum v0.23.0
│       │   └── sea-strum_macros v0.23.0 (proc-macro)
├── sea-orm v0.8.0 (*)
├── sea-query v0.25.2
│   └── sea-query-derive v0.2.0 (proc-macro) (*)
```

Using SQLite on Ubuntu.

## Additional Information

If we think maintaining the integer affinity is important, we could be cheeky and call the type `integerboolean` or something like that, which the entity generator could recognise to mean boolean.

The algorithm at https://www.sqlite.org/datatype3.html#determination_of_column_affinity suggests that will be absolutely fine as long as the name contains 'int'.

anshulxyz:
As mentioned in #375 the SQLite engine already supports boolean type implicitly. More details are provided in the issue's comments about this.

## PR Info

- Closes #375 

## Adds

- [ ] Explicit support for booleans in SQLite backend

## Breaking Changes

- [ ] I don't think this is backwards compatible, because after this update the Entity will look for a `boolean`, until now we were using `int` there. So in case, someone makes an update, and they provide and int where we expect a bool, it might break some code. This still needs to be checked.
- [ ] Immediate change the this PR should bring is the difference in generated Entity's field type should change from `i32` to `boolean`.

## Changes

- [x] I have added a few unit test cases related to this PR, for SQLite's bool support

### DIFF
--- a/src/backend/sqlite/table.rs
+++ b/src/backend/sqlite/table.rs
@@ -102,7 +102,7 @@ impl TableBuilder for SqliteQueryBuilder {
                     _ => "blob".into(),
                 },
                 ColumnType::VarBinary(length) => format!("binary({})", length),
-                ColumnType::Boolean => "bool".into(),
+                ColumnType::Boolean => "boolean".into(),
                 ColumnType::Money(precision) => match precision {
                     Some((precision, scale)) => format!("integer({}, {})", precision, scale),
                     None => "integer".into(),
--- a/src/backend/sqlite/table.rs
+++ b/src/backend/sqlite/table.rs
@@ -102,7 +102,7 @@ impl TableBuilder for SqliteQueryBuilder {
                     _ => "blob".into(),
                 },
                 ColumnType::VarBinary(length) => format!("binary({})", length),
-                ColumnType::Boolean => "integer".into(),
+                ColumnType::Boolean => "bool".into(),
                 ColumnType::Money(precision) => match precision {
                     Some((precision, scale)) => format!("integer({}, {})", precision, scale),
                     None => "integer".into(),
--- a/src/tests_cfg.rs
+++ b/src/tests_cfg.rs
@@ -108,3 +108,30 @@ impl Iden for Glyph {
         .unwrap();
     }
 }
+
+/// Representation of a database table named `Task`.
+///
+/// A `Enum` implemented [`Iden`] used in rustdoc and test to demonstrate the library usage.
+///
+/// [`Iden`]: crate::types::Iden
+#[derive(Debug)]
+pub enum Task {
+    Table,
+    Id,
+    IsDone,
+}
+
+impl Iden for Task {
+    fn unquoted(&self, s: &mut dyn FmtWrite) {
+        write!(
+            s,
+            "{}",
+            match self {
+                Self::Table => "task",
+                Self::Id => "id",
+                Self::IsDone => "is_done",
+            }
+        )
+        .unwrap();
+    }
+}
--- a/tests/sqlite/query.rs
+++ b/tests/sqlite/query.rs
@@ -1085,6 +1085,36 @@ fn insert_8() {
 }
 
 #[test]
+#[allow(clippy::approx_constant)]
+fn insert_9() {
+    assert_eq!(
+        Query::insert()
+            .into_table(Task::Table)
+            .columns([Task::IsDone])
+            .values_panic(vec![
+                true.into(),
+            ])
+            .to_string(SqliteQueryBuilder),
+        r#"INSERT INTO "task" ("is_done") VALUES (TRUE)"#
+    );
+}
+
+#[test]
+#[allow(clippy::approx_constant)]
+fn insert_10() {
+    assert_eq!(
+        Query::insert()
+            .into_table(Task::Table)
+            .columns([Task::IsDone])
+            .values_panic(vec![
+                false.into(),
+            ])
+            .to_string(SqliteQueryBuilder),
+        r#"INSERT INTO "task" ("is_done") VALUES (FALSE)"#
+    );
+}
+
+#[test]
 fn insert_from_select() {
     assert_eq!(
         Query::insert()
--- a/tests/sqlite/table.rs
+++ b/tests/sqlite/table.rs
@@ -148,6 +148,30 @@ fn create_5() {
 }
 
 #[test]
+fn create_6() {
+    assert_eq!(
+        Table::create()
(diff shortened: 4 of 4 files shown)


src/backend/sqlite/table.rs
```diff
@@ -102,7 +102,7 @@ impl TableBuilder for SqliteQueryBuilder {
                     _ => "blob".into(),
                 },
                 ColumnType::VarBinary(length) => format!("binary({})", length),
-                ColumnType::Boolean => "integer".into(),
+                ColumnType::Boolean => "bool".into(),
```
billy1624, line 105: Hey @anshulxyz, sorry for the long delay! Thank you so much for the contributions!

I agree with this breaking changes. But I prefer mapping it as `boolean`. Just to be aligned with the type affinity listed on SQLite docs: https://www.sqlite.org/datatype3.html#affinity_name_examples

```suggestion
                ColumnType::Boolean => "boolean".into(),
```

billy1624:
Context:
image

---

Hey @Anshul#9329, if you look at the rules of type affinity

```
For tables not declared as STRICT, the affinity of a column is determined by the declared type of the column, according to the following rules in the order shown:

1. If the declared type contains the string "INT" then it is assigned INTEGER affinity.

2. If the declared type of the column contains any of the strings "CHAR", "CLOB", or "TEXT" then that column has TEXT affinity. Notice that the type VARCHAR contains the string "CHAR" and is thus assigned TEXT affinity.

3. If the declared type for a column contains the string "BLOB" or if no type is specified then the column has affinity BLOB.

4. If the declared type for a column contains any of the strings "REAL", "FLOA", or "DOUB" then the column has REAL affinity.

5. Otherwise, the affinity is NUMERIC.
```
_https://www.sqlite.org/datatype3.html#determination_of_column_affinity_

The column defined with `bool` / `boolean` datatype are essentially treated as `NUMERIC` inside SQLite internal.

I prefer `boolean` over `bool` because of two reasons:

1. I mentioned above - Just to be aligned with the type affinity listed on SQLite docs: https://www.sqlite.org/datatype3.html#affinity_name_examples
2. We don't need to update any code inside `sea-schema` to parse the boolean column correctly. Because it's searching for `boolean` instead of `bool`: https://github.com/SeaQL/sea-schema/blob/1d377ed9ef33e3b0bd090c35b54c9c893e3076f2/src/sqlite/def/types.rs#L72

anshulxyz:
@billy1624 Thanks you for explaining. I'd totally missed this part of the table.
Screenshot 2022-08-09 at 5 15 33 PM

I've already renamed it to `boolean` in previous commits. What should be the next steps from here?

billy1624:
> I've already renamed it to `boolean` in previous commits. What should be the next steps from here?

I think it's all good! After this PR is being merged, we should mention this bareaking change in the CHANGELOG. This PR should fix the original issue - https://github.com/SeaQL/sea-query/issues/375 - once we bumped the SeaORM version on SeaSchema side.

billy1624 (approved):
Thanks again for contributing!! @anshulxyz

ikrivosheev:
@billy1624 should we make fix in sea-schema?

tyt2y3:
> We don't need to update any code inside sea-schema to parse the boolean column correctly. Because it's searching for boolean instead of bool: https://github.com/SeaQL/sea-schema/blob/1d377ed9ef33e3b0bd090c35b54c9c893e3076f2/src/sqlite/def/types.rs#L72

According to my understanding, it will 'just work' with SeaSchema

billy1624:
> According to my understanding, it will 'just work' with SeaSchema. Am I right?

Hey @tyt2y3, correct!

PR #410, opened 2022-08-08
exclude chrono default-features

robjtede:
in an effort to remove time v0.1 from the dep graph

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,7 +43,7 @@ sea-query-derive = { version = "^0.2.0", path = "sea-query-derive", optional = t
 sea-query-driver = { version = "^0.2.0", path = "sea-query-driver", optional = true }
 serde_json = { version = "^1", optional = true }
 bytes = { version = "^1", optional = true }
-chrono = { version = "^0", optional = true }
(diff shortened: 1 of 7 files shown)


robjtede (commented):
(oops did this in github.dev by accident)

ikrivosheev:
@robjtede, hello! Can you explane, what you want?

robjtede:
`cargo audit` compains about vulnerabilities in time v0.1 which is enabled transitively via chrono when the `oldtime` feature is enabled. This PR disables that feature of chrono.

billy1624 (commented):
Hey @robjtede, thanks for the auditing!

Same here :)
I would suggest pinning the version to `^0.4` insead of `^0`

robjtede:
>  I would suggest pinning the version to ^0.4 insead of ^0

yep so would I but i've left them all as they were since it's not the purpose of this PR

billy1624:
Oh... sorry I thought it was `^0.4` but in fact originally it's `^0`

tyt2y3:
Thank you for letting us know @robjtede

PR #413, opened 2022-08-11
issues-336 Array support

ikrivosheev:
## PR Info

This is continue: https://github.com/SeaQL/sea-query/pull/390


src/value/types.rs
```diff
@@ -0,0 +1,126 @@
...
+    fn to_sql_string(&self) -> String;
+}
+
+pub trait ValueType: Sized {
+    fn type_name() -> String;
+    fn column_type() -> ColumnType;
+}
+
+#[derive(Debug, Clone)]
+pub struct Value(Rc<Box<dyn ValueTrait>>);
```
billy1624, line 33: Hey @ikrivosheev, thanks for the draft!!

One question come up in my mind. For `sea-query-binder` and `sea-query-driver`, how do we bind the value on db driver? Because we can't figure out what's the type that implemented `ValueTrait` without the help of [`downcast`](https://doc.rust-lang.org/std/any/trait.Any.html#method.downcast) and `Any` trait.

billy1624: @tyt2y3 ideas?

ikrivosheev: @billy1624 this is the biggest problem... Another idea is move `sqlx` and `driver` into sea-query

ikrivosheev: @Sytten ideas? 

@billy1624 I can make two PR with with `downcast` and with move `sqlx` and `driver` into sea-query.

billy1624: The chain of pings loll

billy1624: I wonder what's the performance implication of using `downcast`?

Sytten: This what I discussed on discord: https://discord.com/channels/873880840487206962/900758332187484160/1007648248745447524

ikrivosheev: @tyt2y3 I think, I undestand!

tyt2y3: > keeping the enum when you have the boxed trait

I wanted to put primitives (and to some extend anything that is less than 128 bits and impl Copy) on the stack.

tyt2y3:
Perhaps we can start from removing the object variants from Value (while keeping the primitive variants)?

src/value/types.rs
billy1624: I imagine it's like

```rs
// Inside the root of sea-query

trait ValueTrait {}

enum SeaValue {
    I32(i32),
    Object(Box<dyn ValueTrait>),
}

impl ValueTrait for SeaValue {}

// Inside sea-orm-binder

struct SqlxMySqlValue<T>(T)
where
    T: ValueTrait + for<'q> Encode<'q, MySql>;

impl<'p, T> Encode<'p, MySql> for SqlxMySqlValue<T>
where
    T: ValueTrait + for<'q> Encode<'q, MySql>,
{
    fn encode_by_ref(
        &self,
        buf: &mut <MySql as sqlx::database::HasArguments<'_>>::ArgumentBuffer,
    ) -> sqlx::encode::IsNull {
        self.0.encode_by_ref(buf)
    }
}
```

billy1624: And, maybe...

```rs
trait ValueTrait {}

enum SeaValue {
    I32(i32),
    Object(Box<dyn ValueTrait>),
}

impl ValueTrait for SeaValue {}

impl<'p> Encode<'p, MySql> for dyn ValueTrait
where
    Self: for<'q> Encode<'q, MySql>,
{
    fn encode_by_ref(
        &self,
        buf: &mut <MySql as sqlx::database::HasArguments<'p>>::ArgumentBuffer,
    ) -> sqlx::encode::IsNull {
        self.encode_by_ref(buf)
    }
}
```

billy1624:
The requirements:
1. No downcast
2. Keep SeaQuery clean, outsource heavy lifting of value binding to binder
3. Include binder in sea-query's dependency but do not turn on the features by default
4. It could be `Value` in sea-query, `ValueTrait` in binder
5. Could we enable multiple backend support in an additive compatible way? I.e. if one downstream turned on [SQLx](https://crates.io/crates/sqlx) and another turn on [postgres](https://crates.io/crates/postgres). The `?` cargo syntax should be helpful

src/value/types.rs
ikrivosheev: @billy1624 it seems like it won't work...

```
// Inside the root of sea-query

trait ValueTrait {}

enum Value {
    I32(i32),
    Object(Box<dyn ValueTrait>),
}

impl ValueTrait for Value {}

// Inside sea-query-rusqlite

#[derive(Clone, Debug)]
struct RusqliteValue<T>(T)
where
    T: ValueTrait + ToSql;

impl<T> ToSql for RusqliteValue<T>
where
    T: ValueTrait + ToSql,
{
    fn to_sql(&self) -> Result<ToSqlOutput<'_>> {
        self.0.to_sql()
    }
}

#[derive(Clone, Debug)]
pub struct RusqliteValues<T>(pub Vec<RusqliteValue<T>>);  // <--- I cannot put here enum Value because it does not impl ToSql and I cannot make this parameter generic and I cannot impl ToSql for Value because it from other trait
```

ikrivosheev:
@billy1624 @tyt2y3 , hello! I have a bad surprise. `trait ToSql` from `postgres-types` crate need: `Self: Sized`, but I cannot make `ValueTrait: Sized`, because I need `dyn ValueTrait`

ikrivosheev:
It looks like instead of create our own `ValueTrait` make all types, which hold `Value` should be generic...

ikrivosheev:
Hmmm another sought... Black magic with unsafe. This should work. I can store row pointer to data and then transmute it to value and type get from enum.

tyt2y3:
May be we should postpone `ValueTrait` to after 0.27, and trying to make array working with `sqlx-postgres` for now

ikrivosheev:
Ah, one more problem... I try something like this:

```rust
pub trait Sqlx<'q, T: Database>: Encode<'q, T> + Type<T> {}

pub trait ValueTrait
where
    Self: Debug,
{
    fn to_sql_string(&self) -> String;

    #[cfg(feature = "with-postgres")]
    fn as_postgres_to_sql(&self) -> &(dyn PostgresToSql + Sync + Send);

    #[cfg(feature = "sqlx-mysql")]
    fn as_sqlx_mysql<'q>(&'q self) -> &(dyn Sqlx<'q, MySql>);

    #[cfg(feature = "sqlx-postgres")]
    fn as_sqlx_postgres<'q>(&'q self) -> &(dyn Sqlx<'q, Postgres>);

    #[cfg(feature = "sqlx-sqlite")]
    fn as_sqlx_sqlite<'q>(&'q self) -> &(dyn Sqlx<'q, Sqlite>);
}
```

Problems:

1. I cannot combine `rusqlite` and `sqlx` in one crate (`failed to select a version for `libsqlite3-sys`)
2. I cannot make `&dyn Sqlx` because `Type<T>` - ``Sqlx` cannot be made into an object`
3. `sea-query-postgres` throw compile error: `the `to_sql` method cannot be invoked on a trait object`

tyt2y3:
1: I think we won't combine them
2: we have to create our own trait based on that but is object safe

billy1624:
I guess we can close this?

ikrivosheev:
> I guess we can close this?

@billy1624 I think, no, we cannot... I will continue work on it.

billy1624:
Alright, a trait based `Value` with array support would be cool. And a huge step forward :)

PR #417, opened 2022-08-13
Change `cond_where` semantics

### LINKED ISSUE #414: Adding `filter`s after adding a `Condition::any` in another call results in `OR` instead of `AND`
## Description
When querying data, adding a filter using a `Condition::any` causes any further calls to `filter` to also be part of that `any` condition.

## Steps to Reproduce

1. Create a query with a filter using a `Condition::any`
2. Add another condition in a separate call to `filter`

### Expected Behavior
I'd expect this:
```rust
Entity::find()
        .filter(
            Condition::any()
                .add(Column::Field2.eq("dog"))
                .add(Column::Field3.eq("dog")),
        )
        .filter(Column::Field1.eq(1))
```
to be equivalent to this:
```rust
Entity::find()
        .filter(Column::Field1.eq(1))
        .filter(
            Condition::any()
                .add(Column::Field2.eq("dog"))
                .add(Column::Field3.eq("dog")),
        )
```

### Actual Behavior

The first produces:
```sql
SELECT `table`.`field1`, `table`.`field2`, `table`.`field3` FROM `table` WHERE `table`.`field2` = 'dog' OR `table`.`field3` = 'dog' OR `table`.`field1` = 1
```
While the second produces the expected output of:
```sql
SELECT `table`.`field1`, `table`.`field2`, `table`.`field3` FROM `table` WHERE `table`.`field1` = 1 AND (`table`.`field2` = 'dog' OR `table`.`field3` = 'dog')
```

### Reproduces How Often
Every time

## Versions
```
sea-orm v0.9.1
    ├── sea-orm-macros v0.9.1 (proc-macro)
    ├── sea-query v0.26.2
    │   ├── sea-query-derive v0.2.0 (proc-macro)
    │   ├── sea-query-driver v0.2.0 (proc-macro)
    ├── sea-strum v0.23.0
    │   └── sea-strum_macros v0.23.0 (proc-macro)
```

Ubuntu 22.04 on WSL in Windows 11

## Additional Information

<https://github.com/e-rhodes/sea-orm-filter-and-or>

tyt2y3:
resolves #414

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,62 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Breaking changes
 
 * Removed `join_alias`
-* Change `in_tuples` interface to accept `IntoValueTuple` https://github.com/SeaQL/sea-query/pull/386
+* Changed `in_tuples` interface to accept `IntoValueTuple` https://github.com/SeaQL/sea-query/pull/386
+* **Changed `cond_where` semantics**
+```rust
+// Before: will extend current Condition
+assert_eq!(
+    Query::select()
+        .cond_where(
+            Cond::any()
+                .add(Expr::col(Glyph::Id).eq(1))
+                .add(Expr::col(Glyph::Id).eq(2)),
+        )
+        .cond_where(Expr::col(Glyph::Id).eq(3))
+        .to_owned()
+        .to_string(PostgresQueryBuilder),
+    r#"SELECT WHERE "id" = 1 OR "id" = 2 OR "id" = 3"#
+);
+// Before: confusing, since it depends on the order of invocation:
+assert_eq!(
+    Query::select()
+        .cond_where(Expr::col(Glyph::Id).eq(3))
+        .cond_where(
+            Cond::any()
+                .add(Expr::col(Glyph::Id).eq(1))
+                .add(Expr::col(Glyph::Id).eq(2)),
+        )
+        .to_owned()
+        .to_string(PostgresQueryBuilder),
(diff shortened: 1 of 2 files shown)


tyt2y3:
@e-rhodes

CHANGELOG.md
```diff
@@ -14,7 +14,62 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Breaking changes
 
 * Removed `join_alias`
-* Change `in_tuples` interface to accept `IntoValueTuple` https://github.com/SeaQL/sea-query/pull/386
+* Changed `in_tuples` interface to accept `IntoValueTuple` https://github.com/SeaQL/sea-query/pull/386
+* **Changed `cond_where` semantics**
+```rust
+// Before: will extend current Condition
+assert_eq!(
+    Query::select()
```
ikrivosheev, line 22: @tyt2y3 can you rewrite examples using macros: `all!` and `any!`? 

I think it's easier to read)

ikrivosheev (commented):
@tyt2y3 hello! Some comments)

CHANGELOG.md
tyt2y3: changed

Sytten:
Can we add a test for backward compatibility where you first put a `.cond_where(Condition::all())` then a couple of filters.

tyt2y3:
> Can we add a test for backward compatibility where you first put a `.cond_where(Condition::all())` then a couple of filters.

Not quite got that. This is a breaking change, and the behaviour only affect the case where `Condition::any()` is preceding. I already added some more doc tests. Is that what you need?

PR #421, opened 2022-08-16
fix #420

### LINKED ISSUE #420: It is meaningless to pass the object ownership of generic type, which can be passed by reference
```diff
pub trait QueryStatementWriter: QueryStatementBuilder {
    /// Build corresponding SQL statement for certain database backend and return SQL string
    ///
    /// # Examples
    ///
    /// ```
    /// use sea_query::{*, tests_cfg::*};
    ///
    /// let query = Query::select()
    ///     .column(Glyph::Aspect)
    ///     .from(Glyph::Table)
    ///     .and_where(Expr::expr(Expr::col(Glyph::Aspect).if_null(0)).gt(2))
    ///     .order_by(Glyph::Image, Order::Desc)
    ///     .order_by_tbl(Glyph::Table, Glyph::Aspect, Order::Asc)
    ///     .to_string(MysqlQueryBuilder);
    ///
    /// assert_eq!(
    ///     query,
    ///     r#"SELECT `aspect` FROM `glyph` WHERE IFNULL(`aspect`, 0) > 2 ORDER BY `image` DESC, `glyph`.`aspect` ASC"#
    /// );
    /// ```
+    fn to_string<T: QueryBuilder>(&self, query_builder: T) -> String {
+       let (sql, values) = self.build_any(&query_builder);
        inject_parameters(&sql, values.0, &query_builder)
    }
...
```

Because next step, we give up `T` ownership, so suggestion: use `reference type: &T` or use `&dyn QueryBuilder` more than friendly.

```diff
+    fn to_string(&self, query_builder: &dyn QueryBuilder) -> String {
+       let (sql, values) = self.build_any(&query_builder);
        inject_parameters(&sql, values.0, &query_builder)
    }
```

Then, how to use:
```diff
    /// let query = Query::select()
    ///     .column(Glyph::Aspect)
    ///     .from(Glyph::Table)
    ///     .and_where(Expr::expr(Expr::col(Glyph::Aspect).if_null(0)).gt(2))
    ///     .order_by(Glyph::Image, Order::Desc)
    ///     .order_by_tbl(Glyph::Table, Glyph::Aspect, Order::Asc)
+    ///     .to_string(&MysqlQueryBuilder);
```

baoyachi:
fix #420 

Then, sea-orm call `sea-query` crate, not always use Object owership with `use sea_query::{MysqlQueryBuilder, PostgresQueryBuilder, SqliteQueryBuilder};`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,7 @@ assert_eq!(
         .cond_where(any![Expr::col(Glyph::Id).eq(1), Expr::col(Glyph::Id).eq(2)])
         .cond_where(Expr::col(Glyph::Id).eq(3))
         .to_owned()
-        .to_string(PostgresQueryBuilder),
+        .to_string(&PostgresQueryBuilder),
     r#"SELECT WHERE "id" = 1 OR "id" = 2 OR "id" = 3"#
 );
 // Before: confusing, since it depends on the order of invocation:
@@ -36,7 +36,7 @@ assert_eq!(
         .cond_where(Expr::col(Glyph::Id).eq(3))
         .cond_where(any![Expr::col(Glyph::Id).eq(1), Expr::col(Glyph::Id).eq(2)])
         .to_owned()
-        .to_string(PostgresQueryBuilder),
+        .to_string(&PostgresQueryBuilder),
     r#"SELECT WHERE "id" = 3 AND ("id" = 1 OR "id" = 2)"#
 );
 // Now: will always conjoin with `AND`
@@ -45,7 +45,7 @@ assert_eq!(
         .cond_where(Expr::col(Glyph::Id).eq(1))
         .cond_where(any![Expr::col(Glyph::Id).eq(2), Expr::col(Glyph::Id).eq(3)])
         .to_owned()
-        .to_string(PostgresQueryBuilder),
+        .to_string(&PostgresQueryBuilder),
     r#"SELECT WHERE "id" = 1 AND ("id" = 2 OR "id" = 3)"#
 );
 // Now: so they are now equivalent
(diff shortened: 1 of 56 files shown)


billy1624 (commented):
Hey @baoyachi, thanks for the contributions!!

`XxxQueryBuilder` is just a unit struct that serve as a construct to mark which db backend is it. Is it a performance consideration to pass by reference instead of consuming (owning) it?

I doubt there is any practical performance gain @tyt2y3 
This might be related: https://www.forrestthewoods.com/blog/should-small-rust-structs-be-passed-by-copy-or-by-borrow

billy1624:
From API ergonomic point of view: I prefer `XxxQueryBuilder` over `&XxxQueryBuilder`

baoyachi:
> Hey @baoyachi, thanks for the contributions!!
> 
> `XxxQueryBuilder` is just a unit struct that serve as a construct to mark which db backend is it. Is it a performance consideration to pass by reference instead of consuming (owning) it?
> 
> I doubt there is any practical performance gain @tyt2y3

Two part:
* not always create new XXXBuilder
* When I wanted to extend `sea-orm`, I found that it could not be extended. The key code could not be called because this is a built-in enumeration type. 

```rust
#[derive(Debug, Copy, Clone, PartialEq, Eq)]
pub enum DbBackend {
    /// A MySQL backend
    MySql,
    /// A PostgreSQL backend
    Postgres,
    /// A SQLite backend
    Sqlite,
}

```

The limit code :
```diff
 pub fn exec<'a, C>(self, db: &'a C) -> impl Future<Output = Result<XXXResult<A>, DbErr>> + '_
    where
        C: ConnectionTrait,
        A: 'a,
    {
+        let builder = db.get_database_backend();
        exec_insert(self.primary_key, builder.build(&self.query), db)
    }

```
`let builder = db.get_database_backend();` The `builder` return `DbBackend`, 

* Now it is changed to a reference structure. I can extend the trait to the interior of sea-query. The dependency is coupled

baoyachi:
@billy1624 I can modify it a little, pass both `XxxQueryBuilder ` and `ref XxxQueryBuilder `

billy1624:
> @billy1624 I can modify it slightly so that both parameters `XxxQueryBuilder ` and `ref XxxQueryBuilder ` can be accepted

That would be great! That way, it's backward compatible :)

ikrivosheev:
@baoyachi hello. Do you want to continue work in this PR?

ikrivosheev:
A month has passed without a response, I close.

PR #429, opened 2022-08-29
Update Cargo.toml

Dylan-DPC:
PR Info
Mentioning "0" would mean cargo would pick the latest dependency which is versioned "0.*" which would include versions incompatible with the one that this crate was written with and may break your create.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,8 +43,8 @@ sea-query-derive = { version = "^0.2.0", path = "sea-query-derive", optional = t
 sea-query-driver = { version = "^0.2.0", path = "sea-query-driver", optional = true }
 serde_json = { version = "^1", optional = true }
 bytes = { version = "^1", optional = true }
-chrono = { version = "^0", default-features = false, features = ["clock"], optional = true }
-postgres-types = { version = "^0", optional = true }
+chrono = { version = "0.4", default-features = false, features = ["clock"], optional = true }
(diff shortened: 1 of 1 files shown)


ikrivosheev (approved):
@Dylan-DPC hello! LGMT! Thank you)

Cargo.toml
```diff
@@ -43,8 +43,8 @@ sea-query-derive = { version = "^0.2.0", path = "sea-query-derive", optional = t
 sea-query-driver = { version = "^0.2.0", path = "sea-query-driver", optional = true }
 serde_json = { version = "^1", optional = true }
 bytes = { version = "^1", optional = true }
-chrono = { version = "^0", default-features = false, features = ["clock"], optional = true }
-postgres-types = { version = "^0", optional = true }
+chrono = { version = "0.4", default-features = false, features = ["clock"], optional = true }
+postgres-types = { version = "0.2", optional = true }
```
billy1624, line 47: Hey @Dylan-DPC, thanks for pointing this out!

I think we can relax it to be `^0.x`. That way, we're always on the latest available patch version.

Dylan-DPC: Cargo does ^ by default so this is the same thing just implicit

tyt2y3:
https://github.com/SeaQL/sea-query/commit/c9b9c66af932c6f8f34b63b17ad6242af79a1f0e

Thank you for the suggestion. I made the changes now.

PR #430, opened 2022-09-01
Create Random function

alper:
As discussed here: https://github.com/SeaQL/sea-orm/discussions/1009#discussioncomment-3523213

### DIFF
--- a/src/backend/mysql/query.rs
+++ b/src/backend/mysql/query.rs
@@ -95,6 +95,10 @@ impl QueryBuilder for MysqlQueryBuilder {
         "()"
     }
 
+    fn random_function(&self) -> &str {
+        "RAND()"
+    }
+
     fn prepare_select_distinct(
         &self,
         select_distinct: &SelectDistinct,
--- a/src/backend/query_builder.rs
+++ b/src/backend/query_builder.rs
@@ -691,6 +691,7 @@ pub trait QueryBuilder: QuotedBuilder + EscapeBuilder + TableRefBuilder {
                     Function::Upper => "UPPER",
                     Function::Custom(_) => "",
                     Function::CurrentTimestamp => "CURRENT_TIMESTAMP",
+                    Function::Random => self.random_function(),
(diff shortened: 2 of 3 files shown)


ikrivosheev (approved):
@alper hello! Thank you for contribution! LGTM!

src/backend/mysql/query.rs
```diff
@@ -95,6 +95,10 @@ impl QueryBuilder for MysqlQueryBuilder {
         "()"
     }
 
+    fn random_function(&self) -> &str {
+        "RAND()"
```
ikrivosheev, line 99: I don`t think brackets are needed here.

alper: I think they are! 

```
foursquaremaps=# select RANDOM;
ERROR:  column "random" does not exist
LINE 1: select RANDOM;
```

The parens need to be there and they're not added. The other example does work without parens:

```
foursquaremaps=# select CURRENT_TIMESTAMP;
       current_timestamp       
-------------------------------
 2022-09-01 16:38:11.467524+02
(1 row)
```

alper: Query generation does not insert the paren automatically and seems roundabout if it's a function with zero arguments.

alper: There's also `"(DEFAULT)"` where the parens are included in the output immediately.

ikrivosheev: Ok, thank you for explain!

alper:
OK. Who'll merge it?

ikrivosheev:
> OK. Who'll merge it?

@billy1624 @tyt2y3 can merge. We are waiting for them)

billy1624 (approved):
Thanks!! @alper :)

src/backend/mysql/query.rs
tyt2y3: Sadly, I think it's only because of `CURRENT_TIMESTAMP` which arguably is not a function

Issue #383, opened 2022-07-15
Deprecate / remove driver in favour of binder


Sytten:
I don't think the two should be maintained as it is confusing for the users and the binder is a better / more flexible solution long term.

tyt2y3:
Umm yes. But driver also works for other DB drivers, i.e. the vanilla `postres` and `rusqlite`. So we either have to migrate (I think `postgres` should be easy because we have been implementing `ToSql` all the time) or maintain both. We can remove SQLx support from driver as a first step though.

ikrivosheev:
@tyt2y3 hello! I have a PoC for this. I can draft PR.

tyt2y3:
> @tyt2y3 hello! I have a PoC for this. I can draft PR.

Sure

ikrivosheev:
@tyt2y3 maybe it is better to create another crate for rusqlite and Postgres?

tyt2y3:
Um... not sure will it be a lot of hassle if use feature flags. Otherwise, if we have multiple crates, isn't it still a lot similar with the situation we are in now (having multiple driver crates) ?

ikrivosheev:
@Sytten @tyt2y3 I create PR for rusqlite: https://github.com/SeaQL/sea-query/pull/388. I have a problem. I cannot create it in `sea-query-binder`, because I got error:

```
error: failed to select a version for `libsqlite3-sys`.
    ... required by package `sqlx-core v0.6.0`
    ... which satisfies dependency `sqlx-core = "^0.6.0"` of package `sqlx v0.6.0`
    ... which satisfies dependency `sqlx = "^0.6"` of package `sea-query-binder v0.1.0 (/home/ikrivosheev/projects/opensource/sea-query/sea-query-binder)`
    ... which satisfies path dependency `sea-query-binder` (locked to 0.1.0) of package `sea-query-sqlx-any-example v0.1.0 (/home/ikrivosheev/projects/opensource/sea-query/examples/sqlx_any)`
versions that meet the requirements `^0.24.1` are: 0.24.2, 0.24.1

the package `libsqlite3-sys` links to the native library `sqlite3`, but it conflicts with a previous package which links to `sqlite3` as well:
package `libsqlite3-sys v0.25.0`
    ... which satisfies dependency `libsqlite3-sys = "^0.25.0"` of package `rusqlite v0.28.0`
[...]

PR #431, opened 2022-09-03
issues-383 Remove thread-safe feature

ikrivosheev:
## PR Info

- Closes <https://github.com/SeaQL/sea-query/issues/383>

## Adds

- remove `thread-safe` feature fro `sea-query-rusqlite`


tyt2y3:
Actually, why do we need this? Downstream can enable that feature flag (on their own) without going through `binder`

ikrivosheev:
@tyt2y3 yes! You are rigth, cargo create union for features. This PR remove `thread-safe` feature from `sea-query-rusqlute`

PR #433, opened 2022-09-04
issues-383 Move postgres driver into sea-query-postgres

ikrivosheev:
## PR Info

- Closes <https://github.com/SeaQL/sea-query/issues/383>

## Adds

- Create new crate: `sea-query-postgres` with 

## Breaking Changes

- Remove driver for postgres from `sea-query`.


sea-query-postgres/src/lib.rs
```diff
@@ -55,12 +96,12 @@ impl ToSql for Value {
...
             #[cfg(feature = "with-time")]
             Value::TimeDateTimeWithTimeZone(v) => v.as_deref().to_sql(ty, out),
-            #[cfg(feature = "postgres-rust_decimal")]
+            #[cfg(feature = "with-rust_decimal")]
             Value::Decimal(v) => v.as_deref().to_sql(ty, out),
             #[cfg(feature = "with-uuid")]
             Value::Uuid(v) => v.as_deref().to_sql(ty, out),
             #[cfg(feature = "postgres-array")]
-            Value::Array(v) => v.as_deref().to_sql(ty, out),
+            Value::Array(_v) => todo!(),
```
ikrivosheev, line 104: Can someone help me with this part?

billy1624: Consider this: originally `ToSql` is implemented by `Value`. And, the `v` of `Value::Array(v)` is `Vec<Value>`. So, it satisfy the blanket implementation of `Vec<T: ToSql>` provided by `postgres`.

However, now, we only implement `ToSql` for `PostgresValue` but not `Value`. That's why we need to clone it as you did on latest commit.

tyt2y3:
Awesome, thank you!

Issue #347, opened 2022-06-06
Table Column with Default Expression


billy1624:
## Motivation
- Currently, we support column with default value but not default expression
- Having default expression in column such as timestamp / datetime is very common, e.g. `CURRENT_TIMESTAMP` as the default expression

## Proposed Solutions
- `ColumnSpec::Default(Value)` should be changed to holding `SimpleExpr`, i.e. `ColumnSpec::Default(SimpleExpr)`
  https://github.com/SeaQL/sea-query/blob/6c9527ae08e32c55490ae510be41f646e030c2a2/src/table/column.rs#L48-L58
- Keep the old `default` method and introduce a new method called `default_expr` which take any `Into<SimpleExpr>`
  https://github.com/SeaQL/sea-query/blob/6c9527ae08e32c55490ae510be41f646e030c2a2/src/table/column.rs#L142-L149

ikrivosheev:
@billy1624, hello! I will do this.

billy1624:
Thanks! @ikrivosheev

billy1624:
Related Discussion:
- https://github.com/SeaQL/sea-query/discussions/346

ikrivosheev:
@billy1624 what do you think about new `Enum` for table? We only need two variants from `SimpleExpr`

1. FuncionCall
2. Value
3. Custom

ikrivosheev:
No, bad ideas... We need to split QueryBuilder. For unavailable operations panic.

billy1624:
Hey @ikrivosheev, sorry for the delay. 

> @billy1624 what do you think about new `Enum` for table? We only need some variants from `SimpleExpr`

Do you mean introducing a new `Enum` for the "Default Value" of a column? https://github.com/SeaQL/sea-query/blob/63e18941fad55e391e4abb1316fd81e7a7768f91/src/table/column.rs#L51-L61

```
pub enum ColumnSpec {
...
    Default(DefaultValue)
...
}

// I just makeup the name, feel free to rename it :)
pub enum DefaultValue {
    Func(Function),
    Value(Value),
    Custom(String),
}
```

Something like this?

tyt2y3:
1. I think we should rename the old `default` to `default_value` and the new `default` is an alias of `default_expr`
2. Yes I think having a dedicated `DefaultColumnExpr` is a good idea

ikrivosheev:
[...]

PR #428, opened 2022-08-29
`ColumnDef` with default expression

billy1624:
## PR Info

- Closes https://github.com/SeaQL/sea-query/issues/347

## Adds

- [x] `ColumnDef::default_expr()` setting the default expression of a column
- [x] `ColumnDef::default_value()` setting the default value of a column

## Breaking Changes

- [x] `ColumnDef::default()` now takes `Into<SimpleExpr>` instead of `Into<Value>`. Users are advised to use `ColumnDef::default_value()`.

PR #436, opened 2022-09-08
issues-347 Add SqlWriter trait

ikrivosheev:
## PR Info

- Closes https://github.com/SeaQL/sea-query/issues/347

## Adds

- [x] new trait `SqlWriter`
- [x] `impl SqlWriter for String` - write sql to String with inline `Value`
- [x] `impl SqlWriter for SqlWriterValues` - write sql to String with store `Value` into `Vec`

## Breaking Changes

- [x] Remove `struct SqlWriter`
- [x] Remove `collector` from QueryBuilder and other backend traits
- [x] backend traits instead `SqlWriter` parameter get `&mut dyn SqlWriter`

## Context: https://github.com/SeaQL/sea-query/pull/428#issuecomment-1239542662

 ```rust
pub trait SqlWriter: Write + ToString {
    fn push_param(&mut self, value: Value, query_builder: &dyn QueryBuilder);

    fn as_writer(&mut self) -> &mut dyn Write;
}
```

### DIFF
--- a/src/index/drop.rs
+++ b/src/index/drop.rs
@@ -1,4 +1,4 @@
-use crate::{backend::SchemaBuilder, prepare::*, types::*, SchemaStatementBuilder, TableIndex};
+use crate::{backend::SchemaBuilder, types::*, SchemaStatementBuilder, TableIndex};
 
 /// Drop an index for an existing table
 ///
@@ -64,14 +64,14 @@ impl IndexDropStatement {
 
 impl SchemaStatementBuilder for IndexDropStatement {
     fn build<T: SchemaBuilder>(&self, schema_builder: T) -> String {
-        let mut sql = SqlStringWriter::new();
+        let mut sql = String::with_capacity(256);
         schema_builder.prepare_index_drop_statement(self, &mut sql);
-        sql.result()
+        sql
     }
 
     fn build_any(&self, schema_builder: &dyn SchemaBuilder) -> String {
-        let mut sql = SqlStringWriter::new();
+        let mut sql = String::with_capacity(256);
         schema_builder.prepare_index_drop_statement(self, &mut sql);
-        sql.result()
+        sql
     }
 }
--- a/src/index/drop.rs
+++ b/src/index/drop.rs
@@ -64,13 +64,13 @@ impl IndexDropStatement {
 
 impl SchemaStatementBuilder for IndexDropStatement {
     fn build<T: SchemaBuilder>(&self, schema_builder: T) -> String {
-        let mut sql = SqlWriter::new();
+        let mut sql = SqlStringWriter::new();
         schema_builder.prepare_index_drop_statement(self, &mut sql);
         sql.result()
     }
 
     fn build_any(&self, schema_builder: &dyn SchemaBuilder) -> String {
-        let mut sql = SqlWriter::new();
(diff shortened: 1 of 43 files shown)

### CODE REVIEW

File: src/prepare.rs [hunk 1]
```diff
@@ -3,12 +3,77 @@
 use crate::*;
 pub use std::fmt::Write;
 
-#[derive(Debug, Default)]
-pub struct SqlWriter {
-    pub(crate) counter: usize,
+pub trait SqlWriter: Write {
+    fn push_param(&mut self, value: Value, query_builder: &dyn QueryBuilder);
+    fn result(self) -> String;
+}
```
billy1624, line 9: Hey @ikrivosheev, I think we don't need the `result()` method. The `SqlWriter` trait is only responsible for writing SQL and push parameters.

The `to_string` alike method can be place in struct `SqlStringWriter` and `SqlWriterObj`.

ikrivosheev: @billy1624 hello! 

1. `trait SqlWriter: Writer + ToString`
2. remove `SqlStringWriter`, use `String` and `impl SqlWriter for String`
3. Rename `SqlWriterObj` to `SqlWriterValues`

ikrivosheev:
A Small feature grew into big refactoring!

ikrivosheev:
@billy1624 @tyt2y3 can you review this PR?

src/prepare.rs
```diff
@@ -3,10 +3,75 @@
...
+    fn write_str(&mut self, s: &str) -> std::fmt::Result {
+        write!(self.string, "{}", s)
+    }
+}
+
+impl ToString for SqlWriterValues {
+    fn to_string(&self) -> String {
+        self.string.clone()
+    }
+}
```
tyt2y3, line 58: Is this necessary? I think it'd be better if to_string is the version with parameters injected rather than just a clone of the SQL with placeholder

src/index/drop.rs
```diff
@@ -64,14 +64,14 @@ impl IndexDropStatement {
 
 impl SchemaStatementBuilder for IndexDropStatement {
     fn build<T: SchemaBuilder>(&self, schema_builder: T) -> String {
-        let mut sql = SqlWriter::new();
+        let mut sql = String::with_capacity(256);
```
tyt2y3, line 67: I feel like we want to extract this as a common function, or at least extract the default string size as a constant

tyt2y3 (approved):
Thank you for the big effort. Can you also draft the Changelog for a summary of these changes?

src/prepare.rs
ikrivosheev: My thoughts were: injection parameters are expensive. If someone needs an inject parameter into sql query it is better to call it explicit. 
Do I need to fix that?

tyt2y3: Umm I think we have two options 1) remove this impl 2) make it also inject parameters. I am okay with either.

ikrivosheev: @tyt2y3 I have problems:

1. I cannot remove this impl, because I need it, for example: https://github.com/SeaQL/sea-query/blob/master/src/query/traits.rs#L21
2. I cannot make `inject_parameters` because I need: `query_builder` but I didnot store it in SqlWriter...

Issue #440, opened 2022-09-14
Remove `CURRENT_TIMESTAMP` from function


tyt2y3:
We should remove `CURRENT_TIMESTAMP` from function and put it under SystemVariable (or SysVar) instead.

_Originally posted by @tyt2y3 in https://github.com/SeaQL/sea-query/issues/349#issuecomment-1246374594_

billy1624:
Yeah, that make more sense. Basically, `CURRENT_TIMESTAMP` is a system variable (or constant) and `CURRENT_TIMESTAMP()` is a function.

billy1624:
Context: https://github.com/SeaQL/sea-query/commit/70c25784730b7d220e6878804badf1c72e255d43#commitcomment-83866265

> Hey @tyt2y3, I don't think this is correct for SQLite and PostgreSQL.
> 
> Both of them expect `CURRENT_TIMESTAMP` instead of `CURRENT_TIMESTAMP()`.
> 
> * https://www.sqlite.org/lang_keywords.html
> * https://www.postgresql.org/docs/current/functions-datetime.html#FUNCTIONS-DATETIME-CURRENT
> 
> You can write a select query `SELECT CURRENT_TIMESTAMP;` [here](https://dbfiddle.uk/) to test it.

ikrivosheev:
@tyt2y3 @billy1624 i create PR which fix this

PR #441, opened 2022-09-14
issues-440 Add new keywords for work

ikrivosheev:
## PR Info

- Closes <https://github.com/SeaQL/sea-query/issues/440>

## Adds

- Add new Keywords: `CURRENT_DATE`, `CURRENT_TIME`, `CURRENT_TIMESTAMP` (all this keywords support by Mysql, Sqlite, Postgresql)

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -200,6 +200,9 @@ pub struct NullAlias;
 #[derive(Debug, Clone)]
 pub enum Keyword {
     Null,
+    CurrentDate,
+    CurrentTime,
+    CurrentTimestamp,
     Custom(DynIden),
 }
 
--- a/src/backend/query_builder.rs
+++ b/src/backend/query_builder.rs
@@ -688,7 +688,6 @@ pub trait QueryBuilder: QuotedBuilder + EscapeBuilder + TableRefBuilder {
                     Function::Lower => "LOWER",
                     Function::Upper => "UPPER",
                     Function::Custom(_) => "",
-                    Function::CurrentTimestamp => "CURRENT_TIMESTAMP",
                     Function::Random => self.random_function(),
                     #[cfg(feature = "backend-postgres")]
                     Function::PgFunction(_) => unimplemented!(),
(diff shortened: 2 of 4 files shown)


src/types.rs
```diff
@@ -200,6 +200,9 @@ pub struct NullAlias;
 #[derive(Debug, Clone)]
 pub enum Keyword {
     Null,
+    CurrentDate,
+    CurrentTime,
+    CurrentTimestamp,
     Custom(DynIden),
```
billy1624, line 206: Hey @ikrivosheev, thanks for the PR!! I think we can safely remove `Keyword::Custom`? Since no one is using it and it essentially did nothing.

billy1624 (commented):
Hey @ikrivosheev, thanks for the prompt PR! Yeah, put it under `Keyword` enum also works.

Next, we need to provide a public API for user. Would adding `current_time`, `current_date` and `current_timestamp` methods under `Expr` be a good idea? The method would return a `SimpleExpr`.

ikrivosheev:
> Hey @ikrivosheev, thanks for the prompt PR! Yeah, put it under `Keyword` enum also works.
> 
> Next, we need to provide a public API for user. Would adding `current_time`, `current_date` and `current_timestamp` methods under `Expr` be a good idea? The method would return a `SimpleExpr`.

@billy1624 hello! Done. I add helpers methods.

tyt2y3:
Umm strictly speak they are now Keywords. I guess it's kind of an implementation detail, but better to have another struct for these names.

ikrivosheev:
> Umm strictly speak they are not Keywords. I guess it's kind of an implementation detail, but better to have another struct for these names.

I don't quite agree.  SQLite doc (https://www.sqlite.org/lang_keywords.html) says that is a keyword. 
In MySQL this is a function but we can use it without `()`. 
 In Postgres this is a function but without `()`.

ikrivosheev:
@tyt2y3 is current realization ok or need improvements?

PR #443, opened 2022-09-15
fix: provide error hint for unsupported column type

EstebanBorai:
## Changes

When using `auto_increment` with random column types the current error message doesn't describe the actual reason when running migrations:

```
`migration up`
Applying 1 pending migrations
Applying migration 'm20220915_153744_create_links_table'
thread 'main' panicked at 'not implemented', sea-query-0.26.3/src/backend/postgres/table.rs:239:26
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### DIFF
--- a/src/backend/postgres/table.rs
+++ b/src/backend/postgres/table.rs
@@ -236,7 +236,7 @@ impl PostgresQueryBuilder {
                     ColumnType::SmallInteger(_) => write!(sql, "smallserial").unwrap(),
                     ColumnType::Integer(_) => write!(sql, "serial").unwrap(),
                     ColumnType::BigInteger(_) => write!(sql, "bigserial").unwrap(),
-                    _ => unimplemented!(),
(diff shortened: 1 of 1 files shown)


billy1624 (commented):
Hey @EstebanBorai, thanks for reporting this and proposing a change. That's definitely a heading into the right direction!

I think we should check the code base and avoid panicking without a explicit error message. @ikrivosheev @tyt2y3?

ikrivosheev:
@billy1624 hello! I agree with you! I think is better to remove all panic from sea-query, but at the first stage we can replace `unwrap` to `except`

ikrivosheev (approved):

billy1624:
Yeah, replace all `unwrap` with `expect` would be the first step.

Similar to this issue :)
- https://github.com/SeaQL/sea-orm/issues/1023

ikrivosheev:
@billy1624 I think we can merge this PR and then create new issue and continue work in other PR.

Issue #458, opened 2022-09-28
Postgres bytea converted to bit


RichardOverzee:
## Description

While using sea_query version 0.26.3. When inserting a VecU8 into a postgres BYTEA column it is actually converted into a bit which cause the actual insert statement for a BYTEA to be prefixed by x'. Which is not what we want. 
Currently the work around is to parse it as a string so it can be inserted.

## Steps to Reproduce

1. 
Install  sea_query version 0.26.3
2. 
Create a postgres table with a BYTEA column type
3. 
Try insert a VecU8 such as
`let sql = Query:insert()
        .into_table((Alias::new(mySchema), MyEnumTable::Table))
        .columns([
            MyEnumTable::MyBYTEA,
        ])
        .values(vec![
            myVecU8.into(),
        ])?
        .to_string(PostgresQueryBuilder);
`

### Expected Behavior

I was expecting to insert a BYTEA into postgres and not a Bit

### Actual Behavior

Sea_query converted my bytes into a bit by prefixing the query with a 'x'

### Reproduces How Often

Always

## Versions

│   │   │   ├── sea-query v0.26.3
│   │   │   │   ├── sea-query-derive v0.2.0 (proc-macro)
│   │   │   │   ├── sea-query-driver v0.2.2 (proc-macro)
│   │   ├── sea-query v0.26.3 (*)
│   ├── sea-query v0.26.3 (*)
│   ├── sea-query v0.26.3 (*)
│   ├── sea-query v0.26.3 (*)

billy1624:
Hey @RichardOverzee, thanks for catching this! I'll draft a PR to fix this in a bit.

RichardOverzee:
@tyt2y3 the behavior is well explained by billy1624 here: https://github.com/SeaQL/sea-query/pull/459#issuecomment-1271251488
Is it a bug per say? I would say so. 
Postgres Bit type only accept 0's and 1's.
Postgres Bytea type support Hex values and the 'escape' format from postgres.
Rust Vec<u8> is more than juste 0's and 1's.
Why should a Vec<u8> (which can have other values than 0 or 1) be inserted into a postgres Bit type which only accept 0 or 1?.

tyt2y3:
Then what's the correct syntax to encode a byte array indeed?

ikrivosheev:
@tyt2y3 @billy1624 @RichardOverzee any updates? Or I can close issue?

tyt2y3:
[...]

PR #459, opened 2022-09-30
`Value::Bytes` should write as byte string instead of bit string

billy1624:
## PR Info

- Closes https://github.com/SeaQL/sea-query/issues/458

## Fixes

- [x] `QueryBuilder::value_to_string` method write byte as byte string instead of bit string

### DIFF
--- a/src/backend/query_builder.rs
+++ b/src/backend/query_builder.rs
@@ -983,7 +983,7 @@ pub trait QueryBuilder: QuotedBuilder + EscapeBuilder + TableRefBuilder {
             }
             Value::Bytes(Some(v)) => write!(
                 s,
-                "x'{}'",
+                "'{}'",
                 v.iter().map(|b| format!("{:02X}", b)).collect::<String>()
             )
             .unwrap(),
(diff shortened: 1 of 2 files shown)


billy1624:
Hey @tyt2y3, is there any reason we need to add a bit string prefix `x` to `x'ABCDEF'`? Is this db dependent? E.g. MySQL requires the `x` prefix but PostgreSQL doesn't?

ikrivosheev:
@billy1624 maybe is better move `x` and `b` prefixes to another place? For example: add new `enum Prefix` and add support for it into `Expr`?

billy1624:
But for Postgres, `x'FA'` denote it as bit string instead of byte string.

```sql
insert into tbl values (x'0123456789ABCDEF')
```

> ERROR:  column "col" is of type bytea but expression is of type bit
> LINE 1: insert into tbl values (x'0123456789ABCDEF')
>                                 ^
> HINT:  You will need to rewrite or cast the expression.

Demo: https://dbfiddle.uk/T9Cx6dZK

tyt2y3:
I think the correct syntax is '\xABCDEF'

tyt2y3:
The proper fix should be:

1. Move this encoding as a backend specific function
2. MySQL should remain using `x'ABCD'`, while for Postgres, we do `'\xABCD'`
3. Not sure about SQLite

PR #467, opened 2022-10-09
issues-336 Third try impl postgres-array

ikrivosheev:
## PR Info

Other PR:
- https://github.com/SeaQL/sea-query/pull/390
- https://github.com/SeaQL/sea-query/pull/413

This is the third try to impl postgres-array in SeaQuery. Here I store type information into `Value` and then try to convert `Vec<Value>` into `Vec<T>`. This look safe. 

## Adds

- Support one dimension Postgres Array into `sea-query-binder`
- `enum ArrayType` with all possible types 
- Add `Value::except` method

## Breaking Changes

- `ColumnType::Array` store `Box<ColumnType>` instead string
- `Value::Array` store addition field: `ArrayType`, where store type of array elements

## Changes

- Small improvement `Iden::to_string`

### DIFF
--- a/sea-query-binder/Cargo.toml
+++ b/sea-query-binder/Cargo.toml
@@ -17,22 +17,33 @@ rust-version = "1.60"
 [lib]
 
 [dependencies]
-sea-query = { version = "^0", path = ".." }
+sea-query = { version = "^0", path = "..", features = ["thread-safe"] }
 sqlx = { version = "^0.6", optional = true }
+serde_json = { version = "^1", optional = true }
+chrono = { version = "^0.4", default-features = false, features = ["clock"], optional = true }
+postgres-types = { version = "^0", optional = true }
+rust_decimal = { version = "^1", optional = true }
+bigdecimal = { version = "^0.3", optional = true }
+uuid = { version = "^1", optional = true }
+proc-macro2 = { version = "1", optional = true }
+quote = { version = "^1", optional = true }
+time = { version = "^0.3", optional = true, features = ["macros", "formatting"] }
+ipnetwork = { version = "^0.19", optional = true }
+mac_address = { version = "^1.1", optional = true }
 
 [features]
 sqlx-mysql = ["sqlx/mysql"]
 sqlx-postgres = ["sqlx/postgres"]
 sqlx-sqlite = ["sqlx/sqlite"]
 sqlx-any = ["sqlx/any"]
-with-chrono = ["sqlx/chrono", "sea-query/with-chrono"]
-with-json = ["sqlx/json", "sea-query/with-json"]
-with-rust_decimal = ["sqlx/decimal", "sea-query/with-rust_decimal"]
-with-bigdecimal = ["sqlx/bigdecimal", "sea-query/with-bigdecimal"]
-with-uuid = ["sqlx/uuid", "sea-query/with-uuid"]
-with-time = ["sqlx/time", "sea-query/with-time"]
-with-ipnetwork = ["sqlx/ipnetwork", "sea-query/with-ipnetwork"]
-with-mac_address = ["sqlx/mac_address", "sea-query/with-mac_address"]
+with-chrono = ["sqlx/chrono", "sea-query/with-chrono", "chrono"]
+with-json = ["sqlx/json", "sea-query/with-json", "serde_json"]
+with-rust_decimal = ["sqlx/decimal", "sea-query/with-rust_decimal", "rust_decimal"]
+with-bigdecimal = ["sqlx/bigdecimal", "sea-query/with-bigdecimal", "bigdecimal"]
+with-uuid = ["sqlx/uuid", "sea-query/with-uuid", "uuid"]
+with-time = ["sqlx/time", "sea-query/with-time", "time"]
+with-ipnetwork = ["sqlx/ipnetwork", "sea-query/with-ipnetwork", "ipnetwork"]
+with-mac_address = ["sqlx/mac_address", "sea-query/with-mac_address", "mac_address"]
 postgres-array = ["sea-query/postgres-array"]
 runtime-async-std-native-tls = ["sqlx/runtime-async-std-native-tls"]
 runtime-async-std-rustls = ["sqlx/runtime-async-std-rustls", ]
--- a/sea-query-binder/Cargo.toml
+++ b/sea-query-binder/Cargo.toml
@@ -18,7 +18,7 @@ rust-version = "1.60"
 
 [dependencies]
 sea-query = { version = "^0", path = "..", features = ["thread-safe"] }
-sqlx = { version = "^0.6", optional = true }
+sqlx = { version = "^0.6.1", optional = true }
 serde_json = { version = "^1", optional = true }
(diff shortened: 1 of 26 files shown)


ikrivosheev:
@tyt2y3 @billy1624 can you look next my idea?)

Sytten:
For me any solution that cant be extended to support custom types is a no go. That is why the trait based solution is used by all other libraries.

ikrivosheev:
> For me any solution that cant be extended to support custom types is a no go. That is why the trait based solution is used by all other libraries.

I try to implement trait solution, but got a lot of problems. I still don't understand how to solve them all...

sea-query-postgres/Cargo.toml
```diff
@@ -17,7 +17,7 @@ rust-version = "1.60"
 [lib]
 
 [dependencies]
-sea-query = { version = "^0", path = ".." }
+sea-query = { version = "^0", path = "..", feature = ["thread-safe"] }
```
tyt2y3, line 20: I guess this is not strictly necessary

sea-query-binder/Cargo.toml
```diff
@@ -17,7 +17,7 @@ rust-version = "1.60"
 [lib]
 
 [dependencies]
-sea-query = { version = "^0", path = ".." }
+sea-query = { version = "^0", path = "..", features = ["thread-safe"] }
```
tyt2y3, line 20: This makes sense

tyt2y3:
This looks good, especially as an intermediate step. The crate architecture has been revamped in 0.27 that is worthy for a release. What's important now is to roll out the next iteration of SeaORM. And maybe I can make a try on `ValueTrait` sometime after that.

ikrivosheev:
@tyt2y3 @billy1624 @Sytten hello! I have done with some part of this PR.

1. Create new enum: `ArrayType`
2. Add support for `Value::Array` for sqlx/postgres
3. Extend `trait ValueType`: `fn array_type() -> ArrayType` and `fn except(v: Value, msg: &str) -> Self`

**TODO:**

1. Add support for `Vec<Option<T>>` - don't know how to do it yet

sea-query-postgres/Cargo.toml
ikrivosheev: Remove.

tyt2y3:
Great work! I think now I have an idea as to how ValueTrait would work with SQLx, I am willing to give a try after this settled and we released 0.27

tyt2y3:
This seems to be very nearly done?

tyt2y3:
@ikrivosheev check this https://github.com/SeaQL/sea-query/commit/79625e0ebe3abcfc5b84f24d2d4da2682425d67e not sure if it really works at downstream, but it compiles

ikrivosheev:
> > This seems to be very nearly done?
> 
> I want to add support for `Vec<Option<T>>`

Or later...

tyt2y3:
Cool enough. @billy1624 thoughts?

src/value.rs
```diff
@@ -135,8 +218,14 @@ pub trait ValueType: Sized {
         Self::try_from(v).unwrap()
     }
 
+    fn except(v: Value, msg: &str) -> Self {
+        Self::try_from(v).expect(msg)
+    }
```
billy1624, line 223: Would it be better to rename it as `expect`? Since it's a wrapper method of `try_from` then `expect`.

This also align with the API of... few lines above it
```rs
fn unwrap(v: Value) -> Self {
    Self::try_from(v).unwrap()
}
```

src/value.rs
```diff
@@ -185,10 +274,17 @@ impl Value {
     {
         T::unwrap(self)
     }
+
+    pub fn except<T>(self, msg: &str) -> T
+    where
+        T: ValueType,
+    {
+        T::except(self, msg)
+    }
```
billy1624, line 283: Same here :)

billy1624 (commented):
Wow!! That's smart!

sea-query-binder/src/sqlx_postgres.rs
```diff
@@ -117,6 +129,188 @@ impl<'q> sqlx::IntoArguments<'q, sqlx::postgres::Postgres> for SqlxValues {
                 Value::MacAddress(mac) => {
                     args.add(mac.as_deref());
                 }
+                #[cfg(feature = "postgres-array")]
+                Value::Array(ty, v) => match ty {
+                    ArrayType::Bool => {
+                        let value: Option<Vec<bool>> =
+                            Value::Array(ty, v).except("Invalid type for array value");
+                        args.add(value)
+                    }
```
billy1624, line 138: I think ordinary user won't see this message (unless someone intentionally construct a vector of boxed value with different value variants in it) but the error message itself can be more specific.

```rs
Value::Array(ty, v).except("This Value::Array should consist of Value::Bool");
```

src/value.rs
```diff
@@ -222,7 +323,7 @@ macro_rules! type_to_value {
 }
 
 macro_rules! type_to_box_value {
-    ( $type: ty, $name: ident, $col_type: expr ) => {
+    ( $type: ty, $name: ident, $col_type: expr, $array_type: expr ) => {
```
billy1624, line 326: Nitpick: a trivial refactor
- https://github.com/ikrivosheev/sea-query/pull/1

billy1624:
> Cool enough. @billy1624 thoughts?

Looks good to me other than above comments :D

src/value.rs
ikrivosheev: Thank you!

src/value.rs
ikrivosheev: Fix!

src/value.rs
ikrivosheev: Fix!

sea-query-binder/src/sqlx_postgres.rs
ikrivosheev: Done

ikrivosheev:
@billy1624 thank you for review! I corrected all comments

sea-query-binder/Cargo.toml
```diff
@@ -17,22 +17,33 @@ rust-version = "1.60"
...
 sqlx = { version = "^0.6.1", optional = true }
+serde_json = { version = "^1", optional = true }
+chrono = { version = "^0.4", default-features = false, features = ["clock"], optional = true }
```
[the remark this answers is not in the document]
ikrivosheev: We want to make our third-party libraries as part of public interface?

src/value.rs
```diff
@@ -185,6 +274,13 @@ impl Value {
     {
         T::unwrap(self)
     }
+
+    pub fn except<T>(self, msg: &str) -> T
```
billy1624, line 278: Are we keeping this as `except`?

ikrivosheev: Thank you! Fix

sea-query-binder/Cargo.toml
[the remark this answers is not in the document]
tyt2y3: Good point

tyt2y3:
One last thing, can we also update the sqlx-postgres example (with several Array types) and make it part of our CI?

PR #468, opened 2022-10-09
Change datetime string format

ikrivosheev:
## PR Info

- Closes https://github.com/SeaQL/sea-orm/issues/1096

## Breaking Changes

- Change datetime string format for time crate.


ikrivosheev:
The problem is that at the same time use `time` and `chrono` crate. I will rewrite example.

ikrivosheev:
@billy1624 @tyt2y3 well, now it is ready to review.

tyt2y3:
Thank you

PR #477, opened 2022-10-16
ValueTrait for SQLx

tyt2y3:
This is an initial design of `ValueTrait` to allow custom `Value` to be used with SeaQuery.

The tricky part is the dependency tree:

```
sea-query-binder v0.1.0 (/home/chris/sea-query/sea-query-binder)
├── sea-query v0.27.0 (/home/chris/sea-query)
│   ├── sea-query-derive v0.2.0 (proc-macro) (/home/chris/sea-query/sea-query-derive)
│   └── sea-query-sqlx v0.1.0 (/home/chris/sea-query/sea-query-sqlx)
│       └── sqlx v0.6.2
│           ├── sqlx-core v0.6.2
│           │   ├── sqlx-rt v0.6.2
│           └── sqlx-macros v0.6.2 (proc-macro)
│               ├── sqlx-core v0.6.2
│               │   ├── sqlx-rt v0.6.2 (*)
│               ├── sqlx-rt v0.6.2 (*)
├── sea-query-sqlx v0.1.0 (/home/chris/sea-query/sea-query-sqlx) (*)
└── sqlx v0.6.2 (*)
```

Now sadly we can no longer `cargo b --all-features`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ name = "sea_query"
 path = "src/lib.rs"
 
 [dependencies]
+sea-query-sqlx = { version = "^0.1", path = "sea-query-sqlx", optional = true }
 sea-query-attr = { version = "^0.1.1", path = "sea-query-attr", optional = true }
 sea-query-derive = { version = "^0.2.0", path = "sea-query-derive", optional = true }
 serde_json = { version = "^1", optional = true }
--- a/sea-query-binder/Cargo.toml
+++ b/sea-query-binder/Cargo.toml
@@ -1,5 +1,5 @@
 [workspace]
-members = [".", "sea-query-sqlx"]
+members = ["."]
 
 [package]
 name = "sea-query-binder"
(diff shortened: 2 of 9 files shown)


tyt2y3:
Would appreciate some thoughts @ikrivosheev @billy1624

billy1624 (commented):
Oh no, sorry for the delay. My GitHub notification box is flooded.

I think we still need:
- `impl SqlxValueTrait for xxx`, e.g. `i32`, `Vec<i32>`...etc.
- `impl<T: SqlxValueTrait> From<T> for Value`: I doubt this will cause compile error.

For i32, we have following two `impl From<i32> for Value`:
- `impl From<i32> for Value`
- `impl SqlxValueTrait for i32`
  - Hence, `impl<T: SqlxValueTrait> From<T> for Value` valid for i32 as well.

Will Rust compiler compliant?

tyt2y3:
My plan is, we keep using our current implementation for all supported types, i32 etc.
So we don't need to blanket impl anything.

billy1624:
Okay, so the burden is on the user who need to convert the types into `SeaRc<dyn SqlxValueTrait>`

tyt2y3:
> Okay, so the burden is on the user who need to convert the types into `SeaRc<dyn SqlxValueTrait>`

Yes. A hard-way is better than no-way.

tyt2y3:
given up on this idea

PR #486, opened 2022-10-18
feat: added trgm operators

baszalmstra:
## PR Info

Adds [`pg_trgm`](https://github.com/SeaQL/sea-query/discussions/483) binary operators which are required to allow proper use of indexes for Postgres.

## Adds

- Binary operators from the Postgres `pg_trgm` extension

### DIFF
--- a/src/backend/postgres/query.rs
+++ b/src/backend/postgres/query.rs
@@ -41,6 +41,12 @@ impl QueryBuilder for PostgresQueryBuilder {
             BinOper::Contains => write!(sql, "@>").unwrap(),
             BinOper::Contained => write!(sql, "<@").unwrap(),
             BinOper::Concatenate => write!(sql, "||").unwrap(),
+            BinOper::Similarity => write!(sql, "%").unwrap(),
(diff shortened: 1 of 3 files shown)


ikrivosheev:
@baszalmstra thank you for PR! Can you run `rustfmt`?

ikrivosheev (approved):
@baszalmstra thank you! LGTM! 

I think we can also add: `BinOpen::Custom` for this case

baszalmstra:
> I think we can also add: `BinOpen::Custom` for this case

I tried adding `BinOper::Custom(String)` but the `BinOper` is `Copy` and `String` obviously isn't. Removing the `Copy` derive would change the API significantly.

I see a few options:

* Use `&'static str`, but that would limit the use
* Remove the `Copy` trait from `BinOp` 
* Implement the `Copy` trait and use `Rc`, `Arc`, or `DynIden`

WDYT?

tyt2y3 (approved):

tyt2y3:
May be we can cater custom BinOp later.

PR #493, opened 2022-10-31
add support for SQLite strict tables

evgeniy-terekhin:
this PR adds support for [sqlite strict tables](https://www.sqlite.org/stricttables.html)
it does so by adding a new variant to `TableOpt` variant

example:
`create table test(a, b)` (regular create statement)
`create table test(a int, b int) strict` (strict create statement)

## Questions

Is `TableOpt` the right place to put this functionality in?
Does `TableOpt` have to be public given that there are builder methods for each individual option?

### DIFF
--- a/src/backend/table_builder.rs
+++ b/src/backend/table_builder.rs
@@ -101,6 +101,7 @@ pub trait TableBuilder: IndexBuilder + ForeignKeyBuilder + QuotedBuilder + Table
                 TableOpt::Engine(s) => format!("ENGINE={}", s),
                 TableOpt::Collate(s) => format!("COLLATE={}", s),
                 TableOpt::CharacterSet(s) => format!("DEFAULT CHARSET={}", s),
+                TableOpt::Strict => "STRICT".to_string(),
             }
         )
         .unwrap()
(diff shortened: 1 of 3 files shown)


evgeniy-terekhin:
it might make sense to rewrite this PR in terms of `TableModifier` introduced in [this PR](https://github.com/SeaQL/sea-query/pull/494)

src/backend/table_builder.rs
```diff
@@ -101,6 +101,7 @@ pub trait TableBuilder: IndexBuilder + ForeignKeyBuilder + QuotedBuilder + Table
                 TableOpt::Engine(s) => format!("ENGINE={}", s),
                 TableOpt::Collate(s) => format!("COLLATE={}", s),
                 TableOpt::CharacterSet(s) => format!("DEFAULT CHARSET={}", s),
+                TableOpt::Strict => "STRICT".to_string(),
```
billy1624, line 104: Hey @evgeniy-terekhin, sorry for the delay.

Since `STRICT` is a SQLite only keyword, we better append it on SQLite's table_builder instead of doing it in the common table_builder.

ikrivosheev:
@evgeniy-terekhin hello! Sorry with the delay. I think is better split TableOpt into backend specific options, what do you think?

PR #494, opened 2022-11-01
support SQLite virtual tables (introduce TableModifier)

evgeniy-terekhin:
This PR introduces support for SQLite VIRTUAL tables

it does so by introducing a new entity for `CreateTableStatement` - `TableModifier`

The idea of `TableModifier` is that it modifies the very `CREATE TABLE` part of the statement (in this case it is `CREATE VIRTUAL TABLE`). 
At the moment `TableModifier` is a struct with just one field `virtual_table`. The idea is that it will contain as many fields as there are options to modify the table. Fpostgres for instance we can have several modifiers at once (e.g `CREATE GLOBAL TEMPORARY TABLE`).

### DIFF
--- a/src/backend/table_builder.rs
+++ b/src/backend/table_builder.rs
@@ -7,7 +7,17 @@ pub trait TableBuilder: IndexBuilder + ForeignKeyBuilder + QuotedBuilder + Table
         create: &TableCreateStatement,
         sql: &mut dyn SqlWriter,
     ) {
-        write!(sql, "CREATE TABLE ").unwrap();
+        if let Some(modifier) = create.modifier.as_ref() {
+            write!(sql, "CREATE ").unwrap();
+
+            if modifier.virtual_table.is_some() {
+                write!(sql, "VIRTUAL ").unwrap();
+            }
+
+            write!(sql, "TABLE ").unwrap();
+        } else {
+            write!(sql, "CREATE TABLE ").unwrap();
+        }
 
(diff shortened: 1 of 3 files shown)


ikrivosheev (commented):
@evgeniy-terekhin hello! Thank you for the PR! Can you run rustfmt?)

evgeniy-terekhin:
> @evgeniy-terekhin what is difference between `modifier` and `options`?

this is a very good question
In principle they should (and most likely must) serve the same purpose

hovever the way table options are implemented at the moment it is impossible to add `VIRTUAL` table support as another variant of `TableOpt` enum

A viable idea can be to extend `TableModifier` with fields corrsponding to `TableOpt` variants

```
struct TableModifier {
   virtual_table: Option<String>,
   character_set: Option<String>
   ...
}
```

and instead of iterating over table options in  the very end of `prepare_table_create_statement` we could conditionally render depending on contents of individual fields of the modifier struct.

billy1624:
I'm not familiar with virtual table in SQLite. So, please correct me if I'm wrong.

Seems like SQLite virtual table is not the same as temporary table in MySQL and Postgres. In other words, they are two different things and virtual table is a special feature offered by SQLite only.

evgeniy-terekhin:
@billy1624 yes it is an SQLite-specific feature

However I think that it would be really hard to implement this properly and probably the effort wouldn't worth it

So I am closing this PR

Issue #496, opened 2022-10-15
Need value 'FullOuterJoin' in 'JoinType' enum


itsAmitGoyani:
## Motivation

I was working with Join query and found that 'Join' and 'InnerJoin' works exactly same. Apart from that I want to use 'FULL JOIN' and there is no such enum value in 'JoinType' enum.

## Proposed Solutions

Add one new value named 'FullOuterJoin' in 'JoinType' enum and also make required changes in query builder.

ikrivosheev:
@itsAmitGoyani hello! Thank you for the issue. I created PR.

billy1624:
Thanks!! @ikrivosheev

PR #497, opened 2022-11-02
issues-496 Add support FULL OUTER JOIN

ikrivosheev:
## PR Info

- Closes https://github.com/SeaQL/sea-query/issues/496

## Changes

- Add support `FULL OUTER JOIN`

### DIFF
--- a/src/backend/mysql/query.rs
+++ b/src/backend/mysql/query.rs
@@ -30,6 +30,22 @@ impl QueryBuilder for MysqlQueryBuilder {
         // MySQL doesn't support declaring materialization in SQL for with query.
     }
 
+    fn prepare_join_type(&self, join_type: &JoinType, sql: &mut dyn SqlWriter) {
+        write!(
+            sql,
+            "{}",
+            match join_type {
+                JoinType::Join => "JOIN",
+                JoinType::CrossJoin => "CROSS JOIN",
+                JoinType::InnerJoin => "INNER JOIN",
+                JoinType::LeftJoin => "LEFT JOIN",
+                JoinType::RightJoin => "RIGHT JOIN",
+                JoinType::FullOuterJoin => panic!("Mysql does not support FULL OUTER JOIN"),
+            }
+        )
+        .unwrap()
+    }
+
     fn prepare_order_expr(&self, order_expr: &OrderExpr, sql: &mut dyn SqlWriter) {
         match order_expr.nulls {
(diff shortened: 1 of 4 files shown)


billy1624:
Please correct me if I am wrong:

MySQL doesn't support `FULL OUTER JOIN`, only Postgres and SQLite support it. We can still support `FULL OUTER JOIN` in SeaQuery but we should throw error if it's a MySQL query.

CC @itsAmitGoyani

ikrivosheev:
> Please correct me if I am wrong:
> 
> MySQL doesn't support `FULL OUTER JOIN`, only Postgres and SQLite support it. We can still support `FULL OUTER JOIN` in SeaQuery but we should throw error if it's a MySQL query.
> 
> CC @itsAmitGoyani

@billy1624, yes! You are right...

itsAmitGoyani:
@billy1624 Correct.

ikrivosheev:
@billy1624 @itsAmitGoyani Well, which behavior are we expect? `panic` in runtime? Or we try to emulate the join type?

ikrivosheev:
@billy1624 Done!

src/backend/mysql/query.rs
```diff
@@ -30,6 +30,22 @@ impl QueryBuilder for MysqlQueryBuilder {
...
+            sql,
+            "{}",
+            match join_type {
+                JoinType::Join => "JOIN",
+                JoinType::CrossJoin => "CROSS JOIN",
+                JoinType::InnerJoin => "INNER JOIN",
+                JoinType::LeftJoin => "LEFT JOIN",
+                JoinType::RightJoin => "RIGHT JOIN",
+                JoinType::FullOuterJoin => panic!("Mysql does not support FULL OUTER JOIN"),
+            }
```
billy1624, line 44: Personally, I think duplicating all other `JoinType` once more hurt maintainability. Maybe we can create a `prepare_join_type_common` method just like below? What's your take? @ikrivosheev 

https://github.com/SeaQL/sea-query/blob/cfe6bd2ef90fa7483d68364e8563c172de244947/src/backend/sqlite/query.rs#L23-L37

ikrivosheev: @billy1624 yes, you are right! Thank you, fixed.

billy1624 (approved):
Hey @ikrivosheev, thanks for the refactoring!!

PR #498, opened 2022-11-02
Wrap unions into parenthesis (#495)

### LINKED ISSUE #495: union does not add parentheses to the individual selects
## Steps to Reproduce

1. Create two selectstatements on the same table (cake)
2. First one should be a simple select statement
3. Second should have both a join (cake_filling) and an order by on a column from the joined table
4. Union them

### Expected Behavior

`SELECT cake.column FROM cake (UNION SELECT cake.column FROM cake INNER JOIN cake_filling ON cake.id = cake_filling.cake_id ORDER BY cake_filling.filling_id)`

### Actual Behavior

`SELECT cake.column FROM cake UNION SELECT cake.column FROM cake INNER JOIN cake_filling ON cake.id = cake_filling.cake_id ORDER BY cake_filling.filling_id` which will error:

Error 1250(42000): Table `cake_filling` from one of the SELECTs cannot be used in field list.

### Reproduces How Often

Always on Mysql (That I Tested), but I think it's a problem on all database flavours.

## Versions

0.27.1

## Additional Information

I triggered it in personal code so I don't have an example, could make one if needed.

adrian-budau:
## PR Info

- Closes #495

## Bug Fixes

- Union selects had their `ORDER BY` applied to the whole statement, which was counterintuitive to the declarative syntax.
The first select order by will still apply to the whole union to preserve this useful capability.

### DIFF
--- a/src/backend/query_builder.rs
+++ b/src/backend/query_builder.rs
@@ -122,12 +122,13 @@ pub trait QueryBuilder: QuotedBuilder + EscapeBuilder + TableRefBuilder {
         if !select.unions.is_empty() {
             select.unions.iter().for_each(|(union_type, query)| {
                 match union_type {
-                    UnionType::Intersect => write!(sql, " INTERSECT ").unwrap(),
-                    UnionType::Distinct => write!(sql, " UNION ").unwrap(),
-                    UnionType::Except => write!(sql, " EXCEPT ").unwrap(),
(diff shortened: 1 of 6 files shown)


ikrivosheev (approved):
@adrian-budau thank you! LGTM!

tyt2y3 (approved):

PR #502, opened 2022-11-04
Remove unnecessary panicking catchall from sea-query-postgres

### LINKED ISSUE #501: Specifying feature for wrong crate leads to a runtime error
Run `cargo init` and edit `Cargo.toml` as follows:

```toml
[package]
name = "demo"
version = "0.1.0"
edition = "2021"

[dependencies]
chrono = { version = "0.4.22", default-features = false }
postgres-types = "0.2.4"
sea-query = { version = "0.27.1", features = ["with-chrono"] }
sea-query-postgres = { version = "0.1.1" }
```

edit `main.rs` as follows:

```rs
use chrono::Utc;
use postgres_types::{private::BytesMut, ToSql, Type};
use sea_query::Value;
use sea_query_postgres::PostgresValue;

fn main() {
    let val: Value = Utc::now().naive_utc().into();
    let mut out = BytesMut::new();
    PostgresValue(val).to_sql(&Type::TIMESTAMP, &mut out);
}
```

Now attempt `cargo run` and get a runtime panic: `thread 'main' panicked at 'not implemented'`.

https://github.com/SeaQL/sea-query/blob/f9b8f6ad31b804f3545565e495b063c531a6a24b/sea-query-postgres/src/lib.rs#L111-L112

as it turns out the above `unimplemented!()` is very much reachable.

This of course does not happen when you specify the `"with-chrono"` feature for `sea-query-postgres` like you're supposed to ... however that is a very easy mistake to make (I just made it). It would be nice if messing the feature flags up like this resulted in a compile time error instead ... because if it compiles it should also run.

not-my-profile:
Fixes #501.

### DIFF
--- a/sea-query-postgres/src/lib.rs
+++ b/sea-query-postgres/src/lib.rs
@@ -108,8 +108,6 @@ impl ToSql for PostgresValue {
                 .to_sql(ty, out),
             #[cfg(feature = "postgres-array")]
             Value::Array(_, None) => Ok(IsNull::Yes),
-            #[allow(unreachable_patterns)]
-            _ => unimplemented!(),
         }
     }
 


ikrivosheev (approved):
@not-my-profile thank you! LGTM

ikrivosheev:
@not-my-profile build failed. Can you fix it?

ikrivosheev:
https://github.com/SeaQL/sea-query/pull/503

PR #504, opened 2022-11-05
add support for sqlite MATCH operator

evgeniy-terekhin:
this PR adds support for sqlite MATCH operator used primarily for full text search purposes
It does so by allowing `Expr::Matches` to be used not only for postgres but also for sqlite

### DIFF
--- a/src/backend/sqlite/query.rs
+++ b/src/backend/sqlite/query.rs
@@ -40,6 +40,13 @@ impl QueryBuilder for SqliteQueryBuilder {
         }
     }
 
+    fn prepare_bin_oper(&self, bin_oper: &BinOper, sql: &mut dyn SqlWriter) {
+        match bin_oper {
+            BinOper::Matches => write!(sql, "MATCH").unwrap(),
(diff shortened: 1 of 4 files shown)


ikrivosheev:
@evgeniy-terekhin thank you for PR! I think Postgres `Matches` and Sqlite `MATCH` are not equal...

evgeniy-terekhin:
> @evgeniy-terekhin thank you for PR! I think Postgres `Matches` and Sqlite `MATCH` are not equal...

can you please elaborate on this? and maybe suggest a different approach?

ikrivosheev:
@evgeniy-terekhin I think is better to move sqlite specific operators into new enum. Something like I did here: https://github.com/SeaQL/sea-query/pull/506.

@billy1624 @tyt2y3 what do you think?

evgeniy-terekhin:
@ikrivosheev ok when that PR gets merged I'll change this one accordingly

evgeniy-terekhin:
closing in favor of [this PR](https://github.com/SeaQL/sea-query/pull/513)

Issue #485, opened 2022-10-18
Negating an empty condition results in a syntax error


nitnelave:
## Description

```rust
assert_eq!(Query::select()
    .from(Users::Table)
    .column(Users::Name)
    .cond_where(Cond::all().not())
    .to_string(SqliteQueryBuilder),
    r#"SELECT name FROM users WHERE NOT ()"#
);
```

The empty `Cond::all()` gets translated to `""` instead of a syntactically valid statement.

`all()` should get represented as `SimpleExpr::Value(true)`, and `any()` as `SimpleExpr::Value(false)`.

ikrivosheev:
@nitnelave hello! I opened PR.

PR #505, opened 2022-11-05
issues-485 Fix syntax error

ikrivosheev:
## PR Info

- Closes https://github.com/SeaQL/sea-query/issues/485

## Bug Fixes

- Fix syntax error when have empty `Condition`

### DIFF
--- a/src/backend/query_builder.rs
+++ b/src/backend/query_builder.rs
@@ -1204,17 +1204,16 @@ pub trait QueryBuilder: QuotedBuilder + EscapeBuilder + TableRefBuilder {
         keyword: &str,
         sql: &mut dyn SqlWriter,
     ) {
-        if !condition.is_empty() {
-            write!(sql, " {} ", keyword).unwrap();
-        }
         match &condition.contents {
             ConditionHolderContents::Empty => (),
             ConditionHolderContents::Chain(conditions) => {
+                write!(sql, " {} ", keyword).unwrap();
                 for (i, log_chain_oper) in conditions.iter().enumerate() {
                     self.prepare_logical_chain_oper(log_chain_oper, i, conditions.len(), sql);
                 }
             }
             ConditionHolderContents::Condition(c) => {
+                write!(sql, " {} ", keyword).unwrap();
                 self.prepare_condition_where(c, sql);
             }
         }
@@ -1224,13 +1223,22 @@ pub trait QueryBuilder: QuotedBuilder + EscapeBuilder + TableRefBuilder {
     /// Translate part of a condition to part of a "WHERE" clause.
     fn prepare_condition_where(&self, condition: &Condition, sql: &mut dyn SqlWriter) {
         if condition.negate {
-            write!(sql, "NOT (").unwrap();
+            write!(sql, "NOT ").unwrap()
         }
-        let mut is_first = true;
-        for cond in &condition.conditions {
-            if is_first {
-                is_first = false;
-            } else {
+
+        if condition.is_empty() {
+            match condition.condition_type {
+                ConditionType::All => self.prepare_constant(&true.into(), sql),
+                ConditionType::Any => self.prepare_constant(&false.into(), sql),
+            }
+            return;
+        }
+
+        if condition.negate {
+            write!(sql, "(").unwrap();
+        }
+        condition.conditions.iter().fold(true, |first, cond| {
+            if !first {
                 match condition.condition_type {
                     ConditionType::Any => write!(sql, " OR ").unwrap(),
                     ConditionType::All => write!(sql, " AND ").unwrap(),
@@ -1256,7 +1264,8 @@ pub trait QueryBuilder: QuotedBuilder + EscapeBuilder + TableRefBuilder {
                     }
                 }
             }
-        }
+            false
+        });
         if condition.negate {
             write!(sql, ")").unwrap();
         }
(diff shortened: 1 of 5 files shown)


tests/mysql/query.rs
```diff
@@ -540,27 +540,27 @@ fn select_36() {
...
-    let (statement, values) = sea_query::Query::select()
+    let (statement, values) = Query::select()
         .column(Glyph::Id)
         .from(Glyph::Table)
         .cond_where(Cond::any().add(Cond::all()).add(Cond::any()))
-        .build(sea_query::MysqlQueryBuilder);
+        .build(MysqlQueryBuilder);
 
-    assert_eq!(statement, r#"SELECT `id` FROM `glyph`"#);
+    assert_eq!(statement, r#"SELECT `id` FROM `glyph` WHERE FALSE"#);
```
nitnelave, line 549: I think there's some unfortunate interaction between 2 features here:
 - Collapsing an empty any/all into a boolean
 - Not adding an any/all that's empty.

This should be `any(TRUE, FALSE)`, so `TRUE`

nitnelave (changes requested):
Maybe we should add a test with the negation of an empty any/all?

tests/mysql/query.rs
tyt2y3: Fair

tyt2y3:
Actually, I think we should remove the behaviour `Not adding an any/all that's empty`, this would make the reasoning much easier.

nitnelave:
Just got bitten by this bug again, with an `any(not(any()), not(any()))` which evaluated to true instead of false.

src/backend/query_builder.rs
```diff
@@ -1215,13 +1214,22 @@ pub trait QueryBuilder: QuotedBuilder + EscapeBuilder + TableRefBuilder {
...
-        let mut is_first = true;
-        for cond in &condition.conditions {
-            if is_first {
-                is_first = false;
-            } else {
+
+        if condition.is_empty() {
+            match condition.condition_type {
+                ConditionType::All => self.prepare_constant(&true.into(), sql),
+                ConditionType::Any => self.prepare_constant(&false.into(), sql),
```
tyt2y3, line 1223: I think we flipped the logic here (or this is a deep topic?)
`ANY of ()`: anything from an empty set, so `TRUE`?
`ALL of ()`: all of the set, which is empty, so much be `FALSE`?

tests/mysql/query.rs
tyt2y3: Update: I am try to understand why it output `FALSE` here:

I imagine the expected output to be:

```SQL
SELECT `id` FROM `glyph` WHERE TRUE OR FALSE
```

tyt2y3: Ah found it

tyt2y3:
@nitnelave Can you go over this again and may add some tests as you see fit?

ikrivosheev:
@tyt2y3 I think, we can compress conditional when we add it.

tyt2y3:
> we can compress conditional when we add it.

I imagine `TRUE OR FALSE` rarely appears in real life, so might not worth the effort?

nitnelave:
I'm of the opinion that the query builder should not try to be too smart, since that's the role of the DB. They'll have no problem with an extra "OR FALSE" in the query.

tests/mysql/query.rs
```diff
@@ -540,27 +540,30 @@ fn select_36() {
...
         .from(Glyph::Table)
         .cond_where(Cond::any().add(Cond::all()).add(Cond::any()))
-        .build(sea_query::MysqlQueryBuilder);
+        .build(MysqlQueryBuilder);
 
-    assert_eq!(statement, r#"SELECT `id` FROM `glyph`"#);
+    assert_eq!(
+        statement,
+        r#"SELECT `id` FROM `glyph` WHERE (TRUE) OR (FALSE)"#
+    );
```
nitnelave, line 552: Can we add a negation test as well?
`Cond::all().add(Cond::all().not()).add(Cond::any().not()).not()`

Should turn into `NOT (NOT TRUE OR NOT FALSE)`

nitnelave (approved):

ikrivosheev:
@nitnelave @tyt2y3 ok.

src/backend/query_builder.rs
```diff
@@ -1215,13 +1214,22 @@ pub trait QueryBuilder: QuotedBuilder + EscapeBuilder + TableRefBuilder {
...
+            }
+            return;
+        }
```
billy1624, line 1226: Sorry for the coming late. I wonder why they are different?
- `Cond::all()` => `TRUE`
- `Cond::any()` => `FALSE`

Shouldn't it both be `FALSE` to prevent unintentional empty condition?

nitnelave: They should each reduce to the identity element of the operation: See https://en.wikipedia.org/wiki/Boolean_algebra#Monotone_laws .

Basically, `all(a, b, c)` is equivalent to `all(true, a, b, c)`, so `all()` should be `all(true)`. Same for `any` and `false`.

In other terms, think about "I want to delete if _any_ of these conditions are true" (and you didn't add conditions) -> You don't want to delete.

tests/mysql/query.rs
ikrivosheev: @nitnelave done!

tyt2y3 (approved):

src/backend/query_builder.rs
tyt2y3: 👍

PR #506, opened 2022-11-05
Move postgres specific operator to new enum

ikrivosheev:
## Breaking Changes

- Move all postgres specific operator to own enum

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -3,6 +3,8 @@
 use crate::{expr::*, query::*, FunctionCall, ValueTuple, Values};
 use std::fmt;
 
+#[cfg(feature = "backend-postgres")]
+use crate::extension::postgres::PgBinOper;
 #[cfg(not(feature = "thread-safe"))]
 pub use std::rc::Rc as SeaRc;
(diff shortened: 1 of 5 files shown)


src/types.rs
```diff
@@ -136,25 +136,7 @@ pub enum BinOper {
...
-    WordSimilarity,
-    #[cfg(feature = "backend-postgres")]
-    StrictWordSimilarity,
-    #[cfg(feature = "backend-postgres")]
-    SimilarityDistance,
-    #[cfg(feature = "backend-postgres")]
-    WordSimilarityDistance,
-    #[cfg(feature = "backend-postgres")]
-    StrictWordSimilarityDistance,
+    PgOperators(PgBinOper),
```
evgeniy-terekhin, line 139: a minor remark but still: it might be better to have `PgOperator` (or `PgOp`/`PgOper`) as singular rather than plural (suggestion: remove trailing 's')

ikrivosheev: Thank you! I opened new PR: https://github.com/SeaQL/sea-query/pull/507 with fix)

PR #507, opened 2022-11-05
Move postgres specific operator to new enum

ikrivosheev:
## Breaking Changes

- Move all postgres specific operator to own enum

Continue: https://github.com/SeaQL/sea-query/pull/506


tyt2y3 (approved):
Agree )

Issue #484, opened 2022-10-18
Using `Iden` with `sqlx` is cumbersome


nitnelave:
## Motivation

When writing `sqlx` queries, to avoid using raw strings for column identifiers, the best way to write it is:

```rust
let query = Query::select()
    .from(Users::Table)
    .column(Users::Name)
    .to_string(SqliteQueryBuilder);

let names = sqlx::query(&query)
    .map(|row| row.get::<String, _>(&*Users::Name.to_string()))
    .fetch_all(&pool);
```

I'm mainly focusing on the `&*Users::Name.to_string()` part, which essentially resolves to `"name"`. Could we have a simpler way to express that, especially since they are often derived from enums?

## Proposed Solutions

If we want to keep `Iden` a potentially dynamic trait, then how about adding a new trait `StaticIden : Iden` that has a `.as_str()` method returning a `&'static str`? Then this could be implemented by the `#[derive(Iden)]` macro transparently, or we could add a `#[derive(StaticIden)]`.

Actually if we get `StaticIden` to implement `Into<&'static str>` then I'm sure I could convince `sqlx` devs to accept an `Into<&str>` rather than a `&str` for `row.get`, letting us write the much more elegant `row.get::<String, _>(Users::Name)`.

## Additional Information

[`sqlx::Row::get`](https://docs.rs/sqlx/latest/sqlx/trait.Row.html#method.get)

ikrivosheev:
@nitnelave hello! Thank you for the issue! Why we need new trait? Can we add a new method to `Iden`: `fn as_str(self) -> &'static str`?

ikrivosheev:
And I thin is better `AsRef<str>` then `Into<&str>`

nitnelave:
Adding a method would be a breaking change, that could disrupt anyone who implemented `Iden` based on non-static strings. There's no fundamental requirement that `Iden` instances are not dynamic, so it could be impossible to implement generally.

And +1 for `AsRef<str>`

phungleson:
Hey peeps, as this requires changes in sqlx, an easier stopgap solution is to add a trait like this (not necessary part of `sea-query`).

```
pub trait TryGetIden {
    fn try_get_iden<'a, T: Type<sqlx::Postgres> + Decode<'a, sqlx::Postgres>>(
        &'a self,
        iden: &impl Iden,
    ) -> Result<T, sqlx::Error>;
}
```

Then you can do like this `pg_row.try_get_iden(&Users::Name)`

[...]

PR #508, opened 2022-11-06
#484 Iden static

ikrivosheev:
## PR Info

- Closes https://github.com/SeaQL/sea-query/issues/484

## New Features

- Create `trait IdenStatic`

### DIFF
--- a/sea-query-derive/src/iden_variant.rs
+++ b/sea-query-derive/src/iden_variant.rs
@@ -1,4 +1,5 @@
 use std::convert::TryFrom;
+use std::marker::PhantomData;
 
 use heck::ToSnakeCase;
 use proc_macro2::{Span, TokenStream};
@@ -7,14 +8,47 @@ use syn::{Error, Fields, FieldsNamed, Ident, Variant};
 
 use crate::{error::ErrorMsg, find_attr, iden_attr::IdenAttr};
 
-pub struct IdenVariant<'a> {
+pub(crate) trait WriteArm {
+    fn variant(variant: TokenStream, name: TokenStream) -> TokenStream;
+    fn flattened(variant: TokenStream, name: &Ident) -> TokenStream;
+}
+
+pub(crate) struct DeriveIden;
+
+impl WriteArm for DeriveIden {
+    fn variant(variant: TokenStream, name: TokenStream) -> TokenStream {
+        quote! { Self::#variant => write!(s, "{}", #name).unwrap() }
+    }
+
+    fn flattened(variant: TokenStream, name: &Ident) -> TokenStream {
+        quote! { Self::#variant => #name.unquoted(s) }
+    }
+}
+
+pub(crate) struct DeriveIdenStatic;
+
+impl WriteArm for DeriveIdenStatic {
+    fn variant(variant: TokenStream, name: TokenStream) -> TokenStream {
+        quote! { Self::#variant => #name }
+    }
+
+    fn flattened(variant: TokenStream, name: &Ident) -> TokenStream {
+        quote! { Self::#variant => #name.as_str() }
+    }
+}
+
+pub(crate) struct IdenVariant<'a, T> {
     ident: &'a Ident,
     fields: &'a Fields,
     table_name: &'a str,
     attr: Option<IdenAttr>,
+    _p: PhantomData<T>,
 }
 
-impl<'a> TryFrom<(&'a str, &'a Variant)> for IdenVariant<'a> {
+impl<'a, T> TryFrom<(&'a str, &'a Variant)> for IdenVariant<'a, T>
+where
+    T: WriteArm,
+{
     type Error = Error;
 
     fn try_from((table_name, value): (&'a str, &'a Variant)) -> Result<Self, Self::Error> {
@@ -30,7 +64,10 @@ impl<'a> TryFrom<(&'a str, &'a Variant)> for IdenVariant<'a> {
     }
 }
 
-impl ToTokens for IdenVariant<'_> {
+impl<T> ToTokens for IdenVariant<'_, T>
+where
+    T: WriteArm,
+{
     fn to_tokens(&self, tokens: &mut TokenStream) {
         match self.fields {
             Fields::Named(named) => self.to_tokens_from_named(named, tokens),
@@ -40,7 +77,10 @@ impl ToTokens for IdenVariant<'_> {
     }
 }
 
-impl<'a> IdenVariant<'a> {
+impl<'a, T> IdenVariant<'a, T>
+where
+    T: WriteArm,
+{
     fn new(
         ident: &'a Ident,
         fields: &'a Fields,
@@ -73,6 +113,7 @@ impl<'a> IdenVariant<'a> {
             fields,
             table_name,
             attr,
+            _p: PhantomData::<T>,
         })
     }
 
@@ -85,7 +126,7 @@ impl<'a> IdenVariant<'a> {
             // Unwrapping the ident is also safe because a named field always has an ident.
             let capture = field.ident.as_ref().unwrap();
             let variant = quote! { #ident{#capture} };
-            write_flattened(variant, capture)
+            T::flattened(variant, capture)
         } else {
             let variant = quote! { #ident{..} };
             self.write_variant_name(variant)
@@ -101,7 +142,7 @@ impl<'a> IdenVariant<'a> {
             // The case where unnamed fields length is not 1 is handled by new
             let capture = Delegated.into();
             let variant = quote! { #ident(#capture) };
-            write_flattened(variant, &capture)
+            T::flattened(variant, &capture)
         } else {
             let variant = quote! { #ident(..) };
             self.write_variant_name(variant)
@@ -138,7 +179,7 @@ impl<'a> IdenVariant<'a> {
             })
             .unwrap_or_else(|| self.table_or_snake_case());
 
-        write_variant(variant, name)
+        T::variant(variant, name)
     }
 }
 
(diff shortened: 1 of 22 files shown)


ikrivosheev:
@nitnelave @tyt2y3 @billy1624 can you review my PR?

sea-query-derive/src/iden_variant.rs
```diff
@@ -7,14 +8,47 @@ use syn::{Error, Fields, FieldsNamed, Ident, Variant};
 
 use crate::{error::ErrorMsg, find_attr, iden_attr::IdenAttr};
 
-pub struct IdenVariant<'a> {
+pub trait WriteArm {
```
tyt2y3, line 11: Can this be `pub(crate)`

src/types.rs
```diff
@@ -30,6 +30,11 @@ macro_rules! iden_trait {
 
             fn unquoted(&self, s: &mut dyn fmt::Write);
         }
+
+        /// Identifier
+        pub trait IdenStatic: Iden  {
+            fn as_str(&self) -> &str;
+        }
```
tyt2y3, line 37: Can we make this definition same as SeaORM's ?

```rust
pub trait IdenStatic: Iden + Copy + Debug + 'static {
    /// Method to call to get the static string identity
    fn as_str(&self) -> &str;
}
```

This way the next step is to just remove the SeaORM version in the next major release.

tyt2y3 (approved):
Nice addition. I have a feeling like SeaQuery will be slowly more-and-more like SeaORM

billy1624 (approved):
Thanks!! @ikrivosheev

src/types.rs
nitnelave: Can we implement `AsRef<str>`? And potentially an `as_str` that returns a `&'static str`

ikrivosheev: @nitnelave Do you mean something like this:

```
impl<T> AsStr<str> for T 
where T: IdenStatic 
{
    ...
}
```
?

nitnelave: Yes, or at least make it part of the derive (but a blanket implementation is probably better).

ikrivosheev: Well, I try do it.

ikrivosheev: @nitnelave I cannot do this for generic type, because I got error: `Only traits defined in the current crate can be implemented for arbitrary types [E0117]`

ikrivosheev: @tyt2y3 or @billy1624 can someone explain why we need: `pub trait IdenStatic: Iden + Copy + Debug + 'static`? I understand only: `Iden` trait. other two - I don't...

ikrivosheev: @nitnelave I add `AsRef<str>` implementation, can you review?

sea-query-derive/src/iden_variant.rs
ikrivosheev: Done

ikrivosheev:
> Nice addition. I have a feeling like SeaQuery will be slowly more-and-more like SeaORM

I think this is good. I would like the SeaQuery to be the best query builder) And SeaORM - ORM and SeaQuery helps it for build query) This is one part of this.

src/types.rs
billy1624: Great question! I see your concern.

If we have `trait IdenStatic: Copy {}`, that means below isn't possible anymore because `String` doesn't `Copy`.

```rs
#[derive(Iden, EnumIter)]
enum Custom {
    Custom(String),
}
```

tyt2y3: I think `Debug` is just lazy-ness. It should be impl by hand. But we need that to be able to use `assert!`

nitnelave: @billy1624 I think it's intentional that you can't derive `IdenStatic` for a `String`, no? Unless I'm mistaken, the idea behind `IdenStatic` is that it's just a straightforward compile-time mapping between the enum name and a static string. What would be the meaning of `Custom(String)` here?

ikrivosheev: @nitnelave I think @billy1624 mean: https://github.com/SeaQL/sea-query/pull/508/files#diff-2a80134bd72b33f35551f1f2467f60acab30cad611bfc1c4999f59adeeecd221R20

sea-query-derive/tests/pass/flattened_unnamed.rs
```diff
@@ -28,7 +28,9 @@ fn main() {
     let expected = ["asset", "id", "asset_name", "user_id", "creation_date"];
     Asset::iter()
         .chain(std::iter::once(Asset::Creation(CreationInfo::Date)))
-        .map(|var| Iden::to_string(&var))
         .zip(expected)
-        .for_each(|(iden, exp)| assert_eq!(iden, exp))
+        .for_each(|(var, exp)| {
+            assert_eq!(Iden::to_string(&var), exp);
```
nitnelave, line 33: Why do you write it like that and not `var.to_string()` and `var.as_str()`?

sea-query-derive/tests/pass/3_level_flatten.rs
```diff
@@ -1,7 +1,7 @@
-use sea_query::Iden;
+use sea_query::{Iden, IdenStatic};
 use strum::{EnumIter, IntoEnumIterator};
 
-#[derive(Iden, EnumIter)]
+#[derive(Iden, IdenStatic, EnumIter)]
```
nitnelave, line 4: Out of curiosity, can we derive `IdenStatic` without deriving `Iden`?

src/types.rs
nitnelave: (I think your link got broken or something, it just links to the entire diff)

sea-query-derive/tests/pass/3_level_flatten.rs
ikrivosheev: No, we cannot...

src/types.rs
ikrivosheev: Ups, I mean: https://github.com/SeaQL/sea-query/blob/master/sea-query-derive/tests/pass/renaming_everything.rs#L20

nitnelave: I see. For me, that's a chief example of something that can't derive `IdenStatic`: it's not _actually_ static :)

tyt2y3: Yes you are right and I think it was what Billy meant as well.

sea-query-derive/tests/pass/flattened_unnamed.rs
ikrivosheev: Done

src/types.rs
ikrivosheev: @tyt2y3 @nitnelave @billy1624 done. Now trait looks:

```
pub trait IdenStatic: Iden + Copy {
    fn as_str(&self) -> &'static str;
}
```

sea-query-derive/tests/pass/renaming_table.rs
```diff
@@ -13,8 +13,8 @@ enum Something {
 
 fn main() {
     let expected = ["something_else", "id", "asset_name", "user_id"];
-    Something::iter()
-        .map(|var| Iden::to_string(&var))
-        .zip(expected)
-        .for_each(|(iden, exp)| assert_eq!(iden, exp))
+    Something::iter().zip(expected).for_each(|(var, exp)| {
+        assert_eq!(var.to_string(), exp);
+        assert_eq!(&var.as_str(), &exp)
```
nitnelave, line 18: Nit: Shouldn't that be `assert_eq!(var.as_str(), exp)`?

nitnelave (approved):

billy1624:
Hey @tyt2y3, I wonder why we need to move below into SeaQuery? Any good reasons?

```rs
pub trait IdenStatic: Iden + Copy + Debug + 'static {
    /// Method to call to get the static string identity
    fn as_str(&self) -> &str;
}
```

tyt2y3:
Having two symbols with the same name that works (more or less) the same is confusing.

sea-query-derive/tests/pass/3_level_flatten.rs
tyt2y3: I think this is possible if we also generate the following bit:

```rust
impl Iden for Character {
    fn unquoted(&self, s: &mut dyn std::fmt::Write) {
        write!(s, "{}", self.as_str()).unwrap();
    }
}
```

ikrivosheev:
@tyt2y3 @billy1624 @nitnelave I have done.  Can you have a look at it?

src/types.rs
```diff
@@ -34,6 +34,11 @@ macro_rules! iden_trait {
 
             fn unquoted(&self, s: &mut dyn fmt::Write);
         }
+
+        /// Identifier
+        pub trait IdenStatic: Iden + Copy {
```
tyt2y3, line 39:
```suggestion
        pub trait IdenStatic: Iden + Copy + 'static {
```

tyt2y3 (approved):

src/types.rs
```diff
@@ -34,6 +34,11 @@ macro_rules! iden_trait {
 
             fn unquoted(&self, s: &mut dyn fmt::Write);
         }
+
+        /// Identifier
+        pub trait IdenStatic: Iden + Copy + 'static {
```
ikrivosheev, line 39: @tyt2y3 why we need `'static` lifetime for `IdentStatic`? I understand why we need it for function result.

tyt2y3: This basically prevents `Box<dyn IdenStatic>` which SeaORM somehow needs.
Alternatively, we don't align this with SeaORM and so rename this trait to something else.

PR #524, opened 2022-11-22
Drop the Sized requirement on implementers of SchemaBuilders

nitnelave:
I'm implementing a dynamic query builder that just dispatches to either of Sqlite/Mysql/PostgresQueryBuilder dynamically, so I have an internal function returning a `&dyn GenericBuilder`. This works to implement (delegate) all the functions, except for `prepare_column_spec` that has a `Sized` requirement.

Here we move the `QueryBuilder` requirement to the trait instead of the function, allowing us to get rid of `Sized`.

### DIFF
--- a/src/backend/table_builder.rs
+++ b/src/backend/table_builder.rs
@@ -1,6 +1,8 @@
 use crate::*;
 
-pub trait TableBuilder: IndexBuilder + ForeignKeyBuilder + QuotedBuilder + TableRefBuilder {
+pub trait TableBuilder:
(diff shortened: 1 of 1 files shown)


nitnelave:
Actually, I can make do with `build_any`, so I don't need this PR. I still think it's useful, but up to you.

ikrivosheev (approved):
@nitnelave thank you! LGTM!

tyt2y3:
Interesting

PR #492, opened 2022-10-21
issues-490 Remove heap allocation

ikrivosheev:
## PR Info

- Closes https://github.com/SeaQL/sea-query/issues/490

## Changes

- Remove Heap allocation

PR #525, opened 2022-11-23
Deprecate MySQL column type with display length / precision

billy1624:
## PR Info

- In response to https://github.com/SeaQL/sea-query/pull/492#issuecomment-1288114667

## Breaking Changes

- [x] Integer and date time column types with display length or precision option has been removed.
- [x] Its corresponding constructor methods are also being removed, i.e. `ColumnDef::xxx_len()` method

### DIFF
--- a/examples/postgres/src/main.rs
+++ b/examples/postgres/src/main.rs
@@ -35,7 +35,7 @@ fn main() {
             .col(ColumnDef::new(Document::Timestamp).timestamp())
             .col(ColumnDef::new(Document::TimestampWithTimeZone).timestamp_with_time_zone())
             .col(ColumnDef::new(Document::Decimal).decimal())
-            .col(ColumnDef::new(Document::Array).array(ColumnType::Integer(None)))
+            .col(ColumnDef::new(Document::Array).array(ColumnType::Integer))
             .build(PostgresQueryBuilder),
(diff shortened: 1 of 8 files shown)


ikrivosheev (approved):
@billy1624 thank you! LGTM!

tyt2y3:
The timestamp precision in Postgres has a usability. Not sure whether we should retain it though. 
https://www.postgresql.org/docs/current/datatype-datetime.html

billy1624:
> The timestamp precision in Postgres has a utility. Not sure whether we should retain it though. https://www.postgresql.org/docs/current/datatype-datetime.html

I'm pretty sure it's optional and not specifying the precision will not prevent it from storing date time with fractional digits.

Quote: "where p is an optional precision specification giving the number of fractional digits in the seconds field. Precision can be specified for time, timestamp, and interval types, and can range from 0 to 6. If no precision is specified in a constant specification, it defaults to the precision of the literal value (but not more than 6 digits)."
https://www.postgresql.org/docs/current/datatype-datetime.html

And you can check this demo I just created: https://dbfiddle.uk/so8-Cie_

tyt2y3 (approved):

billy1624:
This should goes into 0.28.x

PR #540, opened 2022-11-30
Deprecated Expr::tbl

ikrivosheev:
I don't now why, but we have two very similar functions: `Expr::col` and `Expr::tbl`.

## Breaking Changes

- Deprecated `Expr::tbl`


billy1624 (commented):
Nice refactoring!

I think we do have other places also use `Expr::tbl`. E.g. https://github.com/ikrivosheev/sea-query/blob/e4a847152e84124446110ed375d3b9ab604f06b4/src/func.rs#L113

ikrivosheev:
@billy1624 hello! Good catch, thank you! Fix!

ikrivosheev:
@billy1624 @tyt2y3 can you review my changes? I think this is last PR before 0.28

Sytten:
col used to be only for the column name without the table table. This is necessary in certain condition like CTE that dont have a table name. Tbl used to be for table name and column name.
Just make sure all use cases are supported before deprecating...

ikrivosheev:
> col used to be only for the column name without the table table. This is necessary in certain condition like CTE that dont have a table name. Tbl used to be for table name and column name. Just make sure all use cases are supported before deprecating...

@Sytten, hi. Method signatures was the same.

PR #543, opened 2022-12-05
Don't use the sqlx's `macros` feature in sea-query-binder

ohadravid:
Hi 😄 

We have a use case where we want to avoid using `sqlx`'s `macros` feature, but it seems that `sea-query-binder` depends on the default features and so on `macros` as well, but as far as I can tell it's not really needed.

## Changes

- [ ] Don't depend on `sqlx`'s default features in `sea-query-binder`


ikrivosheev (approved):
@ohadravid hello! Thank you for your contribution! LGTM!

tyt2y3 (approved):

PR #549, opened 2022-12-13
docs: fix ilike typo

cemremengu:
like -> ilike


ikrivosheev (approved):
@cemremengu hello! Good catch!

Issue #516, opened 2022-11-09
`Function::Coalesce` not only support `SimpleExpr`


bingryan:
Test language=MySQL

## Description

I found it difficult to handle the logic that `Function::Coalesce` supports the coexistence of `SimpleExpr` and `QueryStatement`

The following logic is correct
```sql
create table bakery (
	id bigint,
	name varchar(255),
	profit_margin	
)


create table cake{
	id bigint,
	name varchar(255),
	price bigint,
	bakery_id bigint,
	gluten_free bigint,
	serial bigint
}
```

mysql query:

```sql
select
	name,
	bakery,
	EXISTS ( SELECT 1 FROM `cake` WHERE `id` = 1 ) AS `exist`,
	-- Coalesce not only support SimpleExpr, logic is correct
	COALESCE (
		-- QueryStatement
		( SELECT count(*) FROM `cake` WHERE `bakery_id` = 27 ), 
		-- SimpleExpr, select or const 
		(SELECT 0),
		0 
	) AS `count` 
from bakery
where id = 27
```

but the existing logic does not support  the coexistence of `SimpleExpr` and `QueryStatement`

https://github.com/SeaQL/sea-query/blob/master/src/func.rs#L430-L435

billy1624:
Hey @bingryan, thanks for the suggestions!! I think we can make `Func::coalesce` takes iterator of `Into<SimpleExpr>`.

```rs
impl Func {
    pub fn coalesce<I, T>(args: I) -> FunctionCall
    where
        I: IntoIterator<Item = T>,
        T: Into<SimpleExpr>,
    {
        FunctionCall::new(Function::Coalesce).args(args)
    }
}
```

Which also requires changing

```rs
impl FunctionCall {
    pub fn args<I, T>(mut self, args: I) -> Self
    where
        I: IntoIterator<Item = T>,
        T: Into<SimpleExpr>,
    {
        self.args = args.into_iter().map(Into::into).collect();
        self
    }
}
```

And implementing `From<SelectStatement> for SimpleExpr`

```rust
impl From<SelectStatement> for SimpleExpr {
    fn from(sel: SelectStatement) -> Self {
        SimpleExpr::SubQuery(
            None,
            Box::new(sel.into_sub_query_statement()),
        )
    }
}
```

Sounds like a good plan? @ikrivosheev

ikrivosheev:
@billy1624 I think we need only last one:

```rust
impl From<SelectStatement> for SimpleExpr {
[...]

PR #572, opened 2023-01-02
gh-516: support coexistence of SimpleExpr and QueryStatement

alphavector:
Took the liberty of implementing this feature to better understand the project

Closes: #516


ikrivosheev (approved):
@alphavector thank you! LGTM!

src/func.rs
```diff
@@ -407,7 +407,11 @@ impl Func {
     ///
     /// let query = Query::select()
     ///     .expr(Func::coalesce([
-    ///         Expr::col(Char::SizeW).into(),
+    ///         Query::select()
+    ///             .from(Char::Table)
+    ///             .expr(Func::max(Expr::col(Char::SizeW)))
+    ///             .to_owned()
```
tyt2y3, line 413: This can be `take()`

src/func.rs
```diff
@@ -407,7 +407,11 @@ impl Func {
...
+    ///             .into(),
```
tyt2y3, line 414: I would rather be more conservative on this type conversion. Would a method `into_sub_query_expr` with an explicit intent better for this use case?

tyt2y3 (commented):
Thank you for the contribution

src/func.rs
alphavector: Sorry for the slow reply.
Yes, perhaps `into_sub_query_expr` would be better (perhaps even better something like `into_scalar_subquery_expr` and implementing a separate `ScalarSubQueryStatement`)

But doesn't it create confusion that for `coalesce` SelectStatement needs to be explicitly converted to `SubQueryStatement`, and all other functions need to pass `SelectStatement` when called

Seems like this would require redoing the whole api or am I complicating things and it's simpler?

src/expr.rs
```diff
@@ -2163,6 +2163,12 @@ impl Expression for SimpleExpr {
     }
 }
 
+impl From<SelectStatement> for SimpleExpr {
```
tyt2y3, line 2166: What I mean is we remove this `From` and instead have a concrete impl `into_sub_query_expr` somewhere.

src/func.rs
```diff
@@ -416,15 +420,15 @@ impl Func {
     ///
     /// assert_eq!(
     ///     query.to_string(MysqlQueryBuilder),
-    ///     r#"SELECT COALESCE(`size_w`, `size_h`, 12) FROM `character`"#
+    ///     r#"SELECT COALESCE((SELECT MAX(`size_w`) FROM `character`), `size_h`, 12) FROM `character`"#
```
tyt2y3, line 423: And btw, can you add a new test case instead? For better test coverage.

ikrivosheev:
@alphavector hello! Any updates?

alphavector:
Sorry for the long reply...

src/func.rs
alphavector: What do you mean by a new test case? I added test cases, insert_coalesce for example. The current test case involves checking for a subquery in coalesce

src/expr.rs
```diff
@@ -37,6 +37,10 @@ pub enum SimpleExpr {
     Constant(Value),
 }
 
+pub trait IntoSimpleExpr {
+    fn into_sub_query_expr(self) -> SimpleExpr;
```
ikrivosheev, line 41: @alphavector hello! Naming confused me a little bit) 

@tyt2y3 what do you think?

```rust
pub trait IntoSimpleExpr {
   fn into_simple_expr(self) -> SimpleExpr;
}
```

tyt2y3: Actually, there is only one impl for this trait. May be we can simply make it an inherit method?

PR #576, opened 2023-01-09
Fix quoted string bug while inserting array of strings to Postgres

### LINKED ISSUE #573: Error inserting document containing array in Postgres
## Description

Generating an insert query (and possibly other types of queries, haven't tested) returns an invalid query. The array is return in values as `VALUES ('{'value'}')` when it should be `VALUES ('{value}')`.

## Steps to Reproduce

1. Generate the insert query
2. Execute it

### Expected Behavior

It should generate a valid insert query but it doesn't.

### Reproduces How Often

Always

## Versions

```
├── sea-query v0.28.1
│   ├── sea-query-attr v0.1.1 (proc-macro)
│   └── sea-query-derive v0.3.0 (proc-macro)
```

## Additional Information

```rs
#[enum_def]
#[derive(Debug, Serialize, Deserialize)]
pub struct RefreshTokenTree {
    pub id: String,
    pub tokens: Vec<String>,
}

Table::create()
            .table(RefreshTokenTreeIden::Table)
            .if_not_exists()
            .col(
                ColumnDef::new(RefreshTokenTreeIden::Id)
                    .string()
                    .not_null()
                    .primary_key(),
            )
            .col(
                ColumnDef::new(RefreshTokenTreeIden::Tokens)
                    .array(ColumnType::String(None))
                    .not_null(),
            )
            .to_string(PostgresQueryBuilder)

Query::insert()
            .into_table(RefreshTokenTreeIden::Table)
            .columns(vec![
                RefreshTokenTreeIden::Id,
                RefreshTokenTreeIden::Tokens,
            ])
            .values_panic(vec![
                self.id.clone().into(),
                self.tokens.clone().into(),
            ])
            .to_string(PostgresQueryBuilder)
```

sh-soltanpour:
## PR Info

- Closes #573 

- Dependencies:
  - 

- Dependents:
  - 

## Bug Fixes

- [X] Fix the bug while generating an insert query string. The result should be like `VALUES '{val}'` while now it's like `VALUES '{'val'}'`.

### DIFF
--- a/tests/postgres/query.rs
+++ b/tests/postgres/query.rs
@@ -1256,6 +1256,21 @@ fn insert_8() {
 }
 
 #[test]
+fn insert_10() {
+    assert_eq!(
+        Query::insert()
+            .into_table(Glyph::Table)
+            .columns([Glyph::Aspect, Glyph::Tokens])
+            .values_panic([
+                3.1415.into(),
+                vec!["Token1".to_string(), "Token2".to_string(), "Token3".to_string()].into()
+            ])
+            .to_string(PostgresQueryBuilder),
+        r#"INSERT INTO "glyph" ("aspect", "tokens") VALUES (3.1415, '{Token1,Token2,Token3}')"#
+    );
+}
+
+#[test]
 #[allow(clippy::approx_constant)]
 fn insert_on_conflict_1() {
     assert_eq!(
--- a/tests/postgres/query.rs
+++ b/tests/postgres/query.rs
@@ -1271,7 +1271,7 @@ fn insert_10() {
                 .into()
             ])
             .to_string(PostgresQueryBuilder),
-        r#"INSERT INTO "glyph" ("aspect", "tokens") VALUES (3.1415, '{Token1,Token2,Token3}')"#
+        r#"INSERT INTO "glyph" ("aspect", "tokens") VALUES (3.1415, ARRAY ['Token1','Token2','Token3'])"#
     );
 }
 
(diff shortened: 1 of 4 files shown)


sh-soltanpour:
Hi @billy1624 @ikrivosheev,

I believe this is ready for review, I'm not sure if I have access to change the status to 'Ready for review' or not.

Thanks.

tests/postgres/query.rs
```diff
@@ -1255,6 +1255,26 @@ fn insert_8() {
...
+                3.1415.into(),
+                vec![
+                    "Token1".to_string(),
+                    "Token2".to_string(),
+                    "Token3".to_string()
+                ]
+                .into()
+            ])
+            .to_string(PostgresQueryBuilder),
+        r#"INSERT INTO "glyph" ("aspect", "tokens") VALUES (3.1415, '{Token1,Token2,Token3}')"#
```
tyt2y3, line 1274: Umm I am not sure whether the `Token1` should be quoted at all?
It seems more readable using the `ARRAY` syntax instead?
```sql
INSERT INTO "glyph" ("aspect", "tokens") VALUES (3.1415, ARRAY ['Token1','Token2','Token3'])
```

sh-soltanpour: If I understand your first note correctly, I think yes, we should put those strings inside quotes, otherwise, the database would think of them as they're variables or column names and it throws an error like this: 
`ERROR: column "Token1" does not exist`

About using `ARRAY`: I agree, I can make this change as well. Thanks for noting.

ikrivosheev (approved):
@sh-soltanpour thank you! LGTM! @tyt2y3 or @billy1624 can someone review the PR?)

tyt2y3 (approved):

sh-soltanpour:
Thanks for your approval!
I think it's ready to merge @ikrivosheev. Thanks

PR #582, opened 2023-01-12
Add support for bitwise aggregate functions

tooboredtocode:
## New Features

- [x] Bitwise Aggregate Functions
  - [x] Bitwise And
  - [x] Bitwise Or

## Breaking Changes

This change is backwards compatible

### DIFF
--- a/src/backend/query_builder.rs
+++ b/src/backend/query_builder.rs
@@ -621,6 +621,8 @@ pub trait QueryBuilder: QuotedBuilder + EscapeBuilder + TableRefBuilder {
                     Function::Cast => "CAST",
                     Function::Lower => "LOWER",
                     Function::Upper => "UPPER",
+                    Function::BitAnd => "BIT_AND",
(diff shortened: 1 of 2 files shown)


ikrivosheev (commented):
@tooboredtocode hello! Thank you for the PR. If I understand correctly functions support only by Mysql and Postgres. 
I think, it's better to add it here: https://github.com/SeaQL/sea-query/tree/master/src/extension

@billy1624 @tyt2y3 what do you think about it?

tyt2y3:
Umm the idea has been "If 2 out of 3 backends support it we can put it on the common API. If the feature is specific to 1 backend then we put them behind extensions"

tyt2y3 (approved):

tooboredtocode:
It might be worth putting it behind the mysql and postgres features, what do you think?

ikrivosheev (approved):
@tooboredtocode thank you! LGTM!

PR #584, opened 2023-01-13
Compile test_cfg when needed and hide it from docs

billy1624:
## Breaking Changes

- [x] `tests_cfg` module available if and only if you enabled `tests-cfg` feature

Hide implementors that are purely for testing purposes:

> image
> 
> https://docs.rs/sea-query/latest/sea_query/types/trait.Iden.html#implementors


ikrivosheev (approved):
@billy1624 is a good catch! LGTM!

PR #589, opened 2023-01-24
Support deferrable constraints

smrtrfszm:
## New Features

- [x] Deferrable constraints

## Breaking Changes

It is backwards compatible.

### DIFF
--- a/src/foreign_key/common.rs
+++ b/src/foreign_key/common.rs
@@ -10,6 +10,7 @@ pub struct TableForeignKey {
     pub(crate) ref_columns: Vec<DynIden>,
     pub(crate) on_delete: Option<ForeignKeyAction>,
     pub(crate) on_update: Option<ForeignKeyAction>,
+    pub(crate) deferrable: Option<Deferrable>,
 }
 
 /// Foreign key on update & on delete actions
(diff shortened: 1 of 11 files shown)


src/foreign_key/common.rs
```diff
@@ -10,6 +10,8 @@ pub struct TableForeignKey {
     pub(crate) ref_columns: Vec<DynIden>,
     pub(crate) on_delete: Option<ForeignKeyAction>,
     pub(crate) on_update: Option<ForeignKeyAction>,
+    #[cfg(any(feature = "backend-postgres", feature = "backend-sqlite"))]
+    pub(crate) deferrable: Option<Deferrable>,
```
billy1624, line 14: I think we should split `Deferrable` and `Initially` into two separate enum. i.e.

```rs
enum Deferrable { // Or represent it as a `bool`?
    Yes,
    No,
}

enum Initially {
    Deferred,
    Immediate,
}
```

src/foreign_key/common.rs
billy1624, line 1: And we should not add any feature gate `#[cfg(feature = ...)]`

billy1624:
Hey @smrtrfszm, sorry for the delay. Thanks for contributing! I just leave some comments :)

PR #592, opened 2023-01-30
Fix clippy warning

billy1624:
## House Keeping

- [x] Fix clippy warnings


ikrivosheev:
@billy1624 thank you!

PR #594, opened 2023-02-01
String parameter should take generic `Into<String>`

billy1624:
## PR Info

- Dependents:
  - 

## Breaking Changes

- [ ] Changed the parameter of to `LikeExpr::new(T) where T: Into<String>`

## Changes

- [ ] Provide blanket `IntoLikeExpr` implementations to types implemented `Into<String>`

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -410,9 +410,12 @@ impl Iden for NullAlias {
 }
 
 impl LikeExpr {
-    pub fn new(pattern: String) -> Self {
+    pub fn new<T>(pattern: T) -> Self
+    where
+        T: Into<String>,
+    {
         Self {
-            pattern,
+            pattern: pattern.into(),
             escape: None,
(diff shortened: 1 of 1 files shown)


ikrivosheev:
@billy1624 hello! I have pushed new commit.

billy1624:
Hey @ikrivosheev, sorry for the delay. Had been starting a new proprietary project in the past two weeks loll. Now, I'm back :)

ikrivosheev (approved):
@billy1624 hello! LGTM!

tyt2y3 (commented):
I think this has a real benefit - if given a String, we don't have to allocate a second time.

billy1624:
> I think this has a real benefit - if given a String, we don't have to allocate a second time.

Yes, we only `.clone()` / `.to_owned()` while necessary

PR #598, opened 2023-02-06
Hashable value

tyt2y3:
Fixes https://github.com/SeaQL/sea-orm/issues/1254

I have considered different implementations, and using a derive macro seems to be the easiest to maintain.

However, there is a catch when this feature is enabled: NaN == NaN, which contradicts Rust's built-in implementation of NaN != NaN


ikrivosheev (approved):
@tyt2y3 good! Thank you! LGTM!

PR #606, opened 2023-02-20
Impl prepare if all names are valid

tyt2y3:
This improves the Iden, IdenStatic macros to be "smarter": where all idens are statically known to be valid, provide a short-circuit implementation of prepare.

```rust
#[derive(Debug, Iden)]
pub enum Char {
    Table,
    Id,
    #[iden = "hello"]
    Character,
}

// expanded
impl sea_query::Iden for Char {
    fn prepare(&self, s: &mut dyn ::std::fmt::Write, q: char) {
        s.write_fmt(format_args!("{0}", q)).unwrap();
        self.unquoted(s);
        s.write_fmt(format_args!("{0}", q)).unwrap();
    }
    fn unquoted(&self, s: &mut dyn ::std::fmt::Write) {
        match self {
            Self::Table => s.write_fmt(format_args!("{0}", "char")).unwrap(),
            Self::Id => s.write_fmt(format_args!("{0}", "id")).unwrap(),
            Self::Character => s.write_fmt(format_args!("{0}", "hello")).unwrap(),
        };
    }
}
```

```rust
#[derive(Debug, Iden)]
pub enum Char {
    Table,
    Id,
    #[iden = "hel-lo"]
    Character,
}

// expanded
impl sea_query::Iden for Char {
    // `prepare` fallback to default impl
    fn unquoted(&self, s: &mut dyn ::std::fmt::Write) {
        match self {
            Self::Table => s.write_fmt(format_args!("{0}", "char")).unwrap(),
            Self::Id => s.write_fmt(format_args!("{0}", "id")).unwrap(),
            Self::Character => s.write_fmt(format_args!("{0}", "hel-lo")).unwrap(),
        };
    }
}
```

Tried to benchmark it, but the improvement is ~10% at best, so probably does not relevant.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,7 +43,7 @@ mac_address = { version = "1.1", default-features = false, optional = true }
 
 [dev-dependencies]
 sea-query = { path = ".", features = ["tests-cfg"] }
-criterion = { version = "0.3", features = ["html_reports"] }
+criterion = { version = "0.3" }
 pretty_assertions = { version = "1" }
 
 [features]
--- a/benches/basic.rs
+++ b/benches/basic.rs
@@ -1,5 +1,12 @@
 use criterion::{criterion_group, criterion_main, Criterion};
-use sea_query::{tests_cfg::*, *};
+use sea_query::*;
+
+#[derive(Debug, Iden)]
+pub enum Char {
+    Table,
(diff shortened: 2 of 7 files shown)


tyt2y3:
To be handled together with 607

PR #607, opened 2023-02-21
Dual quoting

tyt2y3:
Allow Quote to be potentially different on left and right

Depends on https://github.com/SeaQL/sea-query/pull/606


billy1624 (commented):
Hey @tyt2y3, I just committed some refactoring. Hope it make the API more ergonomic :)

tyt2y3:
Thanks. Although that's not an API normal people would use anyway.

PR #609, opened 2023-02-22
Make `OnConflict` store both `update_columns` and `values`

### LINKED ISSUE #593: `OnConflict::values` and `OnConflict::update_columns` gets overridden by one another
## Description

Since `OnConflict` uses the builder pattern, it is expected that a call to `values` then `update_columns`, or vice-versa, should create a query with both features, but only the last one gets applied to the query.

## Steps to Reproduce

1. Build an `InsertStatement` with an `OnConflict` clause
2. Add actions using `OnConflict::values([(Column::Col1, false.into())])` and `OnConflict::update_columns([Column::Col2])`
3. Build and check the generated query

### Expected Behavior

PostgreSQL:
   `... DO UPDATE SET "col1" = false, "col2" = "excluded"."col2"`

### Actual Behavior

PostgreSQL:
   `... DO UPDATE SET "col2" = "excluded"."col2"`

Only `OnConflict::update_columns` gets applied since it was used last

### Reproduces How Often

Everytime

## Versions
0.24+ (when `OnConflict` was introduced)

elbertronnie:
## PR Info

- Closes #593 

## Breaking Changes

- [x] `OnConflict::values` and `OnConflict::update_columns` will append the new values keeping the old values intact instead of erasing them

## Changes

- [x] Added new enum `OnConflictUpdate` that determines the column update strategy in ON CONFLICT expressions
- [x] Replaced `UpdateColumns` and `UpdateExprs` with `Update(Vec<OnConflictUpdate>)` in `OnConflictAction`

### DIFF
--- a/src/query/on_conflict.rs
+++ b/src/query/on_conflict.rs
@@ -20,10 +20,17 @@ pub enum OnConflictTarget {
 pub enum OnConflictAction {
     /// Do nothing
     DoNothing,
+    /// Update column value of existing row
+    Update(Vec<OnConflictUpdate>),
+}
+
+/// Represents strategies to update column in ON CONFLICT (upsert) actions
+#[derive(Debug, Clone)]
+pub enum OnConflictUpdate {
     /// Update column value of existing row with inserting value
-    UpdateColumns(Vec<DynIden>),
+    Column(DynIden),
     /// Update column value of existing row with expression
-    UpdateExprs(Vec<(DynIden, SimpleExpr)>),
+    Expr(DynIden, SimpleExpr),
 }
 
 impl OnConflict {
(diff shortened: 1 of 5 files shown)


src/query/on_conflict.rs
```diff
@@ -20,10 +20,17 @@ pub enum OnConflictTarget {
...
+/// Represents strategies to update column in ON CONFLICT (upsert) actions
+#[derive(Debug, Clone)]
+pub enum UpdateStrat {
     /// Update column value of existing row with inserting value
-    UpdateColumns(Vec<DynIden>),
+    Column(DynIden),
     /// Update column value of existing row with expression
-    UpdateExprs(Vec<(DynIden, SimpleExpr)>),
+    Expr(DynIden, SimpleExpr),
 }
```
billy1624, line 34: I got a suggested naming for `UpdateStrat`, maybe `OnConflictUpdate`?

billy1624 (commented):
Hey @elbertronnie, thanks for contributing!! The implementation looks good to me :)

src/query/on_conflict.rs
elbertronnie: I have changed the name to `OnConflictUpdate`

ikrivosheev (approved):
@elbertronnie thank you! LGTM!

tyt2y3 (commented):
It would be nice to also add some doc tests

PR #614, opened 2023-03-05
Replace shim to inheret crate

ikrivosheev:
## PR Info

I am continue working on more ergonomic API. We have a lot duplicated code with add methods into Struct from traits (like OrderedStatement and other...). I found a crate: https://docs.rs/inherent/latest/inherent/. It makes something like we did in `shim.rs`. I tried to replace `shim.rs` to `inherent`. 

@billy1624 @tyt2y3 could you review please? I found it very useful)

After this PR all methods from `OrderedStatement`, `SchemaStatementBuilder`, `QueryStatementWriter`, `QueryStatementBuilder`, `ConditionalStatement` can be called without import these traits.


tyt2y3:
I like the idea of using `inherent`. Although we will need to amend our claim of "SeaQuery has no dependency" to "SeaQuery has minimal dependency"

billy1624 (commented):
I like the fact that one no longer has to remember to import the trait into scope before utilizing trait methods :)
If this works well, we should bring this into SeaORM.

ikrivosheev:
@billy1624 @tyt2y3 PR have done and is ready to review)

tyt2y3 (approved):

tyt2y3:
Seems nice!

PR #623, opened 2023-03-19
Fix: Dropping multiple types in Postgres

PreetamSing:
## Bug Fixes

- [x] Following code in Rust omits the comma in SQL statement, which causes error from Postgres (as of `sea-query` version `0.28.3`). This issue has been fixed.
```rust
use sea_query::{extension::postgres::Type, *};

#[derive(Iden)]
enum KycStatus {
   #[iden = "kyc_status"]
   Type,
   Pending,
   Approved,
}

#[derive(Iden)]
enum FontFamily {
   #[iden = "font_family"]
   Type,
   Aerial,
   Forte,
}

assert_eq!(
    Type::drop()
        .if_exists()
        .names([
            SeaRc::new(KycStatus::Type) as DynIden,
            SeaRc::new(FontFamily::Type) as DynIden,
        ])
        .cascade()
    r#"DROP TYPE IF EXISTS "kyc_status""font_family" CASCADE"#
);
```

## Changes

- [x] Add comma if multiple names are passed to `TypeDropStatement`.
- [x] Doc-test/example for dropping multiple types in single `TypeDropStatement`.

## Other considerations
- I'm fairly new to this code and have, sort of, reverse engineered this bug to fix it. Hence, I'm not fully aware of how my code can affect other things, though all the tests passed.
- Please update the dependency in `sea-orm` crate, if you merge this.

### DIFF
--- a/src/backend/postgres/types.rs
+++ b/src/backend/postgres/types.rs
@@ -35,7 +35,13 @@ impl TypeBuilder for PostgresQueryBuilder {
             write!(sql, "IF EXISTS ").unwrap();
         }
 
-        for name in drop.names.iter() {
+        let mut names_iter = drop.names.iter();
+        if let Some(name) = names_iter.next() {
+            self.prepare_type_ref(name, sql);
+        }
+
+        for name in names_iter {
+            write!(sql, ", ").unwrap();
             self.prepare_type_ref(name, sql);
         }
 
--- a/src/backend/postgres/types.rs
+++ b/src/backend/postgres/types.rs
@@ -35,15 +35,13 @@ impl TypeBuilder for PostgresQueryBuilder {
             write!(sql, "IF EXISTS ").unwrap();
         }
 
-        let mut names_iter = drop.names.iter();
-        if let Some(name) = names_iter.next() {
+        drop.names.iter().fold(true, |first, name| {
+            if !first {
+                write!(sql, ", ").unwrap();
+            }
             self.prepare_type_ref(name, sql);
-        }
-
-        for name in names_iter {
-            write!(sql, ", ").unwrap();
-            self.prepare_type_ref(name, sql);
-        }
+            false
+        });
 
         if let Some(option) = &drop.option {
             write!(sql, " ").unwrap();
--- a/src/extension/postgres/types.rs
+++ b/src/extension/postgres/types.rs
@@ -252,18 +252,18 @@ impl TypeDropStatement {
     ///
     /// #[derive(Iden)]
     /// enum KycStatus {
-    ///    #[iden = "kyc_status"]
-    ///    Type,
-    ///    Pending,
-    ///    Approved,
+    ///     #[iden = "kyc_status"]
+    ///     Type,
+    ///     Pending,
+    ///     Approved,
     /// }
     ///
     /// #[derive(Iden)]
     /// enum FontFamily {
-    ///    #[iden = "font_family"]
-    ///    Type,
-    ///    Aerial,
-    ///    Forte,
+    ///     #[iden = "font_family"]
+    ///     Type,
+    ///     Aerial,
+    ///     Forte,
     /// }
     ///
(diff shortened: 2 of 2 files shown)


src/backend/postgres/types.rs
```diff
@@ -35,7 +35,13 @@ impl TypeBuilder for PostgresQueryBuilder {
             write!(sql, "IF EXISTS ").unwrap();
         }
 
-        for name in drop.names.iter() {
+        let mut names_iter = drop.names.iter();
```
ikrivosheev, line 38: I suggest another option:
```rust
drop.names.fold(true, |first, name| {
...
});
```
@PreetamSing what do you sink about it?)

ikrivosheev (commented):
@PreetamSing hello! Thank for the PR. A small suggestion)

src/backend/postgres/types.rs
PreetamSing: I cannot think of any way for using `fold` method without leaving an unnecessary comma after the last name, which in Postgres will throw an error. Like following:
```sql
DROP TYPE IF EXISTS "kyc_status", "font_family",  CASCADE;
```
Though, enlighten me if I'm wrong.

ikrivosheev: I mean:
```rust
drop.names.fold(true, |first, name| {
    if !first {
       write!(sql ", ").unwrap();
    }
    self.prepare_type_ref(name, sql);
    false
});
```

PreetamSing: That works. Though, I think we are evaluating an **if** condition for every iteration, which is unnecessary and is saved from in my initial code.

ikrivosheev (approved):
@PreetamSing thank you! LGTM!

ikrivosheev:
@billy1624 @tyt2y3 can we make a small patch  release with the fix?

src/backend/postgres/types.rs
billy1624: Hey @PreetamSing, I'm going to suggest something similar to what @ikrivosheev said.

In terms of performance, both implementation is roughly the same, with BigO `O(n)`. Benchmark: https://github.com/SeaQL/sea-query/commit/2bbfa34818091ed337c327929409da23b33ad380

```
     Running benches/basic.rs (target/release/deps/basic-2f2ede7235026e9f)

mut_iter                time:   [26.429 ms 26.527 ms 26.628 ms]

fold                    time:   [26.130 ms 26.203 ms 26.286 ms]

Found 8 outliers among 100 measurements (8.00%)
  5 (5.00%) high mild
  3 (3.00%) high severe
```

However, the `.fold(...)` implementation is optimized for readability and maintainability. So I'm more in favour of the refactoring.

billy1624 (changes requested):
Hey @PreetamSing, thanks for contributing!! What your thoughts on this? https://github.com/SeaQL/sea-query/pull/623/files#r1142999511

ikrivosheev (approved):

ikrivosheev:
@billy1624, can you review a PR? And my last questuon is: do you want to include it into 0.29 release?)

src/backend/postgres/types.rs
```diff
@@ -35,9 +35,13 @@ impl TypeBuilder for PostgresQueryBuilder {
             write!(sql, "IF EXISTS ").unwrap();
         }
 
-        for name in drop.names.iter() {
```
tyt2y3, line 38: `iter().enumerate()` and `if i > 0` might align more with the coding style of the repo.

src/backend/postgres/types.rs
tyt2y3: Umm... I like `iter().enumerate()` and `if i > 0` slightly more, just because it is more inline with the original code structure.
But either way is fine.

tyt2y3 (approved):
Thank you for the contribution.

PR #624, opened 2023-03-20
fix: enable required `syn` features

billy1624:
## PR Info

- [ ] We need a PATCH release for this

## Bug Fixes

- [x] Enabled required `syn` v1 features. Those features were enabled implicitly by other crates that depends on `syn`. Now, they have upgraded to `syn` v2 as a result those features of `syn` v1 is not enabled.

### DIFF
--- a/sea-query-derive/Cargo.toml
+++ b/sea-query-derive/Cargo.toml
@@ -15,7 +15,7 @@ rust-version = "1.60"
 proc-macro = true
 
 [dependencies]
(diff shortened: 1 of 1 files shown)


billy1624:
Hey @ikrivosheev, sorry. I just double checked. The bug was introduced after 0.28.x. So, we don't need to backport the changes in this PR. Could you please revert the cherry-pick? Thanks!!

PR #629, opened 2023-04-07
0.28.x into master

ikrivosheev:
Merge 0.28.x into master


tyt2y3:
Thanks. May be https://github.com/SeaQL/sea-query/pull/576 is also worth backporting.

PR #638, opened 2023-05-16
Resolving `IndexCreateStatement` not generating a valid query when no name is provided

mohs8421:
On mysql systems, the `IndexCreateStatement` will generate a name, if no name was provided, resulting in always having a valid statement regarding the name.

## PR Info

- Closes https://github.com/SeaQL/sea-orm/discussions/1620

## Bug Fixes

- [x] The IndexCreateStatement does not work with mysql, if the name is not provided

### DIFF
--- a/src/index/create.rs
+++ b/src/index/create.rs
@@ -229,6 +229,16 @@ impl IndexCreateStatement {
         &self.index
     }
 
+    pub(crate) fn get_name(&self) -> String {
+        if let Some(name) = self.index.name.clone() {
+            return name;
+        }
+        let prefix = if self.is_primary_key() {
+            "pri"
+        } else if self.is_unique_key() { "uni" } else { "idx" };
+        format!("{}-{}", prefix, self.index.get_column_names().join("-"))
+    }
+
     pub fn take(&mut self) -> Self {
         Self {
             table: self.table.take(),
--- a/src/backend/mysql/index.rs
+++ b/src/backend/mysql/index.rs
@@ -1,5 +1,7 @@
 use super::*;
 
+const MYSQL_IDENTIFIER_NAME_MAX_LENGTH: usize = 64;
+
 impl IndexBuilder for MysqlQueryBuilder {
     fn prepare_table_index_expression(
         &self,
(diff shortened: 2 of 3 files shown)


billy1624 (commented):
Hey @mohs8421, nice stuffs!! Yes, then we always have a proper index name regardless.

@ikrivosheev do you still remember the "bug" where MySQL constraint / index name exceed the maximum length of MySQL's limit? Looks like @mohs8421 just truncate the long name.

mohs8421:
> Hey @mohs8421, nice stuffs!! Yes, then we always have a proper index name regardless.
> 
> @ikrivosheev do you still remember the "bug" where MySQL constraint / index name exceed the maximum length of MySQL's limit? Looks like @mohs8421 just truncate the long name.

well I didn't want to introduce an other source of error, so I just looked up the limit, although I know that this approach is still flawed because it could abreviate more properly, but then again, it is rather unlikely that somebody introduces an index, which exceeds this limit and have two of these indexes on the same table. In that case that user would still be able to manually specify a name, so nothing lost on that end.

src/index/create.rs
```diff
@@ -229,6 +229,20 @@ impl IndexCreateStatement {
         &self.index
     }
 
+    pub(crate) fn get_name(&self) -> String {
+        if let Some(name) = self.index.name.clone() {
+            return name;
+        }
+        let prefix = if self.is_primary_key() {
+            "pri"
```
ikrivosheev, line 237: I'm not sure it's a good solution to hard-code prefixes in the library... @billy1624 what do you think?

src/index/create.rs
```diff
@@ -229,6 +229,20 @@ impl IndexCreateStatement {
...
+        } else if self.is_unique_key() {
+            "uni"
+        } else {
+            "idx"
+        };
+        format!("{}-{}", prefix, self.index.get_column_names().join("-"))
```
ikrivosheev, line 243: This is similar to the preview comment...

ikrivosheev (changes requested):
@mohs8421 thank you for the PR! Some comment.

ikrivosheev:
> Hey @mohs8421, nice stuffs!! Yes, then we always have a proper index name regardless.
> 
> @ikrivosheev do you still remember the "bug" where MySQL constraint / index name exceed the maximum length of MySQL's limit? Looks like @mohs8421 just truncate the long name.

@billy1624 hello! Yes, I do. Well, It's makes sense to truncate the long name...

src/index/create.rs
mohs8421: I think it is ok in these cases because:
it was like that before at other locations you can already find these prefixes.
This is merely a fallback if nothing has been specified by the user.

ikrivosheev: CC @billy1624 @tyt2y3)

PR #641, opened 2023-05-18
Add support for custom operators

rakbladsvalsen:
## PR Info

See SeaORM discord 

## New Features

- Add support for arbitrary custom operators
- Add `Regex` op for Postgres
- Add `RegexCaseInsensitive` op for Postgres

## Notes

This PR is still a little bit rough around the edges, but hey, I'm open to comments. :)

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -72,8 +72,6 @@ impl Clone for SeaRc<dyn Iden> {
     }
 }
 
-impl Eq for SeaRc<dyn Iden> {}
-
 impl PartialEq for SeaRc<dyn Iden> {
     fn eq(&self, other: &Self) -> bool {
         let (self_vtable, other_vtable) = unsafe {
@@ -160,7 +158,7 @@ pub enum UnOper {
 }
 
 /// Binary operator
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum BinOper {
     And,
     Or,
@@ -187,7 +185,7 @@ pub enum BinOper {
     RShift,
     As,
     Escape,
-    Custom(DynIden),
+    Custom(&'static str),
     #[cfg(feature = "backend-postgres")]
     PgOperator(PgBinOper),
     #[cfg(feature = "backend-sqlite")]
--- a/src/types.rs
+++ b/src/types.rs
@@ -72,6 +72,8 @@ impl Clone for SeaRc<dyn Iden> {
     }
 }
 
+impl Eq for SeaRc<dyn Iden> {}
+
 impl PartialEq for SeaRc<dyn Iden> {
     fn eq(&self, other: &Self) -> bool {
         let (self_vtable, other_vtable) = unsafe {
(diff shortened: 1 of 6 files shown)


rakbladsvalsen:
Some questions: 

- Should I add a new type for custom operators? Currently I just used `Alias` (since it implements `Into Iden`), but using "alias" for a custom operator feels weird. Maybe a new `CustOp` type would make sense?
- I nuked the `Copy` trait from `BinOper`, but this has the downside of adding a clone call somewhere. This is my first time contributing to sea-query, and I'm rather ashamed to ask this: Is there an easy way to make `Ident` `Copy`? I tried adding the bound but I epicly failed at it.

src/types.rs
```diff
@@ -72,6 +72,8 @@ impl Clone for SeaRc<dyn Iden> {
     }
 }
 
+impl Eq for SeaRc<dyn Iden> {}
```
billy1624, line 75: Hmmmm... `Eq` for `dyn Iden`

Consider this. Where both `dyn Iden` yield the same identifier `id` but it's not the same Rust structure / enum.

```rs
#[derive(Iden)]
struct Id;

// Both yield the string "id"
assert_eq!(Id.to_string(), "id");
assert_eq!(Alias::new("id").to_string(), "id");

// How about this??
assert_eq!(Id, Alias::new("id"));
```

billy1624 (commented):
Hey @rakbladsvalsen, thanks for the contribution!! I got a few comments :)

src/types.rs
rakbladsvalsen: Isn't this already covered by the `PartialEq` implementation that already exists for `DynIden`?  The current implementation checks two things: whether the .to_string() contents are equal and whether the vtable pointer is the same for the two objects, though this only works for dyn objects.

Anyway, if we wanted to provide the ability to do something like the last line in your example, it'd be necessary to implement something like `fn get_iden_id()` to allow users direct comparison between Iden objects, i.e. `Id.get_iden_id() == Alias::new("id").get_iden_id()` (which is probably already doable using `to_string()`). 

Though I must say, this feels rather unnatural. I don't think comparison between two different struct types is something a lot of users are doing. There are no feature requests for this either, so I think we're good in that regard.

billy1624: @tyt2y3 thoughts?

src/types.rs
```diff
@@ -158,7 +160,7 @@ pub enum UnOper {
 }
 
 /// Binary operator
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq)]
```
tyt2y3, line 163: We should not remove the Copy attribute.

src/types.rs
```diff
@@ -185,6 +187,7 @@ pub enum BinOper {
     RShift,
     As,
     Escape,
+    Custom(DynIden),
```
tyt2y3, line 190: My question is why we would like to reuse DynIden here?
It was originally intended for 'identifiers'.
May be we can have a new type CustomOper, which is simply a wrapped `&'static str`

billy1624: Right, I was not aware we need to remove the `Copy` derive. So, yes, `&'static str` is a better alternative.

@rakbladsvalsen what do you think?

rakbladsvalsen: Thanks for the feedback guys! This PR _should_ be good to go.

billy1624 (approved):
Thanks!! @rakbladsvalsen

ikrivosheev (approved):
@rakbladsvalsen thank you! LGTM!

tyt2y3 (approved):
Awesome, looks neat!

PR #652, opened 2023-07-07
Upgrade to SQLx v0.7

billy1624:
## PR Info

- Dependents:
  - https://github.com/SeaQL/sea-orm/pull/1742

## Upgrade

- [ ] Upgrade `sqlx` to 0.7
- [ ] Upgrade `ipnetwork` to 0.20


billy1624:
Blocked by https://github.com/launchbadge/sqlx/issues/2549
CI Error: https://github.com/SeaQL/sea-query/actions/runs/5527364289/jobs/10083033939?pr=652#step:7:14

Waiting a patch release on SQLx side

saiintbrisson:
This is probably good to go now that launchbadge/sqlx#2585

tyt2y3:
Oh thanks, then may be we can revert fa50aa2375f61a833aed7586190a9170e1e25f2f

billy1624:
CHANGELOG of SQLx 0.7
https://github.com/launchbadge/sqlx/blob/main/CHANGELOG.md#071---2023-07-14

examples/sqlx_mysql/src/main.rs
```diff
@@ -4,6 +4,7 @@ use rust_decimal::Decimal;
 use sea_query::{ColumnDef, Expr, Func, Iden, MysqlQueryBuilder, OnConflict, Order, Query, Table};
 use sea_query_binder::SqlxBinder;
 use sqlx::{types::chrono::NaiveDateTime, MySqlPool, Row};
+use std::ops::DerefMut;
```
tyt2y3, line 7: can we remove this

examples/sqlx_mysql/src/main.rs
```diff
@@ -40,7 +41,7 @@ async fn main() {
         .col(ColumnDef::new(Character::Created).date_time())
         .build(MysqlQueryBuilder);
 
-    let result = sqlx::query(&sql).execute(&mut pool).await;
+    let result = sqlx::query(&sql).execute(pool.deref_mut()).await;
```
tyt2y3, line 44: And use the `(&mut *pool)` syntax?

rakshith-ravi: Isn't `.deref_mut()` more explicit?

The `(&mut *pool)` seems like too much magic, especially since this is in an example

examples/sqlx_mysql/src/main.rs
billy1624: We import `std::ops::DerefMut` into scope because we call the `.deref_mut()` method.

examples/sqlx_mysql/src/main.rs
billy1624: Personally, I think `.deref_mut()` looks nicer. And, yes, as @rakshith-ravi said, less magic and more explicit.

Originally, I use `(&mut *pool)` syntax, but then I switched to `.deref_mut()`. Both ways did the same thing. See https://github.com/SeaQL/sea-orm/pull/1742/commits/ee739a02632be166f0f0a56a66cbab8f9a3c0ef2, it's SeaORM but same.

```diff
- query.execute(conn).await.map(Into::into)
+ query.execute(&mut **conn).await.map(Into::into)
```

Isn't it look bad?

With `.deref_mut()`...

```diff
- query.execute(conn).await.map(Into::into)
+ query.execute(conn.deref_mut()).await.map(Into::into)
```

Thoughts? @tyt2y3

tyt2y3: `&mut *pool` +1

tyt2y3: > The Executor impls for Transaction and PoolConnection have been deleted because they cannot exist in the new crate architecture without rewriting the Executor trait entirely.
To fix this breakage, simply add a dereference where an impl Executor is expected, as they both dereference to the inner connection type which will still implement it:
&mut transaction -> &mut *transaction
&mut connection -> &mut *connection

As per https://github.com/launchbadge/sqlx/blob/main/CHANGELOG.md#breaking

tyt2y3 (approved):

sea-query-postgres/Cargo.toml
```diff
@@ -22,7 +22,7 @@ postgres-types = { version = "0.2", default-features = false }
 bytes = { version = "1", default-features = false }
 rust_decimal = { version = "1", default-features = false, optional = true }
 bigdecimal = { version = "0.3", default-features = false, optional = true }
-ipnetwork = { version = "0.19", default-features = false, optional = true }
+ipnetwork = { version = "0.20", default-features = false, optional = true }
```
tyt2y3, line 25: This is not intended?

PR #658, opened 2023-07-13
Diesel integration

Sytten:
## PR Info

See discussion: https://github.com/SeaQL/sea-query/discussions/168

**BLOCKED BY https://github.com/diesel-rs/diesel/pull/3691 and https://github.com/diesel-rs/diesel/pull/3694**

## New Features

- [x] Diesel binding

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -147,7 +147,10 @@ jobs:
       - run: cargo build --manifest-path sea-query-diesel/Cargo.toml --workspace --features postgres,sqlite,mysql --features=with-rust_decimal
       - run: cargo build --manifest-path sea-query-diesel/Cargo.toml --workspace --features postgres --features=with-rust_decimal-postgres
       - run: cargo build --manifest-path sea-query-diesel/Cargo.toml --workspace --features mysql --features=with-rust_decimal-mysql
-      - run: cargo build --manifest-path sea-query-diesel/Cargo.toml --workspace --features postgres,sqlite,mysql --features=with-bigdecimal
+      - run: |
+          cargo install --manifest-path sea-query-diesel/Cargo.toml --workspace --features postgres,sqlite,mysql --features=with-bigdecimal
+          cargo update --manifest-path sea-query-diesel/Cargo.toml --workspace -p bigdecimal:0.4.1 --precise 0.3.1
+          cargo build --manifest-path sea-query-diesel/Cargo.toml --workspace --features postgres,sqlite,mysql --features=with-bigdecimal
       - run: cargo build --manifest-path sea-query-diesel/Cargo.toml --workspace --features postgres,sqlite,mysql --features=with-uuid
       - run: cargo build --manifest-path sea-query-diesel/Cargo.toml --workspace --features postgres,sqlite,mysql --features=with-time
       - run: cargo build --manifest-path sea-query-diesel/Cargo.toml --workspace --features postgres,sqlite,mysql --features=with-ipnetwork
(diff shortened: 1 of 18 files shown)


Sytten:
@weiznich I would greatly appreciate if you could let me know if you think this is a good approach 🙏

Sytten:
@tyt2y3 @billy1624 This is ready for a first review, we still have to wait for the next diesel release before we can merge though

sea-query-diesel/src/query.rs
```diff
@@ -0,0 +1,49 @@
...
+        SeaQuery { query, binds }
+    }
+}
+
+impl<DB> QueryFragment<DB> for SeaQuery<DB>
+where
+    DB: Backend,
+{
+    fn walk_ast<'b>(&'b self, mut out: AstPass<'_, 'b, DB>) -> QueryResult<()> {
+        out.unsafe_to_cache_prepared();
```
weiznich, line 25: It might be useful to only call this method depending on whether sea-query believes that a prepared statement is meaningful to cache or not. It will significantly improve performance for queries that are executed more than once.

Diesel generally tries to cache all queries that are somewhat static, which means all queries that do not contain a user input dependent component (beside specific bind parameters).

weiznich (commented):
From the diesel side the implementation looks OK. The one thing I personally would try to improve is enabling prepared statement caching at least for some queries at it really helps with performance.

sea-query-diesel/src/query.rs
weiznich: It might be already worth to have a simple heuristics there: There are no bound paramaters -> cache the query, as it does not change for different calls.

For more targeted caching you would need to access the query ast somehow to check if there is for example no `IN` expression with a variable number of binds in there or no `INSERT` statement with a variable number of rows.

weiznich: Also there is quite a lot of detailed documentation of diesels query caching here: https://docs.diesel.rs/2.1.x/diesel/connection/statement_cache/index.html#a-primer-on-prepared-statement-caching-in-diesel

tyt2y3: Thanks for the info. I will keep this in mind, and later may be we can have a API that walks the AST and returns false if:

1) any custom expression used
2) any `in` and `not_in` clause
3) insert many

Sytten:
@tyt2y3 builds are failing because the bigdecimal versions dont match without a Cargo.lock, considering what I discussed in #664 are you OK with me adding a lock file to the examples and the crate?

tyt2y3:
> considering what I discussed in #664 are you OK with me adding a lock file to the examples and the crate?

Sure, I am okay with it

Sytten:
@tyt2y3 Ready for review

.github/workflows/rust.yml
```diff
@@ -135,6 +135,26 @@ jobs:
       - run: cargo build --manifest-path sea-query-postgres/Cargo.toml --workspace  --features=with-mac_address
       - run: cargo build --manifest-path sea-query-postgres/Cargo.toml --workspace  --features=postgres-array
 
+  diesel-build:
+    name: Build `sea-query-diesel`
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: dtolnay/rust-toolchain@stable
+      - run: cargo update --manifest-path sea-query-diesel/Cargo.toml --workspace -p bigdecimal:0.4.1 --precise 0.3.1
```
weiznich, line 144: This shouldn't be required anymore as diesel 2.1.1 added support for bigdecimal 0.4: https://crates.io/crates/diesel/2.1.1/dependencies

Sytten: Ha the issue is on sea-query side, they only support 0.3 and the resolver takes 0.4 for diesel so it causes a build failure because the types are incompatible. Its a PITA that we still don't have a way to pin a sub dependency without a lockfile.

tyt2y3 (approved):
Thank you

tyt2y3:
Released in https://crates.io/crates/sea-query-diesel/0.1.0

Sytten:
Thanks much appreciated!

Issue #674, opened 2023-08-11
Bug in logic for dropping parentheses


magbak:
## Description
From https://github.com/SeaQL/sea-query/blob/master/src/expr.rs:

```
pub(crate) fn need_parentheses(&self) -> bool {
        match self {
            Self::Binary(left, oper, _) => !matches!(
                (left.as_ref(), oper),
                (Self::Binary(_, BinOper::And, _), BinOper::And)
                    | (Self::Binary(_, BinOper::Or, _), BinOper::Or)
            ),
            _ => false,
        }
    }
```

This logic appears flawed, since the expression 
(("a" OR "b") OR "c") is determined to not need a parenthesis. 
But it clearly does some times:
((("a" OR "b") OR "c") AND "d")

In binary_expr() (a method of QueryBuilder), need_parenthesis()=false is sufficient to drop them, but SimpleExpr really does not have enough information to be able to decide that.

## Steps to Reproduce

```#[cfg(test)]
mod tests {
    use crate::backend::CommonSqlQueryBuilder;
    use crate::{Alias, BinOper, ColumnRef, DynIden, IntoIden, QueryBuilder, SimpleExpr};

    #[test]
    fn test_parentheses() {
        let qb = CommonSqlQueryBuilder {};
        let mut sql_string = "".to_string();
        let a = SimpleExpr::Column(ColumnRef::Column(Alias::new("a").into_iden()));
        let b = SimpleExpr::Column(ColumnRef::Column(Alias::new("b").into_iden()));
        let c = SimpleExpr::Column(ColumnRef::Column(Alias::new("c").into_iden()));
        let d = SimpleExpr::Column(ColumnRef::Column(Alias::new("d").into_iden()));

        let x_op_y = |x,op,y| SimpleExpr::Binary(Box::new(x), op, Box::new(y));
        let a_or_b = x_op_y(a,BinOper::Or, b);
        let a_or_b_or_c = x_op_y(a_or_b,BinOper::Or, c);
        let a_or_b_or_c_and_d = x_op_y(a_or_b_or_c.clone(),BinOper::And, d);
        println!("Needs parentheses: {}", a_or_b_or_c.need_parentheses());
        qb.prepare_simple_expr(&a_or_b_or_c_and_d, &mut sql_string);

        println!("Sql: {}", sql_string);
    }
}
```

Outputs: 
Needs parentheses: false
Sql: "a" OR "b" OR "c" AND "d"
[...]

PR #675, opened 2023-08-13
Issue #674: Fix issues with semantics of parenthesis removal

magbak:
## PR Info
This is my attempt at fixing #674 
- Closes [#674](https://github.com/SeaQL/sea-query/issues/674)

## Bug Fixes
Parenthesis placement is now done based on positive matches with known associativity and precedence rules, and otherwise defaults to placing parentheses. 

- [#674](https://github.com/SeaQL/sea-query/issues/674) identified issues where SQL parenthesis omission did not respect associativity and precedence rules. sea-query only removes parentheses now where it is certain that this respects such rules. 

## Changes
- Parentheses are now dropped from expressions such as 
``` (`character` LIKE 'D') AND (`character` LIKE 'E') ``` since LIKE has stronger precedence than the logical ops. Previously this happened only some times.

### DIFF
--- a/src/backend/query_builder.rs
+++ b/src/backend/query_builder.rs
@@ -1,4 +1,6 @@
+#[cfg(feature = "backend-postgres")]
 use crate::extension::postgres::PgBinOper;
+
 use crate::*;
 use std::ops::Deref;
 
@@ -298,8 +300,7 @@ pub trait QueryBuilder: QuotedBuilder + EscapeBuilder + TableRefBuilder {
                 self.prepare_un_oper(op, sql);
                 write!(sql, " ").unwrap();
                 let drop_expr_paren =
-                    simple_expr_greater_precedence_than_op(expr, &op.clone().into())
-                        .unwrap_or(false);
+                    inner_greater_precedence_than_outer(expr, &op.clone().into()).unwrap_or(false);
                 if !drop_expr_paren {
                     write!(sql, "(").unwrap();
                 }
@@ -1271,53 +1272,8 @@ pub trait QueryBuilder: QuotedBuilder + EscapeBuilder + TableRefBuilder {
     #[doc(hidden)]
     /// Translate part of a condition to part of a "WHERE" clause.
     fn prepare_condition_where(&self, condition: &Condition, sql: &mut dyn SqlWriter) {
-        if condition.negate {
-            write!(sql, "NOT ").unwrap()
-        }
-
-        if condition.is_empty() {
-            match condition.condition_type {
-                ConditionType::All => self.prepare_constant_true(sql),
-                ConditionType::Any => self.prepare_constant_false(sql),
-            }
-            return;
-        }
-
-        if condition.negate {
-            write!(sql, "(").unwrap();
-        }
-        condition.conditions.iter().fold(true, |first, cond| {
-            if !first {
-                match condition.condition_type {
-                    ConditionType::Any => write!(sql, " OR ").unwrap(),
-                    ConditionType::All => write!(sql, " AND ").unwrap(),
-                }
-            }
-            match cond {
-                ConditionExpression::Condition(c) => {
-                    if condition.conditions.len() > 1 {
-                        write!(sql, "(").unwrap();
-                    }
-                    self.prepare_condition_where(c, sql);
-                    if condition.conditions.len() > 1 {
-                        write!(sql, ")").unwrap();
-                    }
-                }
-                ConditionExpression::SimpleExpr(e) => {
-                    if condition.conditions.len() > 1 && (e.is_logical() || e.is_between()) {
-                        write!(sql, "(").unwrap();
-                    }
-                    self.prepare_simple_expr(e, sql);
-                    if condition.conditions.len() > 1 && (e.is_logical() || e.is_between()) {
-                        write!(sql, ")").unwrap();
-                    }
-                }
-            }
-            false
-        });
-        if condition.negate {
-            write!(sql, ")").unwrap();
-        }
+        let simple_expr = condition.to_simple_expr();
+        self.prepare_simple_expr(&simple_expr, sql);
     }
 
     #[doc(hidden)]
@@ -1390,21 +1346,21 @@ pub trait QueryBuilder: QuotedBuilder + EscapeBuilder + TableRefBuilder {
     ) {
         // If left has higher precedence than op, we can drop parentheses around left
         let drop_left_higher_precedence =
-            simple_expr_greater_precedence_than_op(left, &op.clone().into()).unwrap_or(false);
+            inner_greater_precedence_than_outer(left, &op.clone().into()).unwrap_or(false);
+
+        let known_assoc_op = matches!(
+            op,
+            BinOper::And | BinOper::Or | BinOper::Add | BinOper::Sub | BinOper::Mul | BinOper::Mod
+        );
+
+        #[cfg(feature = "backend-postgres")]
+        let assoc_op = known_assoc_op || matches!(op, BinOper::PgOperator(PgBinOper::Concatenate));
+
+        #[cfg(not(feature = "backend-postgres"))]
+        let assoc_op = known_assoc_op;
 
         //Figure out if left associativity rules allow us to drop the left parenthesis
-        let drop_left_assoc = left.is_binary()
-            && op == left.get_bin_oper().unwrap()
-            && matches!(
-                op,
-                BinOper::And
-                    | BinOper::Or
-                    | BinOper::Add
-                    | BinOper::Sub
-                    | BinOper::Mul
-                    | BinOper::Mod
-                    | BinOper::PgOperator(PgBinOper::Concatenate)
-            );
+        let drop_left_assoc = left.is_binary() && op == left.get_bin_oper().unwrap() && assoc_op;
 
         let left_paren = !drop_left_higher_precedence && !drop_left_assoc;
         if left_paren {
@@ -1421,7 +1377,7 @@ pub trait QueryBuilder: QuotedBuilder + EscapeBuilder + TableRefBuilder {
 
         // If right has higher precedence than op, we can drop parentheses around right
         let drop_right_higher_precedence =
-            simple_expr_greater_precedence_than_op(right, &op.clone().into()).unwrap_or(false);
+            inner_greater_precedence_than_outer(right, &op.clone().into()).unwrap_or(false);
 
         // Due to representation of trinary op between as nested binary ops.
         let drop_right_between_hack = (op == &BinOper::Between || op == &BinOper::NotBetween)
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

**Comment by tyt2y3** [comment]
Very cool! I'll look into it closely tomorrow

**Comment by magbak** [comment]
I am thinking that the (notational, not necessarily mathematical) (left) associativity of a binary operator could be encoded on the binary operator in types.rs instead as a function:

So that: 
```
 //Figure out if left associativity rules allow us to drop the left parenthesis
        let drop_left_assoc = left.is_binary()
            && op == left.get_bin_oper().unwrap()
            && (matches!(
                op,
                BinOper::And
                    | BinOper::Or
                    | BinOper::Add
                    | BinOper::Sub
                    | BinOper::Mul
                    | BinOper::Mod
                    | BinOper::PgOperator(PgBinOper::Concatenate)
            );
```

Becomes:
```
 //Figure out if left associativity rules allow us to drop the left parenthesis
        let drop_left_assoc = left.is_binary()
            && op == left.get_bin_oper().unwrap()
            && op.notationally_left_associative();
```
Where notationally_left_associative(&self) -> bool returns true if it is known to be notationally left associative in all the db backends (and people know about it so the parentheses would be distracting), and false otherwise, e.g. does not have notational associativity such as when chaining the comparison operators. 

Notational left associativity is mostly the case.

**Comment by magbak** [comment]
@tyt2y3: I will fix those failing checks once I know if the approach is in principle ok.

**Comment by tyt2y3** [comment]
Thank you. I am not sure I fully understood the notion of notational, so may be we can decide by examples?

e.g.
```sql
(("character" LIKE 'D') AND ("character" LIKE 'E')) # then
("character" LIKE 'D' AND "character" LIKE 'E') # now

A AND B AND C AND D # no change
```

I am trying to form a mental model for the existing behaviour, i.e. why did we want to have parens surrounding `LIKE`, i.e. `(("character" LIKE 'D') AND ("character" LIKE 'E'))`. And if it's possible, I'd like to minimize the changes if there is no semantic difference.

magbak:
> Thank you. I am not sure I fully understood the notion of notational, so may be we can decide by examples?
> 
> e.g.
> 
> ```sql
> (("character" LIKE 'D') AND ("character" LIKE 'E')) # then
> ("character" LIKE 'D' AND "character" LIKE 'E') # now
> 
> A AND B AND C AND D # no change
> ```
> 
> I am trying to form a mental model for the existing behaviour, i.e. why did we want to have parens surrounding `LIKE`, i.e. `(("character" LIKE 'D') AND ("character" LIKE 'E'))`. And if it's possible, I'd like to minimize the changes if there is no semantic difference.

Let me try to explain notational left associativity.
Notational left associativity means that for a given binary operator OP:
The SQL parser procesesses "a OP b OP c" the same as "(a OP b) OP c"
This means that when we encounter a BinaryExpr(BinaryExpr(a, OP, b)), OP, c) we can choose either of the representations above. But it only makes sense to do so when people are used to reading it without parentheses. The SQL parser will be equally happy. 

I am using the term _notational_ associativity, since there is also _mathematical_ associativity which is about the results of such expressions when they are evaluated when you move parentheses around. 

If we zoom out a bit, we see that in the whole expression, there is actually a bit of inconsistent behaviour:
Before:
```
"SELECT "character" FROM "character" WHERE ("character" LIKE 'C' OR (("character" LIKE 'D') AND ("character" LIKE 'E'))) AND (("character" LIKE 'F') OR ("character" LIKE 'G'))"
```

Now:
```
"SELECT "character" FROM "character" WHERE ("character" LIKE 'C' OR ("character" LIKE 'D' AND "character" LIKE 'E')) AND ("character" LIKE 'F' OR "character" LIKE 'G')"
```

The reason the parentheses are dropped around the binary LIKE-expressions in ``"character" LIKE 'D' AND "character" LIKE 'E'`` now is due to the relative precedence of the AND and LIKE operators. In all popular SQL engines, LIKE has higher precedence than the logical operators AND and OR. This means that LIKE finds the closest arguments before AND does so. Or in other words, AND has to wait until the LIKE expression is parsed, and has to relate to ```"character" LIKE 'D'``` as a whole. 
So that means that an SQL parser will process `("character" LIKE 'D') AND ("character" LIKE 'E')` in exactly the same way as `"character" LIKE 'D' AND "character" LIKE 'E'`. Again, we can choose between either representation when processing BinaryExpr(BinaryExpr("character", LIKE, 'D'), AND, BinaryExpr("character", LIKE, 'E')). But in my opinion we should only choose to drop the parentheses if people have an intuitive grasp of the relative precedence of the operators, such as with = having greater precedence than OR, AND and NOT, and of course where all SQL parsers agree to have a mostly uniform SQL string across databases. 

I can't quite see why it is justified in one case and not the other. After your question I decided to look into the technical reason why. The reason sea-query dropped the parentheses earlier was the fact that the first OR is encoded as a condition, and not as a binary OR. Having two code paths to writing the same kind of expression is not a good idea, since this opens the door to inconsistent behaviour which is hard to maintain over time (as is now becoming evident). IMHO, conditions, which are syntactic sugar should be rewritten as nested binary expressions before being written to the SQL string. Doing so can delete the whole function below, and more, which has subtle inconsistencies compared to how binary_expr works now and worked previously. I think this should be done before merging the PR. 

```
/// Translate part of a condition to part of a "WHERE" clause.
    fn prepare_condition_where(&self, condition: &Condition, sql: &mut dyn SqlWriter) {
        if condition.negate {
            write!(sql, "NOT ").unwrap()
        }

        if condition.is_empty() {
            match condition.condition_type {
                ConditionType::All => self.prepare_constant_true(sql),
                ConditionType::Any => self.prepare_constant_false(sql),
            }
            return;
        }

        if condition.negate {
            write!(sql, "(").unwrap();
        }
        condition.conditions.iter().fold(true, |first, cond| {
            if !first {
                match condition.condition_type {
                    ConditionType::Any => write!(sql, " OR ").unwrap(),
                    ConditionType::All => write!(sql, " AND ").unwrap(),
                }
            }
            match cond {
                ConditionExpression::Condition(c) => {
                    if condition.conditions.len() > 1 {
                        write!(sql, "(").unwrap();
                    }
                    self.prepare_condition_where(c, sql);
                    if condition.conditions.len() > 1 {
                        write!(sql, ")").unwrap();
                    }
                }
                ConditionExpression::SimpleExpr(e) => {
                    if condition.conditions.len() > 1 && (e.is_logical() || e.is_between()) {
                        write!(sql, "(").unwrap();
                    }
                    self.prepare_simple_expr(e, sql);
                    if condition.conditions.len() > 1 && (e.is_logical() || e.is_between()) {
                        write!(sql, ")").unwrap();
                    }
                }
            }
            false
        });
        if condition.negate {
            write!(sql, ")").unwrap();
        }
    }
```

tyt2y3:
Okay, then may be we make the changes first and see.

magbak:
in Cargo.toml:
sea-query = { git = "https://github.com/magbak/sea-query", branch = "parentheses_fix", default-features = false, features = ["thread-safe", "hashable-value", "backend-mysql", "backend-postgres", "backend-sqlite"] }

in sea-orm-codegen/Cargo.toml:
sea-query = { git = "https://github.com/magbak/sea-query", branch = "parentheses_fix", default-features = false, features = ["thread-safe"] }

All tests pass. Doing the revamp tomorrow morning. 
Will do a slight refactoring to get rid of the large closures / lambdas, which are not pretty.

magbak:
Hope this didn't break too much - have tuned it so as to break as little as possible.

src/backend/query_builder.rs
```diff
@@ -1378,29 +1344,58 @@ pub trait QueryBuilder: QuotedBuilder + EscapeBuilder + TableRefBuilder {
...
+        let known_assoc_op = matches!(
+            op,
+            BinOper::And | BinOper::Or | BinOper::Add | BinOper::Sub | BinOper::Mul | BinOper::Mod
+        );
+
+        #[cfg(feature = "backend-postgres")]
+        let assoc_op = known_assoc_op || matches!(op, BinOper::PgOperator(PgBinOper::Concatenate));
+
+        #[cfg(not(feature = "backend-postgres"))]
+        let assoc_op = known_assoc_op;
```
tyt2y3, line 1360: I think we can create a new trait method, have a default impl in this file, which is then overrided by the specific Postgres backend. Sort of like how `quote()` works.

src/backend/query_builder.rs
```diff
@@ -1261,53 +1272,8 @@ pub trait QueryBuilder: QuotedBuilder + EscapeBuilder + TableRefBuilder {
...
-                    }
-                }
-            }
-            false
-        });
-        if condition.negate {
-            write!(sql, ")").unwrap();
-        }
+        let simple_expr = condition.to_simple_expr();
+        self.prepare_simple_expr(&simple_expr, sql);
```
tyt2y3, line 1276: This is very cool if we managed to not break any Condition based test cases!

tyt2y3:
It seems to be very brilliant. Allow me more time to read through the logic in details.

src/backend/query_builder.rs
magbak: Will fix!

src/backend/query_builder.rs
magbak: There are some tests that had to be fixed, but there was no way around it unfortunately. 
I have had a go with sea-orm - and it appears to have no tests broken at least.

src/backend/query_builder.rs
magbak: Fixed now! Learned something new :-)

tyt2y3:
Nice refactoring!

tyt2y3:
This is indeed very brilliant! I was looking through the test cases, and trying to find an example that the old way was better. But so far, it seems like it made more sense the new way.
If the CI passes, I will invite niteave to have a look as well. NB: he contributed the `Condition` API.

tyt2y3:
@nitnelave we are working on a very interesting change that unify the Condition and BinOper query building backend. Can you take a look and review the changes?

magbak:
> This is indeed very brilliant! I was looking through the test cases, and trying to find an example that the old way was better. But so far, it seems like it made more sense the new way. If the CI passes, I will invite niteave to have a look as well. NB: he contributed the `Condition` API.

Great! Interested to hear what he thinks :-)

src/backend/query_builder.rs
```diff
@@ -1378,29 +1343,49 @@ pub trait QueryBuilder: QuotedBuilder + EscapeBuilder + TableRefBuilder {
...
-        let no_paren = matches!(op, BinOper::Equal | BinOper::NotEqual);
-        let left_paren = left.need_parentheses()
-            && left.is_binary()
-            && op != left.get_bin_oper().unwrap()
-            && !no_paren;
+        // If left has higher precedence than op, we can drop parentheses around left
+        let drop_left_higher_precedence =
+            self.inner_expr_well_known_greater_precedence(left, &(*op).into());
+
+        //Figure out if left associativity rules allow us to drop the left parenthesis
```
nitnelave, line 1350: Super-nit: Space at the beginning of the comment. Feel free to ignore.

src/backend/query_builder.rs
```diff
@@ -1378,29 +1343,49 @@ pub trait QueryBuilder: QuotedBuilder + EscapeBuilder + TableRefBuilder {
...
-            || right.is_binary() && op != left.get_bin_oper().unwrap())
-            && !no_right_paren
-            && !no_paren;
+
+        // If right has higher precedence than op, we can drop parentheses around right
+        let drop_right_higher_precedence =
+            self.inner_expr_well_known_greater_precedence(right, &(*op).into());
+
+        // Due to representation of trinary op between as nested binary ops.
+        let drop_right_between_hack = (op == &BinOper::Between || op == &BinOper::NotBetween)
```
nitnelave, line 1373: Reuse is_between

src/backend/query_builder.rs
```diff
@@ -1378,29 +1343,49 @@ pub trait QueryBuilder: QuotedBuilder + EscapeBuilder + TableRefBuilder {
...
+            && right.is_binary()
+            && right.get_bin_oper().unwrap() == &BinOper::And;
```
nitnelave, line 1375: Could we rewrite that as something like
matches!(right, Oper(BinOper::And))
To get rid of the unwrap

Same below

src/backend/query_builder.rs
```diff
@@ -1512,3 +1513,45 @@ impl QuotedBuilder for CommonSqlQueryBuilder {
 impl EscapeBuilder for CommonSqlQueryBuilder {}
 
 impl TableRefBuilder for CommonSqlQueryBuilder {}
+
+pub(crate) fn common_inner_expr_well_known_greater_precedence(
+    inner: &SimpleExpr,
+    outer_oper: &Oper,
+) -> bool {
+    match inner {
+        SimpleExpr::Column(_) => true,
```
nitnelave, line 1522: We could merge the trivial arms together, and maybe add why they are higher precedence.
I'm not sure I understand why e.g. FunctionCall is higher precedence, what does that mean in practice?

src/backend/query_builder.rs
```diff
@@ -1512,3 +1513,45 @@ impl QuotedBuilder for CommonSqlQueryBuilder {
...
+        SimpleExpr::Tuple(_) => true,
+        SimpleExpr::Constant(_) => true,
+        SimpleExpr::FunctionCall(_) => true,
+        SimpleExpr::AsEnum(_, _) => true,
+        SimpleExpr::Value(_) => true,
+        SimpleExpr::Keyword(_) => true,
+        SimpleExpr::SubQuery(_, _) => true,
+        SimpleExpr::Binary(_, inner_oper, _) => {
+            let inner_oper: Oper = (*inner_oper).into();
+            if inner_oper.is_arithmetic() || inner_oper.is_shift() {
```
nitnelave, line 1532: A common way to represent operator precedence is with integer levels, then you can do a lookup in a table and a straight comparison.

Would that approach work here? Or are there cases where it's not a total order? (E.g. lacking transitivity, maybe around the between operator?)

src/expr.rs
```diff
@@ -359,11 +359,11 @@ impl Expr {
     ///
     /// assert_eq!(
     ///     query.to_string(MysqlQueryBuilder),
-    ///     r#"SELECT `character`, `size_w`, `size_h` FROM `character` WHERE `id` = 1 AND 6 = 2 * 3"#
+    ///     r#"SELECT `character`, `size_w`, `size_h` FROM `character` WHERE `id` = 1 AND (6 = 2 * 3)"#
```
nitnelave, line 362: Is that a result we want? It feels to me that this was not an ambiguous case

src/expr.rs
```diff
@@ -2127,15 +2127,15 @@ impl SimpleExpr {
     ///
     /// assert_eq!(
     ///     query.to_string(MysqlQueryBuilder),
-    ///     r#"SELECT `character`, `size_w`, `size_h` FROM `character` WHERE ((`size_w` = 1) AND (`size_h` = 2)) OR ((`size_w` = 3) AND (`size_h` = 4))"#
+    ///     r#"SELECT `character`, `size_w`, `size_h` FROM `character` WHERE (`size_w` = 1 AND `size_h` = 2) OR (`size_w` = 3 AND `size_h` = 4)"#
```
nitnelave, line 2130: Nice win

src/query/condition.rs
```diff
@@ -260,6 +260,38 @@ impl Condition {
...
+        } else {
+            let mut inner_exprs = vec![];
+            for ce in &self.conditions {
+                inner_exprs.push(match ce {
+                    ConditionExpression::Condition(c) => c.to_simple_expr(),
+                    ConditionExpression::SimpleExpr(e) => e.clone(),
+                });
+            }
+            let mut inner_exprs_into_iter = inner_exprs.into_iter();
+            // Guaranteed to exist a next() by case distinction on conditions empty
```
nitnelave, line 279: We could change the .empty check with a match on next

nitnelave (approved):
Looks nice! A couple of comments here and there, but I like the precedence-based approach. There's just one case that I commented on that seems dubious, but overall it's a great change.

src/backend/query_builder.rs
magbak: Fixed :-)

src/backend/query_builder.rs
magbak: Yes, there is a partial ordering here. SQL parsers disagree about the finer details, and my goal was to represent those precedence relations that are well known, where adding unnecessary parentheses would create cluttered SQL. 

I thought just representing the relation directly would be easier to read and think about, and gave me an easier time to try to recreate most current behaviour.

src/expr.rs
magbak: This one could look strange, but it is done since it is in fact a custom expression:
    ///     .and_where(Expr::cust_with_values("6 = ? * ?", [2, 3]))
In general we should be careful and wrap these in parentheses since sea-query knows nothing about what kind of operator is contained.

src/backend/query_builder.rs
magbak: Fixed, also below with is_like()

src/backend/query_builder.rs
magbak: Changed to: 
&& matches!(right.get_bin_oper(), Some(&BinOper::And)); 
Also below with escape.

src/query/condition.rs
magbak: Done! Thanks for the tip :)

src/backend/query_builder.rs
magbak: Merged arms together and added this comment: 

// We only consider the case where an inner expression is contained in either a
// unary or binary expression (with an outer_oper).
// We do not need to wrap with parentheses:
// Columns, tuples (already wrapped), constants, function calls, values,
// keywords, subqueries (already wrapped)

I dropped SimpleEnum::AsEnum from this list, and the code.
In fact, it looks like there is a bug in sea-query wrt. how these are to be written to SQL?
This is from query_builder.rs:

            SimpleExpr::AsEnum(_, expr) => {
                self.prepare_simple_expr(expr, sql);
            }

The query builder for Postgres has different behaviour:

SimpleExpr::AsEnum(type_name, expr) => {
                let simple_expr = expr.clone().cast_as(SeaRc::clone(type_name));
                self.prepare_simple_expr_common(&simple_expr, sql);
            }

But we can follow that up in an issue.

src/expr.rs
tyt2y3: Thanks nitnelave for spotting this. For this particular case, I think a safer way is to have a special case, and always put quotes around `Custom` / `CustomWithExpr` operands.

magbak: @tyt2y3 this is in fact what I started to do and why the test case had to be changed in the first place :-)

tyt2y3:
Thank you @nitnelave for reviewing! 
@magbak Can you add a special case for Custom Expressions?
I think we are going to proceed to merge this into master, to be included in the next release.

tyt2y3:
I would say always have parentheses when combined with any operator. i.e.
`Expr::cust("1 + 1").add(Expr::value(2))` -> `(1 + 1) + 2`

For sole expression in where may be we can leave them naked?

`select()..and_where(Expr::cust("a = 1"))` -> `SELECT .. FROM .. WHERE a = 1`

`select()..and_where(Expr::cust("a = 1").or(Expr::value(1).eq(1)))` -> `SELECT .. FROM .. WHERE (a = 1) OR 1 = 1`

magbak:
> I would say always have parentheses when combined with any operator. i.e. `Expr::cust("1 + 1").add(Expr::value(2))` -> `(1 + 1) + 2`
> 
> For sole expression in where may be we can leave them naked?
> 
> `select()..and_where(Expr::cust("a = 1"))` -> `SELECT .. FROM .. WHERE a = 1`
> 
> `select()..and_where(Expr::cust("a = 1").or(Expr::value(1).eq(1)))` -> `SELECT .. FROM .. WHERE (a = 1) OR 1 = 1`

That is how it works now :-)

tyt2y3:
Oh right! I might have confused with the change set. Should be all good now

tyt2y3 (approved):

tyt2y3:
@prakol16 is your test case based on the latest code in this PR?

tyt2y3:
Merged into a new branch. @prakol16 can you open a PR on `parentheses_fix` and add your test there?

PR #680, opened 2023-08-21
fixed on conflict do nothing for MySQL

darkmmon:
## PR Info

- Closes 

- Dependencies:
  - 

- Dependents:
  - 

## New Features

- [ ] 

## Bug Fixes

- [ ]  fixed insert statement producing invalid statement for on_conflict do_nothing

## Breaking Changes

- [ ] 

## Changes

- [ ]

PR #682, opened 2023-08-22
On conflict do nothing for MySQL

tyt2y3:
Continuing #680

However I want to gather comments here whether `INSERT IGNORE` is a good substitution for ON CONFLICT DO NOTHING

It seems a better way would be to do `ON DUPLICATE KEY UPDATE id=id`, however this requires us to add an API to the frontend to allow users to pass in the primary key.

https://stackoverflow.com/questions/4596390/insert-on-duplicate-key-do-nothing

### DIFF
--- a/src/backend/mysql/query.rs
+++ b/src/backend/mysql/query.rs
@@ -30,6 +30,24 @@ impl QueryBuilder for MysqlQueryBuilder {
         // MySQL doesn't support declaring materialization in SQL for with query.
     }
 
+    fn prepare_insert(
+        &self,
+        replace: bool,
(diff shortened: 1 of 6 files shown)


tyt2y3:
I think `INSERT IGNORE` is an inappropriate solution to `ON CONFLICT DO NOTHING`.

I think we should add a special method for MySQL that would result in `ON DUPLICATE KEY UPDATE id=id`.

In SeaORM we have the PK information.

May be `OnConflictAction::DoNothing` can be extended to `DoNothing(Option<DynIden>)` to hold the PK?

PR #684, opened 2023-08-24
On conflict do nothing for MySql 

darkmmon:
## PR Info

- Partially closes https://github.com/SeaQL/sea-orm/issues/1790

- Dependencies:
  - 

- Dependents:
  - 

## New Features

- [ ]  added do_nothing_mysql([pk_cols]). This is to replace do_nothing() with MySql, require user to input the primary key columns manually.

## Bug Fixes

- [ ]  fixed a bug where do_nothing() will return syntax error with MySql. See New Features for the new function.

## Breaking Changes

- [ ] 

## Changes

- [ ]

### DIFF
--- a/src/query/on_conflict.rs
+++ b/src/query/on_conflict.rs
@@ -134,12 +134,14 @@ impl OnConflict {
     ///     ]
     ///     .join(" ")
     /// );
-    pub fn do_nothing_mysql<C, I>(&mut self, pk_cols: I) -> &mut Self 
+    pub fn do_nothing_mysql<C, I>(&mut self, pk_cols: I) -> &mut Self
     where
         C: IntoIden,
         I: IntoIterator<Item = C>,
     {
-        self.action = Some(OnConflictAction::DoNothing(pk_cols.into_iter().map(IntoIden::into_iden).collect()));
+        self.action = Some(OnConflictAction::DoNothing(
+            pk_cols.into_iter().map(IntoIden::into_iden).collect(),
+        ));
         self
     }
 
--- a/src/query/on_conflict.rs
+++ b/src/query/on_conflict.rs
@@ -19,7 +19,7 @@ pub enum OnConflictTarget {
 #[derive(Debug, Clone, PartialEq)]
 pub enum OnConflictAction {
     /// Do nothing
-    DoNothing,
+    DoNothing(Vec<DynIden>),
     /// Update column value of existing row
     Update(Vec<OnConflictUpdate>),
 }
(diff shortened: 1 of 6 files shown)


tests/mysql/query.rs
```diff
@@ -1141,6 +1141,32 @@ fn insert_from_select() {
...
+            .on_conflict(
+                OnConflict::new()
+                    .do_nothing_mysql(vec![Glyph::Id])
+                    .to_owned()
+            )
+            .to_string(MysqlQueryBuilder),
+        [
+            r#"INSERT INTO `glyph` (`aspect`, `image`)"#,
+            r#"VALUES ('04108048005887010020060000204E0180400400', 3.1415)"#,
+            r#"ON DUPLICATE KEY UPDATE `id` = VALUES(`id`)"#,
```
tyt2y3, line 1164: Can you make it `id` = `id`?

src/query/on_conflict.rs
```diff
@@ -64,8 +64,84 @@ impl OnConflict {
...
+    /// assert_eq!(
+    ///     query.to_string(MysqlQueryBuilder),
+    ///     [
+    ///         r#"INSERT INTO `glyph` (`aspect`, `image`)"#,
+    ///         r#"VALUES ('abcd', 3.1415)"#,
+    ///         r#"ON DUPLICATE KEY UPDATE `id` = VALUES(`id`)"#,
+    ///     ]
+    ///     .join(" ")
+    /// );
+    pub fn do_nothing_mysql<C, I>(&mut self, pk_cols: I) -> &mut Self
```
ikrivosheev, line 137: @darkmmon  I'm not sure, that naming is good... 

@tyt2y3 what do you think?

tyt2y3: Yes, may be `do_nothing_on` and add `MySQL only` in the doc? https://github.com/SeaQL/sea-query/blob/336bf954b94c618daef0b5c15a0dbf65da12ea25/src/table/column.rs#L384

ikrivosheev: I agree with it

tests/mysql/query.rs
```diff
@@ -1161,7 +1183,7 @@ fn insert_on_conflict_0() {
         [
             r#"INSERT INTO `glyph` (`aspect`, `image`)"#,
             r#"VALUES ('04108048005887010020060000204E0180400400', 3.1415)"#,
-            r#"ON DUPLICATE KEY UPDATE `aspect` = VALUES(`aspect`), `image` = VALUES(`image`)"#,
+            r#"ON DUPLICATE KEY UPDATE `aspect` = `aspect`, `image` = `image`"#,
```
tyt2y3, line 1186: Oh sad, this is wrong. `ON DUPLICATE KEY UPDATE id = id` should only applies to `do_nothing_on`.
While the behaviour of `update_columns()` should not be changed.
The two expressions have **different semantics**.

tyt2y3 (changes requested):
As stated

PR #687, opened 2023-08-29
fix: misbehavior in FunctionCall when calling arg multiple times

mohs8421:
## Bug Fixes

- misbehaviour in `FunctionCall` when calling `.arg()` multiple times, it always overrides the previous values instead of appending them.


ikrivosheev (approved):
@mohs8421 good catch! Thank you, LGTM!

tyt2y3 (approved):

PR #692, opened 2023-08-31
Support adding expressions for  ON CONFLICT targets

### LINKED ISSUE #678: OnConflict does not support adding a function expression to list of columns
## Description
I was trying to add a function expression to the column list in an `OnConflict` statement. But this does not seem to be possible. Only objects that implement `IntoIden` are acceptable in the `OnConflict::columns` or the `OnConflict::column` method.

Postgres supports adding function expressions in the column list on the `ON CONFLICT` clause on insert statements. This functionality seems to be missing from sea query.

How do I achieve this? Is there way to do this that I might have overlooked?

```sql
INSERT INTO posts (name, identifier, content)
VALUES ('Getting Started', 'getting-started', 'Some sample content')
ON CONFLICT (lower(identifier)) -- We want to add a function expression to the column list here
DO NOTHING;
```

### Expected Behavior

```rs
let mut statement = Query::insert()
    .into_table(PostSchema::Table)
    .columns([PostSchema::Name, PostSchema::Identifier, PostSchema::Content])
    .values(["Getting Started", "getting-started", "Some sample content"])
    .on_conflict(
        OnConflict::column(
            Expr::expr(Func::lower(PostSchema::Identifier)) // This does not seem to be possible as SimpleExpr does not implement `IntoIden`
        )
        .do_nothing()
        .to_owned(),
    );
```
Adding an `Alias` to the column does not seem to be working correctly.

```rs
OnConflict::column(Alias::new("lower(identifier)"));
```
Postgres returns `"lower(identifier)"` is not a valid column name. It is due to the `"` around the column name when it's converted to an sql statement.

### Actual Behavior

Compile time error:
```rs
error[E0277]: the trait bound `sea_query::Expr: Iden` is not satisfied
   --> src/main.rs
   |
   |       OnConflict::column(
   |       ------------------ required by a bound introduced by this call
...
   |       Expr::expr(Func::lower(Expr::col(PostSchema::Identifier))),
   |       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `Iden` is not implemented for `sea_query::Expr`
```

## Versions

```
cargo tree | grep sea-
    │   ├── sea-query v0.29.0 (https://github.com/SeaQL/sea-query.git?branch=master#1c9a6656)
    │   │   ├── sea-query-derive v0.3.0 (proc-macro) (https://github.com/SeaQL/sea-query.git?branch=master#1c9a6656)
    ├── sea-query v0.29.0 (https://github.com/SeaQL/sea-query.git?branch=master#1c9a6656) (*)
    ├── sea-query-binder v0.4.0 (https://github.com/SeaQL/sea-query.git?branch=master#1c9a6656)
    │   ├── sea-query v0.29.0 (https://github.com/SeaQL/sea-query.git?branch=master#1c9a6656) (*)
```
OS: WSL2 running Ubuntu 20.04.6 LTS

## Suggestions
We could add a `.expr()` method on `OnConflict` just like `Query::select().expr()` that let's us pass custom expressions.
```rs
let conflict = OnConflict::columns([ /* ... */ ]).expr(Func::lower(Expr::col(PostSchema::Identifier)));
```
I could take a stab at it if you are interested in adding this functionality. I would be interested in knowing if this was overlooked or was never planned to be added due to a specific reason.

Thanks again for this awesome library! I have been really enjoying working with it.

sumeetattree:
## PR Info

Supports adding expressions to `ON CONFLICT` targets on supported databases (postgres, and sqlite);
```sql
INSERT ... ON CONFLICT ("id", LOWER("email")) DO UPDATE SET "name" = "excluded"."name"
```
- Closes #678 

## New Features

- [x] Added an `OnConflict::expr()` and `OnConflict::exprs()` impls to support adding custom expressions to on_conflict targets

## Breaking Changes

None

## Changes

None

### DIFF
--- a/src/backend/mysql/query.rs
+++ b/src/backend/mysql/query.rs
@@ -59,7 +59,7 @@ impl QueryBuilder for MysqlQueryBuilder {
         sql.push_param(value.clone(), self as _);
     }
 
-    fn prepare_on_conflict_target(&self, _: &Option<OnConflictTarget>, _: &mut dyn SqlWriter) {
+    fn prepare_on_conflict_target(&self, _: &[OnConflictTarget], _: &mut dyn SqlWriter) {
         // MySQL doesn't support declaring ON CONFLICT target.
     }
 
--- a/src/backend/query_builder.rs
+++ b/src/backend/query_builder.rs
@@ -1108,7 +1108,7 @@ pub trait QueryBuilder:
     fn prepare_on_conflict(&self, on_conflict: &Option<OnConflict>, sql: &mut dyn SqlWriter) {
         if let Some(on_conflict) = on_conflict {
             self.prepare_on_conflict_keywords(sql);
-            self.prepare_on_conflict_target(&on_conflict.target, sql);
+            self.prepare_on_conflict_target(&on_conflict.targets, sql);
             self.prepare_on_conflict_condition(&on_conflict.target_where, sql);
             self.prepare_on_conflict_action(&on_conflict.action, sql);
             self.prepare_on_conflict_condition(&on_conflict.action_where, sql);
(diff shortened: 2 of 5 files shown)


sumeetattree:
There is little weirdness in how `OnConflict` struct is initialized:
```rs
let conflict_column = OnConflict::column(Glyph::Id).to_owned(); // no need to call `new` first

let conflict_expr = OnConflict::new().expr(Func::lower(Expr::col(Glyph::Tokens))).to_owned();
```
`.column()` method acts like a constructor but `.expr()` requires the object to be already constructed. Changing `.expr()` method to also act like a constructor locks us out of using both columns and expressions in the same `ON CONFLICT` clause.

```rs
impl OnConflict {
    pub fn column(col) -> Self {
        // ...
    }

    pub fn expr(expr) -> Self {
        // ...
    }
}
let conflict = OnConflict::column(Glyph::Id)
    .expr(Func::lower(Expr::col(Glyph::Tokens)))  // will NOT work!
    .to_string();
```

On the flip side, using `.expr()` alone is a little cumbersome because we must first call `.new()`.


However, doing this:
```rs
impl OnConflict {
    pub fn column(&mut self, col) -> &mut Self {
        // ...
    }

    pub fn expr(&mut self, expr) -> &mut Self {
        // ...
    }
}
```
is a breaking change for all `.column()` and `.columns()` consumers.

How do we reconcile these api differences?

sumeetattree:
@ikrivosheev I have tried fixing the formatting issues raised by the workflow. I think it's sorted. Can you take another look?

tyt2y3 (approved):
Sorry for the delay

PR #697, opened 2023-09-07
Fix incorrect behavior when adding an autoincrement column to a Postg…

opusbopus:
## Bug Fixes

Using the `PostgresQueryBuilder` I would've expected this:
```rs
Table::alter()
	.table(Test::Table)
	.add_column(
		ColumnDef::new(Test::Id)
			.integer()
			.auto_increment()
			.primary_key(),
	)
```

to give me: (This behavior is true if you use `Table::create`)
```ALTER TABLE "test" ADD COLUMN "id" serial PRIMARY KEY```

Instead it gave me:
```ALTER TABLE "test" ADD COLUMN "id" integer PRIMARY KEY```

## Breaking Changes

If someone was using `auto_increment()` on anything but an `integer`, `smallinteger`, or `biginteger` it would cause a panic.

## Changes

Modified the postgres `alter_10` test to test against the correct behavior.

### DIFF
--- a/src/backend/postgres/table.rs
+++ b/src/backend/postgres/table.rs
@@ -112,7 +112,11 @@ impl TableBuilder for PostgresQueryBuilder {
                     let f = |column_def: &ColumnDef, sql: &mut dyn SqlWriter| {
                         if let Some(column_type) = &column_def.types {
                             write!(sql, " ").unwrap();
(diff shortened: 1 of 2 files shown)


opusbopus:
of course I had to forget rustfmt, sorry ^^

tyt2y3 (approved):
Thank you for your contribution

PR #698, opened 2023-09-09
feat: setup `rust-toolchain.toml` to use nightly by default

EstebanBorai:
Introduces `rust-toolchain.toml` which forces `cargo` to use latest Nightly.
This is useful for users using `stable` as the default toolchain, Rustup
responds to the presence of this file as follows:

```
➜  sea-query git:(master) rustup default stable
info: using existing install for 'stable-aarch64-apple-darwin'
info: default toolchain set to 'stable-aarch64-apple-darwin'

  stable-aarch64-apple-darwin unchanged - rustc 1.72.0 (5680fa18f 2023-08-23)

info: note that the toolchain 'nightly-aarch64-apple-darwin' is currently in use (overridden by '~/Repos/sea-query/rust-toolchain.toml')
```

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,0 +1,2 @@
+[toolchain]
+channel = "nightly"


EstebanBorai:
@ikrivosheev , @billy1624 not sure why we use `stable` for clippy and `nightly` for `rustfmt`?

https://github.com/SeaQL/sea-query/blob/master/.github/workflows/rust.yml#L27-L61

tyt2y3:
I think our default support target is always the latest stable.
nightly... we only needed it for cargo fmt, specifically it fmt the code examples in doc test.

EstebanBorai:
Gotcha! Will se if that can be configured via the toolchain file!

EstebanBorai:
Hi @tyt2y3! After reading through `rust-toolchain.toml` docs and also seeking for more references, I ended up without any approach to have a Rust Toolchain setup that uses `rustfmt` on Nightly but Stable on other commands.

Im closing this to keep the inbox clear!

Issue #689, opened 2023-08-30
`BIGINT PRIMARY KEY AUTOINCREMENT` in SQLite


tyt2y3:
https://sqlite.org/forum/info/2dfa968a702e1506e885cb06d92157d492108b22bf39459506ab9f7125bca7fd

SQLite would give `error: "AUTOINCREMENT is only allowed on an INTEGER PRIMARY KEY"`

So I guess we somehow have to make `BIGINT PRIMARY KEY AUTOINCREMENT` a special case and convert it implicitly to `INTEGER PRIMARY KEY AUTOINCREMENT`

Ref: https://github.com/SeaQL/sea-orm/issues/1832

Perksey:
Yes, just hit this.

```
thread 'backend::tests::sqlite_playground' panicked at 'called `Result::unwrap()` on an `Err` value: SQLite(SqliteFailure(Error { code: Unknown, extended_code: 1 }, Some("AUTOINCREMENT is only allowed on an INTEGER PRIMARY KEY")))', src/backend/mod.rs:288:55
```

For generated query

```sql
CREATE TABLE IF NOT EXISTS "example_table" ( "id" bigint PRIMARY KEY AUTOINCREMENT, "my_field1" integer NOT NULL, "my_field2" bigint )
```

I would personally suggest that sea-query maps the backend's type affinity exactly i.e. always mapping the sized integer types to just `INTEGER`, mapping both float and double to `REAL`, etc. Really sea-query is an abstraction over multiple SQL languages, and while it's very useful to the user to hint sized types for backends that do support more rigid type affinity, its usefulness wears off when the abstraction tries to preserve this rigidity for a database/query language that fundamentally isn't modelled that way and does not support/do anything with the information it is being given.

I'm happy to do the work in a PR if this is agreeable.

tyt2y3:
A PR on the error above would be appreciated.
However I am not sure I understand your preference to type mapping. Or I think we are having the opposite. But only because we wanted to preserve the *intended type*, such that we when doing schema discovery, we'd get back `i32` vs `i64`. But that's more an ORM thing.

tyt2y3:
```rust
pub struct SqliteQueryBuilder<TTypeMapper: SqliteTypeMapper = ExactMapper>;
```

This is a great idea

tyt2y3:
Fixed via 37a30f6cc8dcd9a61c80bbc6ff8ff249831e2129

PR #708, opened 2023-09-25
Add SqliteTypedQueryBuilder for exact type mapping

Perksey:
## PR Info

- Related to (doesn’t necessarily close) #689 

## New Features

Adds customizable type mapping to ensure that the exact type can be used on SQLite if the user needs it. 

## Bug Fixes

Provides the ability for the user to enable BigUnsigned PRIMARY KEY AUTOINCREMENT.

## Breaking Changes

I’ve done my best to avoid any breaking changes. This is done by introducing a SqliteTypedQueryBuilder and implementing a SqliteBuilderVariant trait on both builders to deduplicate the code.

## Changes

I was hesitant to add a blanket implementation of sea-query types on T where T is a SqliteBuilderVariant, as this would definitely cause future problems, either here or downstream. Therefore, I’ve implemented a sqlite_impl macro which will repeat the implementation for both SqliteTypedQueryBuilder<T> and SqliteQueryBuilder.

I forgot about the need of PhantomData when I originally proposed this, and this can’t be done in a non-breaking manor without separate types.


Perksey:
This can close #689 but discussion is needed for whether we actually want to special case BIGINT AUTOINCREMENT on the default mapper.

Perksey:
@tyt2y3 ready for review

src/backend/sqlite/table.rs
```diff
@@ -1,48 +1,168 @@
...
 
-    fn prepare_column_type(&self, column_type: &ColumnType, sql: &mut dyn SqlWriter) {
+impl StaticTypeMapper for DefaultTypeMapper {
+    fn prepare_column_type(column_type: &ColumnType, sql: &mut dyn SqlWriter) {
         write!(
             sql,
             "{}",
             match column_type {
+                // ----------------------------------------------------------------------------------------------------
+                // A NOTE REGRADING SQLITE TYPE AFFINITY: It's great that we try to be very on-point with our mappings,
```
tyt2y3, line 136: That's a very throughout comment section haha

tyt2y3 (commented):
It looks pretty good.

My only hope is to have a prettier macro that can do the following:

```rust
#[impl_many]
impl T for A, B {
    fn method() {}
}
```

that would create no doubt about what the `sqlite_impl` would do

Perksey:
I couldn't find any macros that would support adding a generic parameter for one of the alternate selves, and attempting to make one did not go well!

tyt2y3:
Alternatively, may be we can implement only a subset of the traits?

PR #716, opened 2023-10-18
Add from_clear to allow emptying current from tables in select statement

dragonnn:
## PR Info
Sea-orm does allow to modify the underlying sea-query statements with can be really useful but sea-query itself doesn't have a way to change the table a `SelectStatement` is done. This can be really handy when a database has a few tables that do share the same columns but are separated for other reasons, in my use case I have tables like that: data, data_now, data_1s, data_1m. All of them are identical to each other (data_now does have different PK keys, but that isn't important in my case), with using from_clear I can avoid having separate Entities for each table and avoid a lot of boilerplate when mapping them to a single Grpc struct that my backend outputs.

## New Features

- add new function to `SelectStatement`  `from_clear()` with does remove all from statements from it

## Changes

- new function `from_clear()` added to `SelectStatement`

Let me know if anything more is need in this PR or if you have a better idea to archive what I need for my use case. Or a better name for `from_clear()` because I am myself not that happy with it.

### DIFF
--- a/src/query/select.rs
+++ b/src/query/select.rs
@@ -987,6 +987,38 @@ impl SelectStatement {
         self.from_from(TableRef::FunctionCall(func, alias.into_iden()))
     }
 
+    /// Clears all current from clauses.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use sea_query::{tests_cfg::*, *};
+    ///
+    /// let query = Query::select()
+    ///     .column(ColumnRef::Asterisk)
+    ///     .from(Char::Table)
+    ///     .from_clear()
+    ///     .from(Font::Table)
+    ///     .to_owned();
+    ///
+    /// assert_eq!(
+    ///     query.to_string(MysqlQueryBuilder),
+    ///     r#"SELECT * FROM `font`"#
+    /// );
+    /// assert_eq!(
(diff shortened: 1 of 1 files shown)


tyt2y3:
Just being curious, why not build the same query with different `Table` passed in as parameters?

dragonnn:
> Just being curious, why not build the same query with different `Table` passed in as parameters?

Because I am using sea-orm, so I don't build the query myself in the first place. Here is a sample of the code where I am using it:
```rust
    pub async fn get_now_by_device_uuid(device_uuid: Uuid, orm: &impl ConnectionTrait) -> DbResult<Option<Self>> {
        let mut query = Entity::find().filter(Column::DeviceUuid.eq(device_uuid));
        sea_orm::QueryTrait::query(&mut query).clear_from().from_as(super::data_now::Entity.table_ref(), Entity);
        query.one(orm).await
    }
```

Doing the whole query would be for sure more involving as getting the existing sea-orm query and just swapping the table out.

dragonnn:
If no one has any comments to it I would like to convert from draft to ready :). I would like really to get it merged so I don't have to use my fork in my projects.

tyt2y3 (approved):
Sorry for the delay

PR #729, opened 2023-12-29
refactor(sea-query): Fix clippy warnings on Rust 1.75

siketyan:
## PR Info

- Closes N/A
- Dependencies: N/A
- Dependents: N/A

## New Features

N/A

## Bug Fixes

N/A

## Breaking Changes

N/A

## Changes

- [x] Fix clippy warnings introduced in Rust 1.75 toolchain

### DIFF
--- a/src/extension/postgres/mod.rs
+++ b/src/extension/postgres/mod.rs
@@ -1,8 +1,6 @@
 pub use expr::*;
 pub use extension::*;
 pub use func::*;
-#[allow(unused_imports)]
-pub use interval::*;
 pub use ltree::*;
 pub use types::*;
 
--- a/src/extension/postgres/mod.rs
+++ b/src/extension/postgres/mod.rs
@@ -1,6 +1,7 @@
 pub use expr::*;
 pub use extension::*;
 pub use func::*;
+#[allow(unused_imports)]
 pub use interval::*;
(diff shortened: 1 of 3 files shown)


src/extension/postgres/mod.rs
```diff
@@ -1,6 +1,7 @@
 pub use expr::*;
 pub use extension::*;
 pub use func::*;
+#[allow(unused_imports)]
```
siketyan, line 4: There are no exported members in interval module so clippy warns it, should we just remove this re-export or keep with `#[allow]` for future usages?

tyt2y3: If that's the case it's good to remove `pub` from `pub use`

tyt2y3:
Thank you! You are so fast

src/extension/postgres/mod.rs
siketyan: It seems not used in the module, too.

image

tyt2y3: You are right, all code in the `interval.rs` is impl so we can remove this `use`.

siketyan: ok, removed entire the re-export in b94ea89

tyt2y3 (commented):
Thank you

PR #735, opened 2024-01-05
Remap SQLite types

tyt2y3:
https://www.sqlite.org/datatype3.html


tyt2y3:
This is breaking, upon merge we need to bump to 0.31

tyt2y3:
We forgotten `option-sqlite-exact-column-type`

PR #750, opened 2024-02-01
mysql: drop `year` display length

billy1624:
## Breaking Changes

- [x] mysql: drop `year` display length


tyt2y3:
Please update the changelog thanks.

PR #759, opened 2024-03-09
Add possibility to configure the Table enum variant Iden implementation with enum_def macro

Gaelik-git:
Hello ! 

Add possibility to configure the Table enum variant Iden implementation with enum_def macro

Do you think it would possible to merge the PR ? 
Does the feature fit in what you are trying to do ?

### DIFF
--- a/sea-query-attr/tests/pass/table_iden.rs
+++ b/sea-query-attr/tests/pass/table_iden.rs
@@ -1,0 +1,11 @@
+use sea_query::Iden;
+use sea_query_attr::enum_def;
+
+#[enum_def(table_iden = "HelloTable")]
+pub struct Hello {
+    pub name: String,
+}
+
+fn main() {
+    println!("{:?}", HelloIden::Table.to_string());
+}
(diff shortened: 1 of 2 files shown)


sea-query-attr/tests/pass/table_iden.rs
```diff
@@ -0,0 +1,11 @@
+use sea_query::Iden;
+use sea_query_attr::enum_def;
+
+#[enum_def(table_iden = "HelloTable")]
+pub struct Hello {
+    pub name: String,
+}
+
+fn main() {
+    println!("{:?}", HelloIden::Table.to_string());
```
tyt2y3, line 10: Thank you! Can you convert this to `assert_eq`?

Gaelik-git: Sure !

tyt2y3 (approved):

tyt2y3:
I changed `table_iden` to `table_name`

tyt2y3:
Released https://crates.io/crates/sea-query-attr/versions

tyt2y3:
As mentioned in another comment, I would like to deprecate `sea-query-attr` and move the macro into `sea-query-derive`. However, we'll have to migrate to `syn2` and probably rename a few things. Would you like to do that?

Issue #706, opened 2023-09-16
syn 1.x and 2.x both used in sea-query


sartak:
## Description

I'm shipping a crate that uses sea-query, but getting a warning that multiple versions of `syn` are being pulled in.

I'm sure it wouldn't be trivial to upgrade `sea-query-derive` to `syn` 2.x, but it would make downstream compile times faster, (possibly) binaries smaller, and reduce warnings that are not feasibly fixable.

## Steps to Reproduce

1. `cargo new --lib foo`
2. `cd foo`
3. add `#![warn(clippy::cargo)]` to the top of `src/lib.rs`
4. `cargo add sea-query`
5. `cargo clippy`

### Expected Behavior

Adding `sea-query` to my project is `clippy::cargo`-clean.

### Actual Behavior

Adding `sea-query` to my project generates a new `clippy::cargo` warning;

```
warning: multiple versions for dependency `syn`: 1.0.109, 2.0.35
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#multiple_crate_versions
```

### Reproduces How Often

Every time.

## Versions

```
λ cargo tree | grep sea-
└── sea-query v0.30.1
    └── sea-query-derive v0.4.0 (proc-macro)
```

## Additional Information

`sea-query-derive` depends on `syn` 1.x here: https://github.com/SeaQL/sea-query/blob/b80f14eaa1d4b02389db033ad64953fbc8c925d6/sea-query-derive/Cargo.toml#L18
where `syn` 2.x is pulled in via `inherent` and `thiserror`.

tyt2y3:
Thanks for raising this.

Yeah, SeaORM went through an exercise of upgrading to syn2. It was an involving process, and it left us just there without touching SeaQuery. Contribution is of course welcome.

However, I would like to point out this is an ecosystem problem. Unless *every crate* in a dependency tree upgrades to syn2, it will be pretty hard to avoid have two `syn`.

opusbopus:
I wouldn't mind taking a look at this in my spare time. Need more proc macro practice anyway

PR #763, opened 2024-03-31
Replace derivative by educe

negezor:
## PR Info

Digging into my `Cargo.toml` to get rid of syn v1, I saw that sea-query still depends indirectly through the `derivative` module. It hasn't been updated for a long time https://github.com/mcarton/rust-derivative/issues/117. I found a more compact alternative to `educe`.

- Closes https://github.com/SeaQL/sea-query/issues/706 (It's already closed, but I continued working)

## Changes

- [X] Replace `derivative` derive by `educe` for `Value`


tyt2y3:
Thanks!

tyt2y3 (approved):
This is indeed a good solution

PR #766, opened 2024-04-10
Pin `time` to version 0.3.34

billy1624:
## PR Info

- Dependents:
  - https://github.com/SeaQL/sea-query/pull/765

## Changes

- [x] Pin `time` to version 0.3.34, because the breaking changes released in version 0.3.35 isn't handled in our upstream crates such as `sqlx` https://github.com/launchbadge/sqlx/issues/3189.


RWDai:
Waiting for the merge and release, otherwise it will affect our current engineering code

tyt2y3:
If we release now, we'll be broken again when sqlx makes a new release. Or may be we also pin the sqlx version.

tyt2y3:
I've done the changes on master

PR #767, opened 2024-04-11
Added the ability to rename the table for structs while using enum_def

louiseyousre:
## New Features

- Now you can rename the table for structs while using enum_def


tyt2y3:
Very interesting, this is a duplicate of https://github.com/SeaQL/sea-query/pull/759 I'll merge that one, sorry

tyt2y3:
Actually, there is one task. I would like to deprecate `sea-query-attr` and move the macro into `sea-query-derive`. However, we'll have to migrate to `syn2`. Would you have the time to do that?

Issue #725, opened 2023-12-18
postgres: `CAST($2 AS CaseSensitiveEnumType)` enum type must be enquoted


niklaskorz:
## Description

When a model contains a field with a case-sensitive enum type name, querying the table crashes because the type name is not enquoted in the `CAST`.

## Steps to Reproduce

1. Have a model field with a case-sensitive enum type name
2. Query the model for rows with a specific value for this enum field

### Expected Behavior

The query should not crash.

### Actual Behavior

```
2023-12-18T11:15:22.411880Z  INFO JobScheduler::recover_jobs: sqlx::query: summary="SELECT \"KjuJob\".\"id\", \"KjuJob\".\"type\", \"KjuJob\".\"data\", …" db.statement="\n\nSELECT\n  \"KjuJob\".\"id\",\n  \"KjuJob\".\"type\",\n  \"KjuJob\".\"data\",\n  CAST(\"KjuJob\".\"state\" AS text),\n  \"KjuJob\".\"created\",\n  \"KjuJob\".\"updated\"\nFROM\n  \"KjuJob\"\nWHERE\n  \"KjuJob\".\"state\" = CAST($1 AS KjuJobState)\n  OR \"KjuJob\".\"state\" = CAST($2 AS KjuJobState)\n" rows_affected=0 rows_returned=0 elapsed=19.019766ms
Error: Query Error: error returned from database: type "kjujobstate" does not exist
```

All type names in this query are enquoted and thus case-sensitive, except in `\"KjuJob\".\"state\" = CAST($2 AS KjuJobState)`.

### Reproduces How Often

Every time

## Versions

0.30.0

PR #789, opened 2024-06-25
Case-sensitive enum type casts

niklaskorz:
## PR Info

- Closes https://github.com/SeaQL/sea-query/issues/725
- Dependents:
  - https://github.com/SeaQL/sea-orm/pull/2271

## Bug Fixes

- [x] Casting to an enum whose type name was declared case-sensitive does not crash anymore

## Breaking Changes

- [x] Enum type casts are now case sensitive, so casting to `SomeEnum` will now cast to `"SomeEnum"` on PostgreSQL, while it previously casted to `someenum`.


niklaskorz:
This still has problems with arrays, as sea-orm constructs array types directly as `{enum_name}[]`, making it hard to only quote the type name but not the array brackets from sea-query.

tyt2y3 (approved):
Thank you!

mpyw:
@niklaskorz Please check the issue above. Thank you 🙇

PR #769, opened 2024-04-15
issues-721 Add impl `IdenStatic` and `AsRef<str>` to `#[enum_def]` attr

AndreiOrmanji:
## PR Info

- Closes https://github.com/SeaQL/sea-query/issues/721

## New Features

- [x] Copied `#[enum_def]` to `sea-query-derive` (and marked `#[enum_def]` as deprecated at `sea-query-attr`)
- [x] Added impl `IdenStatic` to `#[enum_def]` of `sea-query-derive`
- [x] Added impl `AsRef<str>` to `#[enum_def]` of `sea-query-derive`
- [x] Updated `syn` crate from (`1` -> `2`) (by merging #799, created by @sandhose)
- [x] Updated `heck` crate from (`0.4` -> `0.5`) (by merging #799, created by @sandhose)
- [x] Updated `darling` crate from (`0.14` -> `0.20`) (by merging #799, created by @sandhose)

PR #799, opened 2024-08-06
sea-query-attr: update syn, heck and darling

sandhose:
I wrote this before seing #769, which also upgrades the dependencies.
There are still a few minor improvements #769 should probably pick up, so I'm opening this PR for visibility.


tyt2y3:
769 is merged! thank you

Issue #809, opened 2024-08-30
Allow using sqlx 0.8.1 with sea-query-binder


sandhose:
## Motivation

sqlx 0.8.1 got out a week ago with no breaking API change, but `sea-query-binder` sets the dependency to `<0.8.1`

## Proposed Solutions

Relax the requirement to `<0.9.0`, so that we keep compatibility with old version, but still trust sqlx for not doing API-breaking changes in patch versions

## Additional Information

–

Jupp56:
To add: sqlx 0.8.1 fixes a [security vulnerability](https://rustsec.org/advisories/RUSTSEC-2024-0363.html). This is actually an important update, I think.

billy1624:
Just released https://crates.io/crates/sea-query-binder/0.7.0-rc.2

PR #810, opened 2024-08-30
sea-query-binder: relax the sqlx dependency range

sandhose:
## PR Info

- Closes https://github.com/SeaQL/sea-query/issues/809

## Changes

- Compatibility with sqlx 0.8.1+


billy1624 (approved):
Thanks!! @sandhose I think it's more intuitive to write `0.8` :P

I will release a new RC version today

billy1624:
Just released https://crates.io/crates/sea-query-binder/0.7.0-rc.2

sandhose:
> I think it's more intuitive to write 0.8 :P

Definitely! It's just that I thought the range was there to also support older versions, so my goal was to loosen up the range without denying older versions

PR #814, opened 2024-09-06
Remove separate WithQuery to ensure it can be specified everywhere

### LINKED ISSUE #813: Unify SelectStatement and WithQuery
## Motivation

WithQuery is currently a separate entity from SelectStatement which is unfortunate, because it restricts where you can use it, i.e. I'd like to use a WithQuery as a FROM with `from_subquery` in another query but I don't believe I can, because only SelectStatements are accepted.

https://github.com/SeaQL/sea-query/blob/53fd4e902910cf3db9f4332f1c1d1b6b6b3e49a6/src/query/select.rs#L1024

## Proposed Solutions

Move `with_clause` into `SelectStatement`/`UpdateStatement`/`DeleteStatement`/`InsertStatement` instead of having a separate code path for it.

bouk:
Instead, move the with clause into insert/update/select/delete. This makes sure you can specify with WITH clause to any method that takes a `SelectStatement`/`UpdateStatement`/`InsertStatement`/`DeleteStatement`.

## PR Info

- Closes #813

## Breaking Changes

- [x] This removes `WithQuery` and changes the signature of `.with()`

### DIFF
--- a/sea-query-binder/src/sqlx.rs
+++ b/sea-query-binder/src/sqlx.rs
@@ -25,4 +25,3 @@ impl_sqlx_binder!(SelectStatement);
 impl_sqlx_binder!(UpdateStatement);
 impl_sqlx_binder!(InsertStatement);
 impl_sqlx_binder!(DeleteStatement);
-impl_sqlx_binder!(WithQuery);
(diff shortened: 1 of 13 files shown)


tyt2y3:
Thank you for your contribution. While I appreciate now that we can use a with clause in all statement types, removing `WithQuery` is breaking. Is there a middleground where we still keep it?

bouk:
We could keep both kinds and deprecate WithQuery, removing it in a later version.

rumpuslabs:
@tyt2y3 @bouk see #859 which implements the suggested approach

tyt2y3:
Closed in favour of https://github.com/SeaQL/sea-query/pull/859

PR #817, opened 2024-09-18
Replace Educe derive macro with manual trait implementations

svix-jplatte:
## Changes

This PR replaces the one and only use of the `educe` crate with manual trait implementations of the derived traits.

`educe` takes more than 10s to compile on my machine, whereas `sea-query` itself takes less than 5s. To me that is a very heavy dependency for how little it is used. The manual implementations of the traits in question should be easy to read and not hard to maintain.


tyt2y3:
I understand your concern. Derive macros are indeed heavy and educe has some very advanced features.
I was just too lazy to write all those by hand haha.

tyt2y3 (approved):

PR #818, opened 2024-09-20
Derive additional traits for sea_query::Alias

just-in-chang:
## Changes

Adds derivations for the `PartialEq`, `Eq`, `PartialOrd`, `Ord`, and `Hash` traits to `sea_query::Alias`. This will greatly help those working with aliases by allowing devs to treat them like strings instead of needing to convert them first through the `Iden` trait.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -241,7 +241,7 @@ pub enum Order {
 }
 
 /// Helper for create name alias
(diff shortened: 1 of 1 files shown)


tyt2y3 (approved):
You're right!

Issue #403, opened 2022-08-04
Add support for Postgresql's function and trigger


HigherOrderLogic:
## Motivation

Give users the ablity to create functions and triggers on database by using `sea-query` instead of raw SQL.

## Proposed Solutions

N/A

## Additional Information

https://www.postgresql.org/docs/current/plpgsql-trigger.html

billy1624:
Hey @HigherOrderLogic, thanks for the feature request!

I think creating, dropping and altering of function and trigger can be offered under [`sea_query::extension::postgres`](https://github.com/SeaQL/sea-query/tree/master/src/extension/postgres) module.

- [CREATE FUNCTION](https://www.postgresql.org/docs/current/sql-createfunction.html)
- [DROP FUNCTION](https://www.postgresql.org/docs/current/sql-dropfunction.html)
- [ALTER FUNCTION](https://www.postgresql.org/docs/current/sql-alterfunction.html)
- [CREATE TRIGGER](https://www.postgresql.org/docs/current/sql-createtrigger.html)
- [DROP TRIGGER](https://www.postgresql.org/docs/current/sql-droptrigger.html)
- [ALTER TRIGGER](https://www.postgresql.org/docs/current/sql-altertrigger.html)

billy1624:
I'll open this for contributions! :)

ikrivosheev:
@billy1624 @tyt2y3 we really want to add support for this functionality?

anshulxyz:
If you are open to contribution for this issue, then I'd like to work on it

billy1624:
Hey @anshulxyz, this is still open for contribution! Please go ahead :D

stevenhansel:
@billy1624 I'm not sure whether @anshulxyz is still working on this issue or not but I'm going to add support for the `trigger` first

anshulxyz:
@stevenhansel I'm not, you can take over the issue.

billy1624:
Thanks!! @stevenhansel Let me know if you need any help :)

billy1624:
Hey @HigherOrderLogic, you can take this as a reference [TypeCreateStatement](https://docs.rs/sea-query/0.26.3/sea_query/extension/postgres/struct.TypeCreateStatement.html#). We need to create a struct let say `TriggerCreateStatement` under `extension/postgres` module since this is Postgres specific. Looking at the docs [CREATE TRIGGER](https://www.postgresql.org/docs/current/sql-createtrigger.html).

I imagine the API be like...

[...]

PR #824, opened 2024-10-05
[WIP] Triggers

gronke:
Issue https://github.com/SeaQL/sea-query/issues/403 discusses ideas and interfaces to implement SQL triggers. Inspired by this conversation, this pull-request suggests a way to implement triggers.

This PR is in an early stage and meant to provide surface for discussion and tinkering.

```rust
// the most simple way to create a trigger
let unnamed_trigger: UnnamedTrigger =  UnnamedTrigger::new();

// it can be named later, and become a NamedTrigger
let named_trigger: NamedTrigger = unnamed_trigger.name("my_trigger");

// this is useful because it allows dropping directly
let drop_sql: String = named_trigger.drop().to_string(MysqlQueryBuilder);

// with a table, action and action time we dare do derive a trigger name
let trigger: DefinedTrigger = unnamed_trigger.before_insert(Glyph::Table);
assert_eq!(trigger.trigger_name(), "t_glyph_before_insert");

// demo expression, blocked by missing features (e.g. IF-ELSE statements)
let action: SimpleExpr = Expr::col(Glyph::Id).eq(1);

// multiple actions will be executed in order of insertion
trigger.actions.push(action);

println!(trigger.create().to_string(MysqlQueryBuilder));
```

```sql
CREATE TRIGGER `t_glyph_before_insert` BEFORE INSERT ON `glyph` FOR EACH ROW
BEGIN
`id` = 1;
END
```

The action expression (between BEGIN and END) does not make sense yet and exists to outline the idea. With new features to express logic, variable declaration, loops and custom errors this could be achieved though.

## New Features

- [x] Create and drop TRIGGER
- [x] Named and unnamed triggers (name derived from table, action and time)
- [x] Concatenate list of statements 

## TBD

- Missing statement implementations
  - [ ] IF-ELSE statements
  - [ ] LOOP
  - [ ] custom SQL errors
- [ ] Use wrappers around triggers to polyfill [`CKECK` constraints](https://www.postgresql.org/docs/current/ddl-constraints.html) for MySQL and Sqlite
- [ ] Buil unit-tests for PostgreSQL and SQlite3
- [ ] Address `IF NOT EXISTS` differences between database backends + write unit-tests
- [ ] Integration testing against live backends before engaging in fancy feature wrappers

### DIFF
--- a/src/backend/trigger_builder.rs
+++ b/src/backend/trigger_builder.rs
@@ -1,6 +1,6 @@
 use crate::*;
 
-pub trait TriggerBuilder: TableRefBuilder {
+pub trait TriggerBuilder: TableRefBuilder + QueryBuilder {
     /// Translate [`TriggerCreateStatement`] into SQL statement.
     fn prepare_trigger_create_statement(
         &self,
@@ -21,7 +21,16 @@ pub trait TriggerBuilder: TableRefBuilder {
         self.prepare_table_ref_iden(&create.trigger.table, sql);
         write!(sql, " FOR EACH ROW\nBEGIN\n").unwrap();
 
-        write!(sql, "\nEND").unwrap();
+        self.prepare_trigger_actions(&create.trigger.actions, sql);
+
+        write!(sql, "END").unwrap();
+    }
+
+    fn prepare_trigger_actions(&self, actions: &TriggerActions, sql: &mut dyn SqlWriter) {
+        for action in actions {
+            self.prepare_simple_expr_common(&action, sql);
+            write!(sql, ";\n").unwrap();
+        }
     }
 
     /// Translate IF NOT EXISTS expression in [`TriggerCreateStatement`].
@@ -44,5 +53,9 @@ pub trait TriggerBuilder: TableRefBuilder {
     fn prepare_trigger_drop_statement(&self, drop: &TriggerDropStatement, sql: &mut dyn SqlWriter) {
         write!(sql, "DROP TRIGGER ").unwrap();
         self.prepare_table_ref_iden(&drop.name.clone().into(), sql);
+    }
+
+    fn prepare_simple_expr_yeah(&self, simple_expr: &SimpleExpr, sql: &mut dyn SqlWriter) {
+        self.prepare_simple_expr_common(simple_expr, sql);
     }
 }
--- a/src/backend/trigger_builder.rs
+++ b/src/backend/trigger_builder.rs
@@ -1,0 +1,48 @@
+use crate::*;
+
+pub trait TriggerBuilder: TableRefBuilder {
+    /// Translate [`TriggerCreateStatement`] into SQL statement.
+    fn prepare_trigger_create_statement(
+        &self,
+        create: &TriggerCreateStatement,
+        sql: &mut dyn SqlWriter,
+    ) {
+        write!(sql, "CREATE TRIGGER ").unwrap();
+        self.prepare_create_trigger_if_not_exists(create, sql);
+
+        let trigger_ref = match &create.trigger.name {
+            Some(value) => value,
+            // auto-generate trigger name
+            _ => &create.trigger.trigger_ref(),
+        };
+        let trigger_ref: TableRef = trigger_ref.into();
+        self.prepare_table_ref_iden(&trigger_ref, sql);
+        write!(sql, " {} {} ON ", create.trigger.time, create.trigger.event).unwrap();
+        self.prepare_table_ref_iden(&create.trigger.table, sql);
+        write!(sql, " FOR EACH ROW\nBEGIN\n").unwrap();
+
+        write!(sql, "\nEND").unwrap();
+    }
+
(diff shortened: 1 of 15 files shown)


src/backend/trigger_builder.rs
```diff
@@ -0,0 +1,61 @@
...
+        let trigger_ref = match &create.trigger.name {
+            Some(value) => value,
+            // auto-generate trigger name
+            _ => &create.trigger.trigger_ref(),
+        };
+        let trigger_ref: TableRef = trigger_ref.into();
+        self.prepare_table_ref_iden(&trigger_ref, sql);
+        write!(sql, " {} {} ON ", create.trigger.time, create.trigger.event).unwrap();
+        self.prepare_table_ref_iden(&create.trigger.table, sql);
+        write!(sql, " FOR EACH ROW\nBEGIN\n").unwrap();
```
pcwizz, line 22: It would be nice to have statement level triggers although they aren't available in mysql.

src/trigger/mod.rs
```diff
@@ -0,0 +1,356 @@
...
+            TriggerActionTime::After,
+        )
+    }
+}
+
+pub type TriggerAction = SimpleExpr;
+pub type TriggerActions = Vec<TriggerAction>;
+
+#[derive(Default, Debug, Clone)]
+pub struct NamedTrigger {
```
HigherOrderLogic, line 79: How about unifying the two into one with `fn named(...)` and `fn unnamed(...)`?

gronke: The two structs `UnnamedTrigger` and `NamedTrigger` are distinguished because of slightly different interfaces. Generating trigger names is only needed when a user did not specify an explicit name. Differentiating structs within the Trigger enum allows auto-completion.

The Trigger implementation has two methods `new` and `with_name` that are similar to your suggestion:

```
impl Trigger {
    pub fn new() -> Trigger {
        Trigger::UnnamedTrigger(None, None, None)
    }

    pub fn with_name(name: TriggerRef) -> Trigger {
        Trigger::NamedTrigger(name, None, None, None)
    }
}
```

@HigherOrderLogic do you think we should get rid of the two variants and accept that users might inadvertently use unnamed triggers? Typically I prefer being explicit, but I also see the potential to reduce complexity. Happy to hear your suggestion. 😊

HigherOrderLogic: > do you think we should get rid of the two variants and accept that users might inadvertently use unnamed triggers

I think that an unified enum would be better for end users DX, however, I disagree that unnamed should be the default.

As I stated in the previous comment, we can implement different functions for each variants, eg:

```rs
impl Trigger {
    pub fn unnamed() -> Trigger {
        Trigger::UnnamedTrigger(None, None, None)
    }

    pub fn named(name: TriggerRef) -> Trigger {
        Trigger::NamedTrigger(name, None, None, None)
    }
}
```

gronke: I've played with the thought of a unified Trigger and find it to conflict with the level of definition a trigger requires to be dropped.

Looking at MySQL tests for instance [tests/mysql/trigger.rs#L44-L60](https://github.com/SeaQL/sea-query/blob/cea2aefb1526d4c02ab739a93ebea3f972ac9765/tests/mysql/trigger.rs#L45):

A NamedTrigger is Referencable and can be dropped right away, without further definition:

```rust
#[test]
fn drop_named_trigger() {
    let trigger = NamedTrigger::new("my_trigger");
    assert_eq!(
        trigger.drop().to_string(MysqlQueryBuilder),
        "DROP TRIGGER `my_trigger`"
    );
}
```

To automatically generate a reasonably unique trigger name, we need to know more information that we will define when configuring the Configurable named or unnamed trigger into a DefinedTrigger.

```rust
#[test]
fn drop_unnamed_trigger() {
    let trigger = UnnamedTrigger::new().before_delete(Glyph::Table);
    assert_eq!(
        trigger.drop().to_string(MysqlQueryBuilder),
        "DROP TRIGGER `t_glyph_before_delete`"
    );
}
```

A DefinedTrigger falls back to generating the trigger name by concatenating table name, trigger action time and event (e.g. `t_glyph_before_delete`), which occurs in [src/trigger/mod.rs#L173-L183](https://github.com/gronke/sea-query/blob/cea2aefb1526d4c02ab739a93ebea3f972ac9765/src/trigger/mod.rs#L173-L183)

```rust
fn trigger_name(&self) -> String {
    match &self.name {
        Some(name) => name.to_string(),
        _ => format!(
            "t_{}_{}_{}",
            self.table.to_string().to_lowercase(),
            self.time.to_string().to_lowercase(),
            self.event.to_string().to_lowercase(),
        ),
    }
}
```

How about we call the `UnnamedTrigger` just `Trigger` and remove the variant enum? One can call `.name()` on it and make it named without bothering about internals.

gronke:
@Ichmed yes, I do. Took me a while to get back to this though. Sorry for the delay.

PR #825, opened 2024-10-05
Add Postgre function: `DATE_TRUNC`

IgnisDa:
## PR Info

This PR adds the `DATE_TRUNC` functions along with docs and tests.

## Additional

N/A

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,25 +30,17 @@ path = "src/lib.rs"
 [dependencies]
 inherent = "1.0"
 sea-query-derive = { version = "0.4.2", path = "sea-query-derive", default-features = false, optional = true }
-serde_json = { version = "1", default-features = false, optional = true, features = [
-    "std",
-] }
-chrono = { version = "0.4.27", default-features = false, optional = true, features = [
-    "clock",
-] }
+serde_json = { version = "1", default-features = false, optional = true, features = ["std"] }
+chrono = { version = "0.4.27", default-features = false, optional = true, features = ["clock"] }
 postgres-types = { version = "0", default-features = false, optional = true }
 pgvector = { version = "~0.4", default-features = false, optional = true }
 rust_decimal = { version = "1", default-features = false, optional = true }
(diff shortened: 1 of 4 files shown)


IgnisDa:
@tyt2y3 Could you take a look at this PR? Thanks!

Cargo.toml
```diff
@@ -30,17 +30,25 @@ path = "src/lib.rs"
 [dependencies]
 inherent = "1.0"
 sea-query-derive = { version = "0.4.2", path = "sea-query-derive", default-features = false, optional = true }
-serde_json = { version = "1", default-features = false, optional = true, features = ["std"] }
-chrono = { version = "0.4.27", default-features = false, optional = true, features = ["clock"] }
+serde_json = { version = "1", default-features = false, optional = true, features = [
```
ikrivosheev, line 33: Please, revert changes in the `Cargo.toml`

ikrivosheev (commented):
@IgnisDa hello! Thank you for the PR. Little comment)

Cargo.toml
IgnisDa: thanks for taking a look! I understand that the formatting changes need to be reverted. But i added a new dependency `strum_macros` (which is already a depdency of sea-orm), i should keep it there, right?

IgnisDa: @ikrivosheev Made the request changes. Could you please review it again?

src/table/column.rs
```diff
@@ -1,3 +1,5 @@
+use std::fmt::Display;
```
ikrivosheev, line 1: I prefer to write like: `use std::fmt` and then write everywhere with prefix

ikrivosheev (commented):
Small comment)

src/table/column.rs
IgnisDa: made the requested changes

ikrivosheev (approved):
@IgnisDa thank you! LGTM!

tyt2y3 (commented):
Thank you

PR #829, opened 2024-10-15
feat: ExceptionStatement raises custom SQL errors

gronke:
## PR Info

Allows defining SQL errors, for instance in an `IfElseStatement` #828 inside a `TriggerStatement` #824 that rejects an update on certain conditions.

## New Features

- [ ] `ExceptionStatement` raises custom SQL errors

### DIFF
--- a/src/backend/query_builder.rs
+++ b/src/backend/query_builder.rs
@@ -387,6 +387,9 @@ pub trait QueryBuilder:
             SimpleExpr::Constant(val) => {
                 self.prepare_constant(val, sql);
             }
+            SimpleExpr::Exception(val) => {
+                self.prepare_exception_statement(val, sql);
+            }
         }
     }
 
@@ -988,6 +991,15 @@ pub trait QueryBuilder:
             Keyword::CurrentTimestamp => write!(sql, "CURRENT_TIMESTAMP").unwrap(),
             Keyword::Custom(iden) => iden.unquoted(sql.as_writer()),
         }
+    }
+
+    // Translate [`Exception`] into SQL statement.
+    fn prepare_exception_statement(
+        &self,
+        _exception: &ExceptionStatement,
+        _sql: &mut dyn SqlWriter,
+    ) {
+        panic!("Exception handling not implemented for this backend");
     }
 
     /// Convert a SQL value into syntax-specific string
--- a/src/exception.rs
+++ b/src/exception.rs
@@ -1,0 +1,46 @@
+//! Custom SQL exceptions and errors
+use inherent::inherent;
+
+use crate::backend::SchemaBuilder;
+
+/// SQL Exceptions
+#[derive(Debug, Clone, PartialEq)]
+pub struct ExceptionStatement {
+    pub(crate) message: String,
+}
+
+impl ExceptionStatement {
+    pub fn new(message: String) -> Self {
+        Self { message }
+    }
+}
+
+pub trait ExceptionStatementBuilder {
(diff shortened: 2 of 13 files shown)


src/expr.rs
```diff
@@ -35,6 +35,7 @@ pub enum SimpleExpr {
     AsEnum(DynIden, Box<SimpleExpr>),
     Case(Box<CaseStatement>),
     Constant(Value),
+    Exception(ExceptionStatement),
```
Expurple, line 38: You should mention that this is a breaking change (adding a new variant to a public enum not marked as `#[non_exhaustive]`)

Expurple (commented):
Haven't reviewed the rest yet. Consider running `cargo semver-checks` to find and document the breaking changes.

Disclaimer: I'm not a maintainer

tyt2y3:
Expurple is a big contributor :)

> we should add #[non_exhaustive] in the next major version

I agree, and I am about to start merging breaking changes now

src/exception.rs
```diff
@@ -0,0 +1,46 @@
+//! Custom SQL exceptions and errors
+use inherent::inherent;
+
+use crate::backend::SchemaBuilder;
+
+/// SQL Exceptions
+#[derive(Debug, Clone, PartialEq)]
+pub struct ExceptionStatement {
+    pub(crate) message: String,
+}
```
Expurple, line 10: As I look at the [PostgreSQL docs](https://www.postgresql.org/docs/17/plpgsql-errors-and-messages.html), I'd rather have a powerful Postgres-specific `PgRaiseStatement` with an explicit

```rust
#[non_exhaustive]
pub enum PgRaiseLevel {
    Exception,
    ..
}
```

where later we could extend `PgRaiseStatement` in a backwards-compatible way and add methods for format args, `SQLSTATE` and `USING` (I'm not asking to contribute everything under the sun in this one PR).

We can keep your minimal "portable" API too, if you document that it's just a shortcut for the common subset of database-specific exception APIs (which all seem very different from each other, judging by the SQL you generate in the implementation). Obviously, it'd switch its implementation to just delegate to the database-specific query builder: `PgRaiseStatement`, etc.

src/backend/query_builder.rs
```diff
@@ -1051,6 +1054,15 @@ pub trait QueryBuilder:
...
     }
 
+    // Translate [`Exception`] into SQL statement.
+    fn prepare_exception_statement(
+        &self,
+        _exception: &ExceptionStatement,
+        _sql: &mut dyn SqlWriter,
+    ) {
+        panic!("Exception handling not implemented for this backend");
+    }
```
Expurple, line 1064: Oh, this doesn't look good to me. Why make it a provided method if you don't, in fact... *provide* it?

To me, such a provided method would look OK if it returned `Result<(), ExceptionsNotImplemented>`. But I see that the calling `prepare_simple_expr_common` doesn't return a `Result` anyway. Perhaps, we've hit some problem/limitation of the `sea_query` design that I can't fully understand yet. Perhaps, we shouldn't even try to pretend that exceptions are a portable and universally-supported feature (see https://github.com/SeaQL/sea-query/pull/829#discussion_r2108221156)

src/expr.rs
Expurple: https://github.com/SeaQL/sea-query/pull/891 has been merged, `SimpleExpr` is now `#[non_exhaustive]`. Your change is no longer breaking. You can update the PR description

Issue #608, opened 2023-02-21
Support `JOIN`s in `UpdateStatement` (`join_subquery`, `inner_join`, etc.)


LeoniePhiline:
## Motivation

It is rather handy (and often required?) to gather related entities in an `UPDATE` statement, and update some fields based on joined tables.

For example, here each `institution` has several `subscriptions` (with `subscriptions.institution` being a foreign key to `institutions.uid`). An append-only activity table records users of these institutions using the subscriptions. Periodically, the institutions' and subscriptions' `last_activity` datetime field shall be updated.

### Updating `last_activity` on subscriptions

```sql
UPDATE subscriptions AS entity
INNER JOIN (
    SELECT subscription.uid AS entity_uid, MAX(date_time) AS last_activity
    FROM activity
    INNER JOIN subscriptions AS  subscription ON subscription.uid = activity.subscription
    GROUP BY subscription.uid
) AS entity_activity
    ON entity_activity.entity_uid = entity.uid
SET entity.last_activity = entity_activity.last_activity;
```

### Updating `last_activity` on institutions
```sql
UPDATE institutions AS entity
INNER JOIN (
    SELECT institution.uid AS entity_uid, MAX(date_time) AS last_activity
    FROM activity
    INNER JOIN subscriptions AS subscription ON subscription.uid = activity.subscription
    INNER JOIN institutions AS institution ON institution.uid = subscription.institution
    GROUP BY institution.uid
) AS entity_activity
    ON entity_activity.entity_uid = entity.uid
SET entity.last_activity = entity_activity.last_activity;
```

## My wish here was to use (simplified):

```rust
let query = Query::update()
    .table(Subscriptions::Table)
    .join_subquery(
        JoinType::InnerJoin,
        ...
    )
    .value(...);
```

But joins are unavailable on `UpdateStatement`.

## Proposed Solutions

Add methods available on `SelectStatement` to `UpdateStatement`:

- `join`
- `join_as`
- `join_lateral`
- `join_subquery`
- `inner_join`
- `left_join`
- `right_join`
- `full_outer_join`

## Additional Information

I will try to work around this using CTEs (`UpdateStatement::with`).

tyt2y3:
SeaQuery has support for `UPDATE SELECT FROM` and `UPDATE SET=SUB_QUERY`, so they might also do the same trick.

LeoniePhiline:
I'm afraid these aren't supported by MariaDB. 

[...]

PR #833, opened 2024-11-01
feat: support `UPDATE ... FROM ...` syntax

the-wondersmith:
## PR Info

- Partially addresses [#608](https://github.com/SeaQL/sea-query/issues/608#issuecomment-1441261493)

## New Features

- [ ] Adds support for explicitly specifying a `FROM` clause in an `UPDATE` query
    - e.x. `UPDATE table1 SET some_column = table2.another_column FROM table2 WHERE table1.another_column = table2.a_different_column`

### DIFF
--- a/src/query/update.rs
+++ b/src/query/update.rs
@@ -1,3 +1,5 @@
+use inherent::inherent;
+
 use crate::{
     backend::QueryBuilder,
     expr::*,
@@ -8,7 +10,6 @@ use crate::{
     QueryStatementBuilder, QueryStatementWriter, ReturningClause, SubQueryStatement, WithClause,
     WithQuery,
 };
-use inherent::inherent;
 
 /// Update existing rows in the table
 ///
@@ -99,7 +100,7 @@ impl UpdateStatement {
     ///
     /// assert_eq!(
     ///     query.to_string(MysqlQueryBuilder),
-    ///     "UPDATE `glyph` SET `tokens` = `character`.`character` FROM `character` WHERE `glyph`.`image` = `character`.`user_data`"
+    ///     "UPDATE `glyph` JOIN `character` ON `glyph`.`image` = `character`.`user_data` SET `glyph`.`tokens` = `character`.`character`"
     /// );
     /// assert_eq!(
     ///     query.to_string(PostgresQueryBuilder),
--- a/src/query/update.rs
+++ b/src/query/update.rs
@@ -39,6 +39,7 @@ use inherent::inherent;
 #[derive(Default, Debug, Clone, PartialEq)]
 pub struct UpdateStatement {
     pub(crate) table: Option<Box<TableRef>>,
+    pub(crate) from: Vec<TableRef>,
     pub(crate) values: Vec<(DynIden, Box<SimpleExpr>)>,
     pub(crate) r#where: ConditionHolder,
     pub(crate) orders: Vec<OrderExpr>,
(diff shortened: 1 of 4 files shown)


the-wondersmith:
Just tagging @tyt2y3 @billy1624 (as the most active/recent mergers) to try and get some traction on this 😅

tyt2y3:
Thank you for the PR! sadly `UPDATE ... FROM` is not valid syntax on MySQL.
I think it is also supported in SQLite, so it's worth adding, but may be we need to indicate this in the docs

the-wondersmith:
> Thank you for the PR! sadly `UPDATE ... FROM` is not valid syntax on MySQL. I think it is also supported in SQLite, so it's worth adding, but may be we need to indicate this in the docs

Totally valid. We can restrict it by feature flag, if that works? So it's only available when the proper feature flags are enabled, and the emitted queries are syntactically correct for the enabled feature-set. 

Out of curiosity, do you happen to know the equivalent syntax for MySQL? Once-upon-a-time I was *extremely* familiar with the major flavors of SQL syntax, as I was doing a lot of tinkering in the internals of SQLAlchemy due to being the maintainer for a couple of esoteric dialect for that package. In the intervening time, I've transitioned to rust almost completely, so tl;dr SeaORM/SeaQuery are the gotos AFAIC 😅

Anyway, if your amenable to the update, I'll sort out the syntax / engine pairings and update the PR.

src/query/update.rs
```diff
@@ -66,6 +67,62 @@ impl UpdateStatement {
...
+    ///             SeaRc::new(Char::Table),
+    ///             SeaRc::new(Char::UserData),
+    ///         ))),
+    ///     )
+    ///     .to_owned();
+    ///
+    ///
+    /// assert_eq!(
+    ///     query.to_string(MysqlQueryBuilder),
+    ///     "UPDATE `glyph` SET `tokens` = `character`.`character` FROM `character` WHERE `glyph`.`image` = `character`.`user_data`"
```
gronke, line 102: The syntax of MySQL appears similar, but not identical. Here a MySQL UPDATE JOIN could be used to achieve the same:

```sql
UPDATE `glyph`
JOIN `character`
  ON `glyph`.`image` = `character`.`user_data`
SET `glyph`.`tokens` = `character`.`character`;
```

gronke (changes requested):

src/query/update.rs
the-wondersmith: @gronke Took a wild swing at at least getting started here, and extracted the specific behavior into a discrete method (`prepare_update_from_table_refs`) so it can be overridden as needed by whichever dialect. I've got the MySQL-specific handling as a `unimplemented!` call at the moment though, cause I've got a case of end-of-year-holidays-brain.

If you have a second, would you take a look at the changes and let me know if there's a better / more preferable way of doing this please let me know? Otherwise I'll carry on as time allows 😁

the-wondersmith: @gronke ^bump

just-in-chang:
Would love to see this landed! This would provide support for stuff like using SeaQuery to construct queries for batch update operations that utilize CTEs and value tuples.

the-wondersmith:
> Would love to see this landed! This would provide support for stuff like using SeaQuery to construct queries for batch update operations that utilize CTEs and value tuples.

I'd love to *get* it landed 😅. I see you've got a `Contributor` tag next to your name here. Any chance you could poke @gronke for some feedback [here](https://github.com/SeaQL/sea-query/pull/833#discussion_r1899175577)?

src/query/update.rs
the-wondersmith: @gronke **_^bump_** 🙂

tyt2y3:
I will give it a try

PR #835, opened 2024-11-08
Fix #830, remove incorrect restriction on recursive CTE

### LINKED ISSUE #830: Multiple CTEs should be allowed in recursive WithQuery
## Description

Per https://github.com/SeaQL/sea-query/blob/master/src/backend/query_builder.rs#L780 when you do a recursive with you are only allowed to have one CTE.

In sqlite at least (and I'd expect other dbs as well) you can have any number of CTEs in the with query, per https://sqlite.org/lang_with.html .  Especially see example 3.5 which has 4 recursive CTEs and one non-recursive CTE (`a`).

I'm not sure what the impetus for that line was, but would dropping it be reasonable? Are there other parts of the code that rely on it having only one CTE?

andrewbaxter:
I tested this with sqlite and the queries execute successfully. It's possible that there are other incorrect behaviors that can be pre-emptively caught, perhaps those could be added in a new PR later.

## PR Info

- Closes #830 

## Bug Fixes

- [x] Remove incorrect restriction on recursive CTE

### DIFF
--- a/src/backend/query_builder.rs
+++ b/src/backend/query_builder.rs
@@ -777,15 +777,6 @@ pub trait QueryBuilder:
             "Cannot build a with query that has no common table expression!"
         );
 
-        if with_clause.recursive {
-            assert_eq!(
-                with_clause.cte_expressions.len(),
-                1,
-                "Cannot build a recursive query with more than one common table! \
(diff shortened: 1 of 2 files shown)


tyt2y3 (approved):
thank you

PR #844, opened 2024-11-30
feat(function): Add support for GREATEST / LEAST function

tinou98:
## New Features

- [ ] Add support for `GREATEST` / `LEAST` SQL functions
As those functions are not supported on SQLite, this fallback to plain MAX/MIN.


tyt2y3 (approved):
Thank you for the contribution

PR #852, opened 2024-12-31
feat: add Expr::column

tisonkun:
The same as `Expr::value` to `Expr::val`.

I'm using this and now work it around with `Expr::col(...).into()`.

### DIFF
--- a/src/expr.rs
+++ b/src/expr.rs
@@ -1482,6 +1482,62 @@ impl Expr {
         Self::new_with_left(n.into_column_ref())
     }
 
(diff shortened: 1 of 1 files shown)


ikrivosheev (approved):
LGTM

tisonkun:
@ikrivosheev Thanks for your review!

May I ask what the normal process for merging a patch and cutting a new release? I'd expect a new release with this patch so that I can switch my project to depend on the new version for the changeset.

tyt2y3:
Thanks @ikrivosheev ! Let's merge this

PR #854, opened 2025-01-02
fix: build empty postgres-array with auto type infer

### LINKED ISSUE #853: Empty PostgreSQL ARRAY should be printed with a cast
Otherwise, it will fail with "cannot determine type of empty array".

For example, an empty string vec will be converted to `ARRAY []` but PG needs `ARRAY[]::TEXT[]` or things like that for understanding the type.

tisonkun:
This fixes #853.


ikrivosheev (approved):
@tisonkun thanks! LGTM

tyt2y3:
Thank you

PR #859, opened 2025-01-24
Add with_cte to include WITH clauses in Statements

### LINKED ISSUE #813: Unify SelectStatement and WithQuery

rumpuslabs:
…instead of separate WithQuery

## PR Info

- Closes #813
- Replaces #814, keeping `WithQuery` for backward compatibility

## New Features

- Adds `with_cte` method to `SelectStatement`/`UpdateStatement`/`InsertStatement`/`DeleteStatement` to allow adding WITH clauses to any query

## Notes

- I had to choose another method name to avoid breaking backward compatibiity of the existing `fn with(...) -> WithQuery`. I'm not completely happy with `with_cte` but couldn't think of anything better.
- If the intention is to eventually remove `WithQuery`, we could add `#[deprecated]` to the`with` methods to encourage users to migrate

### DIFF
--- a/src/backend/query_builder.rs
+++ b/src/backend/query_builder.rs
@@ -18,6 +18,10 @@ pub trait QueryBuilder:
 
     /// Translate [`InsertStatement`] into SQL statement.
     fn prepare_insert_statement(&self, insert: &InsertStatement, sql: &mut dyn SqlWriter) {
+        if let Some(with) = &insert.with {
+            self.prepare_with_clause(with, sql);
+        }
+
         self.prepare_insert(insert.replace, sql);
 
         if let Some(table) = &insert.table {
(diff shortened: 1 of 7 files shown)


tyt2y3 (approved):
Thank you

PR #869, opened 2025-03-22
Support for creating functional indexes in PostgreSQL and MySQL

### LINKED ISSUE #756: Postgres: Support creating functional indexes
## Motivation

Create functional indexes

> for example, a common way to do case-insensitive comparisons is to use the lower function:

> SELECT * FROM test1 WHERE lower(col1) = 'value';
> This query can use an index, if one has been defined on the result of the lower(column) operation:

> CREATE INDEX test1_lower_col1_idx ON test1 (lower(col1));

Source: https://www.postgresql.org/docs/7.3/indexes-functional.html

yuly3:
## PR Info

- Closes #756 

## New Features

- [x] Support for creating functional indexes in PostgreSQL and MySQL

## Breaking Changes

- Change public Struct `index::common::IndexColumn` to Enum

## References

- PostgreSQL 17: <https://www.postgresql.org/docs/17/indexes-expressional.html>
- MySQL 8.4: <https://dev.mysql.com/doc/refman/8.4/en/create-index.html>

## Notes

I'm not entirely confident that this implementation is appropriate.
If you have any suggestions for a better implementation, I would greatly appreciate your feedback.

### DIFF
--- a/src/backend/index_builder.rs
+++ b/src/backend/index_builder.rs
@@ -57,6 +57,23 @@ pub trait IndexBuilder: QuotedBuilder + TableRefBuilder {
     }
 
     #[doc(hidden)]
+    /// Write the index column with table column.
+    fn prepare_index_column_with_table_column(
+        &self,
+        column: &IndexColumnTableColumn,
+        sql: &mut dyn SqlWriter,
+    ) {
+        column.name.prepare(sql.as_writer(), self.quote());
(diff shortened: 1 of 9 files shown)


tyt2y3 (commented):
thank you. looks good at first glance. I need more time to read it again

tyt2y3 (approved):
thank you

PR #873, opened 2025-04-01
Remove redundant `Expr::expr` from docs and internal code

Expurple:
## PR Info

- Closes: none
- Dependencies: none
- Dependents: none

## Changes

(probably not changelog-worthy)

- Removed redundant `Expr::expr` calls from docs and internal code.
- Updated the wording in the relevant docs.

Explained in more defail there:

```md
Since `sea_query` 0.32.0 (`sea_orm` 1.1.1), **this is not necessary** in most cases!

Some SQL operations used to be defined only as inherent methods on [`Expr`].
Thus, to use them, you needed to manually convert from other types to [`Expr`].
But now these operations are also defined as [`ExprTrait`] methods
that can be called directly on any expression type,
```

The diff speaks for itself. Let's move the ecosystem towards the cleaner style!

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,7 +218,8 @@ assert_eq!(CharacterIden::Foo.to_string(), "foo");
 
 ### Expression
 
-Use [`Expr`] to construct select, join, where and having expression in query.
+Use [`Expr`] constructors and [`ExprTrait`] methods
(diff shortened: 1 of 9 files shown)


tyt2y3 (approved):
thank you! it definitely looks nicer

PR #875, opened 2025-04-02
Make `RcOrArc` a documented type alias instead of a direct reexport

Expurple:
## PR Info

- Closes: none
- Dependencies: none
- Dependents: none

## Changes

(probably not changelog-worthy)

- Made `RcOrArc` a documented type alias instead of a direct reexport.

It [used to show](https://docs.rs/sea-query/0.32.3/sea_query/types/struct.RcOrArc.html) full documentation for `Rc`/`Arc`. I find it weird.

Now, it documents what `RcOrArc` actually is and how to control the representation.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -7,10 +7,19 @@ use std::{fmt, mem, ops};
 use crate::extension::postgres::PgBinOper;
 #[cfg(feature = "backend-sqlite")]
 use crate::extension::sqlite::SqliteBinOper;
+
+/// A reference counted pointer: either [`Rc`][std::rc::Rc] or [`Arc`][std::sync::Arc],
+/// depending on the feature flags.
+///
+/// [`Arc`][std::sync::Arc] is used when `thread-safe` feature is activated.
 #[cfg(not(feature = "thread-safe"))]
-pub use std::rc::Rc as RcOrArc;
+pub type RcOrArc<T> = std::rc::Rc<T>;
+/// A reference counted pointer: either [`Rc`][std::rc::Rc] or [`Arc`][std::sync::Arc],
+/// depending on the feature flags.
+///
+/// [`Arc`][std::sync::Arc] is used when `thread-safe` feature is activated.
 #[cfg(feature = "thread-safe")]
-pub use std::sync::Arc as RcOrArc;
+pub type RcOrArc<T> = std::sync::Arc<T>;
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub struct Quote(pub(crate) u8, pub(crate) u8);


tyt2y3:
nice. but apart from the docs, is there any functional difference between an alias and re-export? I am asking so as to not unknowingly introduced a breaking change

Expurple:
`cargo-semver-checks` doesn't report anything:

```txt
sea-query$ git status
On branch rc-or-arc-type-alias
Your branch is up to date with 'origin/rc-or-arc-type-alias'.

nothing to commit, working tree clean
```
```txt
sea-query$ cargo semver-checks --version
cargo-semver-checks 0.40.0
```
```txt
sea-query$ cargo semver-checks --release-type patch --baseline-rev master --all-features
     Cloning master
    Building sea-query v0.32.3 (current)
       Built [   2.118s] (current)
     Parsing sea-query v0.32.3 (current)
      Parsed [   0.065s] (current)
    Building sea-query v0.32.3 (baseline)
       Built [   2.105s] (baseline)
     Parsing sea-query v0.32.3 (baseline)
      Parsed [   0.062s] (baseline)
    Checking sea-query v0.32.3 -> v0.32.3 (assume patch change)
     Checked [   0.224s] 148 checks: 148 pass, 0 skip
     Summary no semver update required
    Finished [  13.238s] sea-query
```

A quick Google search found a few gotchas with type aliases ([1](https://predr.ag/blog/moving-and-reexporting-rust-type-can-be-major-breaking-change/), [2](https://predr.ag/blog/re-exporting-enum-with-type-alias-breaking-change-not-major/)), apparently not applicable here.

I think, technically this is a breaking change on nightly Rust, because my type alias doesn't have the second type parameter ([`A = Global`](https://docs.rs/sea-query/0.32.3/sea_query/types/struct.RcOrArc.html)). That type parameter is unstable, so it's inaccessible on stable Rust. It means that there's no breaking change for stable Rust users. But it also means that I can't forward it by default, because it wouldn't compile on stable. It would have to be gated behind a on opt-in `#[cfg(feature = "nightly")]`, or something like that. Which is still breaking, because it's opt-in.

However, I wouldn't worry about this, because this breakage is *really* niche:
- It affects only nightly Rust users who use custom allocators with `RcOrArc`.
- Which doesn't make sense, because the rest of `sea_query` assumes the global allocator and doesn't allow to change it. E.g., these two types wrap `RcOrArc` and hardcode it to use the global allocator:
    ```rust
    pub struct SeaRc<I>(pub(crate) RcOrArc<I>)
    where
        I: ?Sized;
    ```
    ```rust
    pub enum ColumnType {
        Array(RcOrArc<ColumnType>),
        // ..
    }
    ```
- So, this only affects nightly Rust users who use custom allocators with `RcOrArc`, but only for their own purposes, without passing custom-allocated `RcOrArc` into `sea_query`???

tyt2y3 (approved):
thank you. I trust your judgement!

tyt2y3:
sorry I forgot to click merge.

PR #860, opened 2025-01-29
fix: make AsEnum quoted to get the correct enum in postgres

MedHeikelBouzayene:
## PR Info

- Closes [issue link ](https://github.com/SeaQL/sea-orm/issues/2087)

## Bug Fixes

- Enums are always lowercased in queries as shown in the issue example

PR #880, opened 2025-04-16
fix: make AsEnum quoted to get the correct enum in postgres 

tyt2y3:
cont'd (#860)


tyt2y3:
Oh, I think it works
https://github.com/SeaQL/sea-orm/pull/2570

PR #882, opened 2025-04-29
Don't wrap strings in `Alias::new`

Expurple:
## PR Info

- Closes: none
- Dependencies: none
- Dependents: none

## Changes

(EDIT: outdated, see the up-to-date changelog in the comments below)

- Implemented `Iden` for `&str`, `&String` and `String`. Now you don't need to wrap strings into `Alias::new` where `T: IntoIden` is needed.

This was so noisy! Really drove me insane to hand-write every time. Just look at the diff that I was able to achieve.

Most of the diff is an automaned find-and-replace in VSCode (`Alias::new\(([^)]+)\)` -> `$1`, ignoring the changelog). Other than that, there are only two files with interesting changes:

- In `src/types.rs`, I implemented `Iden` for `&str`, `&String` and `String`. That's the actual change that causes strings to implicitly implement `IntoIden` and be accepted in methods that need `T: IntoIden`.

- This caused a type inference regression, because now these types have multiple `to_string` methods from multiple traits. I replaced the now-ambiguous `to_string` call in `src/prepare.rs`.

### Compatibility

The type inference issue is technically a breaking change, but it's commonly excluded from semver guarantees and doesn't require a major version bump. To quote [`cargo-semver-checks`](https://github.com/obi1kenobi/cargo-semver-checks/issues/5):

>  implementing an existing pub trait for an existing type
> - breaking because the trait's methods or associated types on that type may be ambiguous relative to those from traits implemented for that type in another crate (Rust for Rustaceans, chapter 3, pg. 52, "Trait Implementations")
> - inference breakage like this doesn't require a semver major bump, so this should be a warning, not an error

My mid-sized project (and the latest SeaORM 1.1.10 that it uses) compile fine with these changes.

There shouldn't be any other breaking changes.

`cargo semver-checks --release-type patch --baseline-rev master --all-features` doesn't report anything.

## Discussion: future breaking changes

Actually, I'd prefer a general blanket `impl Iden for T where T: AsRef<str>`. I think, it makes a lot of sense. Most `Iden` impls that I see in `sea_query` just `match Self -> &'static str` and then `write!` it. `impl AsRef<str>` could handle getting the right `&str`, and then the blanket `impl Iden for T where T: AsRef<str>` takes care of the common `s.write_str(self.as_ref()).unwrap()`. But this requires a new major version, so I just left a `TODO` in the code until then.

I'm kinda suspicious of `Iden::unquoted` that looks like `Display::fmt`, and `Iden::to_string` that obviously duplicates `ToString::to_string` (which is exactly what we would get by requiring `Iden: Display`!). But I'm not sure whether the `unquoted` string is semantically the same as `Display` and whether it could be replaced by it. Consider documenting this!

Also, I don't see the point of having an `Alias` newtype, instead of using strings directly. This area has very few docs and doesn't explain why `Alias` is necessary. So far, I'd like to just delete `Alias` in the next major version.

And it would be nice to hear something about https://github.com/SeaQL/sea-orm/discussions/2327. These `Into*` traits are very annoying and don't document why they're necessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -453,7 +453,7 @@ assert_eq!(
 
 ```rust
 let query = Query::select()
-    .expr(Func::cast_as("hello", Alias::new("MyType")))
+    .expr(Func::cast_as("hello", "MyType"))
     .to_owned();
 
 assert_eq!(
@@ -575,7 +575,7 @@ assert_eq!(
 let table = Table::alter()
     .table(Font::Table)
     .add_column(
-        ColumnDef::new(Alias::new("new_col"))
+        ColumnDef::new("new_col")
             .integer()
             .not_null()
             .default(100),
@@ -622,7 +622,7 @@ assert_eq!(
 
 ```rust
 let table = Table::rename()
-    .table(Font::Table, Alias::new("font_new"))
+    .table(Font::Table, "font_new")
     .to_owned();
 
 assert_eq!(
--- a/examples/diesel_mysql/src/main.rs
+++ b/examples/diesel_mysql/src/main.rs
@@ -197,7 +197,7 @@ fn main() {
 
     let query = Query::select()
         .from(Character::Table)
-        .expr_as(Func::count(Expr::col(Character::Id)), Alias::new("count"))
+        .expr_as(Func::count(Expr::col(Character::Id)), "count")
         .to_owned();
 
     print!("Count character: ");
--- a/examples/diesel_sqlite/src/main.rs
+++ b/examples/diesel_sqlite/src/main.rs
@@ -196,7 +196,7 @@ fn main() {
 
     let query = Query::select()
         .from(Character::Table)
-        .expr_as(Func::count(Expr::col(Character::Id)), Alias::new("count"))
+        .expr_as(Func::count(Expr::col(Character::Id)), "count")
         .to_owned();
 
     print!("Count character: ");
--- a/src/expr.rs
+++ b/src/expr.rs
@@ -48,12 +48,12 @@ pub enum SimpleExpr {
 /// ```no_run
 /// # use sea_query::*;
 /// #
-/// let expr = 1_i32.cast_as(Alias::new("REAL"));
+/// let expr = 1_i32.cast_as("REAL");
 ///
 /// let expr = Func::char_length("abc").eq(3_i32);
 ///
 /// let expr = Expr::current_date()
-///     .cast_as(Alias::new("TEXT"))
+///     .cast_as("TEXT")
 ///     .like("2024%");
 /// ```
 pub trait ExprTrait: Sized {
@@ -100,7 +100,7 @@ pub trait ExprTrait: Sized {
     /// let query = Query::insert()
     ///     .into_table(Char::Table)
     ///     .columns([Char::FontSize])
-    ///     .values_panic(["large".as_enum(Alias::new("FontSizeEnum"))])
+    ///     .values_panic(["large".as_enum("FontSizeEnum")])
     ///     .to_owned();
     ///
     /// assert_eq!(
@@ -205,7 +205,7 @@ pub trait ExprTrait: Sized {
     /// use sea_query::{tests_cfg::*, *};
     ///
     /// let query = Query::select()
-    ///     .expr("1".cast_as(Alias::new("integer")))
+    ///     .expr("1".cast_as("integer"))
     ///     .to_owned();
     ///
     /// assert_eq!(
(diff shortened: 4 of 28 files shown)


Expurple:
Regarding CI failures:

- `diesel-tests` fails due to some unrelated issue with `IpNetwork`.
- `cargo fmt` fails because the CI uses nightly `rustc 1.88.0-nightly (25cdf1f67 2025-04-28)`, while I use the latest stable (1.86.0) that doesn't format doctests. I'll reformat on nightly

tyt2y3:
thank you. although I have to say the original intention of not accepting strs is to prevent mistyping, and to encourage you to use `Derive(Iden)` on structs / enums. I get that you may have a different style of codebase. So I'd like to put this under a feature flag, wdyt?

Expurple:
> the original intention of not accepting strs is to prevent mistyping, and to encourage you to use `Derive(Iden)` on structs / enums. I get that you may have a different style of codebase.

I understand your sentiment, and I try to use `Iden` types for table and column names in my queries. But my queries (as well as `sea_query` codebase, as seen in the diff) still have plenty of and-hoc table/column aliases, or stored procedure calls, or casts between standard SQL types (for which you don't provide a Rust `Iden` representation, as far as I know). And you already have `Alias` to support these use cases, anyway. I don't see the point in making it artificially verbose. If anything, verbosity results in inattentive reading and MORE typos / bugs.

> May be we should only accept `&'static str`?

That's already the case in methods with `IntoIden` parameters, like `cast_as` / `expr_as` / `join_as` (these motivated me the most). `IntoIden` implies `'static` because of the bound on the blanket [`impl<T> IntoIden for T where T: Iden + 'static`](https://docs.rs/sea-query/0.32.4/sea_query/types/trait.IntoIden.html#impl-IntoIden-for-T). So, `Iden` is implemented for any `&str` (makes sense to me), but `IntoIden` is implemented only for `&'static str`.

If that's a hard requirement for merging, I could remove the impls for `String` and `&String`. But I don't see the point. `Alias` is the same thing, just more verbose. See my verbosity argument above.

> I'd like to put this under a feature flag, wdyt?

If that's a hard requirement for merging, I could do that. That would solve my personal issue in my codebase. However, I don't see the point in keeping `sea_query` more verbose by default*. It has many similar hurdles where the obvious code doesn't compile, and then you need to waste time and manually figure out the right wrappers / type conversions. I try to resolve these papercuts little by little and make the crate better for everyone.

\*That is, if you mean a non-`default` feature flag, where users don't get the impls by default. A `default` feature flag (with an option to manually disable it and disallow string identifiers) would be better, but I don't see much benefit in that either. I doubt that many people would find it useful and manually disable the impls.

Expurple:
On a side note, I'd like to come to an understanding and document this area:

> I'm kinda suspicious of `Iden::unquoted` that looks like `Display::fmt`, and `Iden::to_string` that obviously duplicates `ToString::to_string` (which is exactly what we would get by requiring `Iden: Display`!). But I'm not sure whether the `unquoted` string is semantically the same as `Display` and whether it could be replaced by it.

tyt2y3:
> I'm kinda suspicious of Iden::unquoted that looks like Display::fmt

yes, but I don't think it should be replaced. one could have a struct that impl Display differently from Iden

> but IntoIden is implemented only for &'static str

good to know

> If that's a hard requirement for merging, I could remove the impls for String and &String. But I don't see the point. Alias is the same thing, just more verbose. See my verbosity argument above.

how about pub-ing the internal field, i.e. `struct Alias(pub String)`, so we can do `Alias(my_string)`?

my thinking: we remove  String and &String for now, then feature flag not needed

Expurple:
> my thinking: we remove String and &String for now, then feature flag not needed

Ok, removed these. My most important use case are static string literals, so that's ok for me. The PR should now be mergeable for you too.

> how about pub-ing the internal field, i.e. struct Alias(pub String), so we can do Alias(my_string)?

I've decided not to do this for now, because the benefit is unclear. It's only beneficial for `String`s. But most of the time (confirmed by the diff with prior usages of `Alias::new` in `sea_query`), the calling code passes static string literals. Unlike `Alias::new("str")`, `Alias("str")` doesn't compile and still requires something like `Alias("str".into())`.

## Updated changelog

- Implemented `Iden` for `&str`. Now you don't need to wrap string literals into `Alias::new` where `T: IntoIden` is needed.

Expurple:
> yes, but I don't think it should be replaced. one could have a struct that impl Display differently from Iden. In some of my old code bases, I'd impl Iden manually on a struct `MyTable(i32)`

Thanks for the info, I added docs about this decision to the code.

tyt2y3 (approved):
thank you. we'll fix CI later and then make a release

tyt2y3:
I've removed some `Alias::new` from sea-orm as well

Expurple:
Cool. I too had a note to myself to do that after this change is released

tisonkun:
> we remove String and &String for now, then feature flag not needed

Implement `Iden` for String can be still worthwide because the current impl for `&str` must capture a reference so that it can't bring the ownership.

`Alias` can work as a workaround but the name indicates a different semantic so perhaps a direct `impl Iden for String` is still valuable.

Days before, I made:

```rust
#[derive(Debug, Clone)]
pub struct IdenStr<'a>(pub &'a str);

impl Iden for IdenStr<'_> {
    fn unquoted(&self, s: &mut dyn Write) {
        write!(s, "{}", self.0).unwrap();
    }
}

#[derive(Debug, Clone)]
pub struct IdenString(pub String);

impl Iden for IdenString {
    fn unquoted(&self, s: &mut dyn Write) {
        write!(s, "{}", self.0).unwrap();
    }
}
```

Now I can try to gradually remove `IdenStr`, but `IdenString` is still required for the reason above.

cc @tyt2y3 @Expurple

Expurple:
@tisonkun I agree with you and `impl Iden for String` makes sense to me too. I simply decided to not spend time defending it, and instead quickly merge the MVP that solves most of my issues. I'll happily open another PR for `String` if the maintainers change their mind

PR #888, opened 2025-05-22
Unify `Expr` and `SimpleExpr`

Huliiiiii:
## PR Info
Unify `Expr` and `SimpleExpr`

- Closes 

- Dependencies:
  - 

- Dependents:
  - 

## New Features

- [ ] 

## Bug Fixes

- [ ] 

## Breaking Changes

- [x] Add `non_exhaustive` to `Expr`
- [x] Rename `new_with_left` to `new`
- [x] Remove method `expr` because we can use `new`
- [x] Remove inherent methods of Expr from ExprTrait

## Changes

- [x] Replace generic with impl syntax on some method to simplify code

### DIFF
--- a/src/expr.rs
+++ b/src/expr.rs
@@ -1,38 +1,28 @@
 //! Building blocks of SQL statements.
 //!
-//! [`Expr`] representing the primitive building block in the expressions.
-//!
-//! [`SimpleExpr`] is the expression common among select fields, where clauses and many other places.
+//! [`Expr`] is the expression common among select fields, where clauses and many other places.
 
 use crate::{func::*, query::*, types::*, value::*};
 
-/// Helper to build a [`SimpleExpr`].
-#[derive(Debug, Clone)]
-pub struct Expr {
-    pub(crate) left: SimpleExpr,
-    pub(crate) right: Option<SimpleExpr>,
-    pub(crate) uopr: Option<UnOper>,
-    pub(crate) bopr: Option<BinOper>,
-}
-
 /// Represents a Simple Expression in SQL.
 ///
-/// [`SimpleExpr`] is a node in the expression tree and can represent identifiers, function calls,
+/// [`Expr`] is a node in the expression tree and can represent identifiers, function calls,
 /// various operators and sub-queries.
 #[derive(Debug, Clone, PartialEq)]
-pub enum SimpleExpr {
+#[non_exhaustive]
+pub enum Expr {
     Column(ColumnRef),
-    Tuple(Vec<SimpleExpr>),
-    Unary(UnOper, Box<SimpleExpr>),
+    Tuple(Vec<Expr>),
+    Unary(UnOper, Box<Expr>),
     FunctionCall(FunctionCall),
-    Binary(Box<SimpleExpr>, BinOper, Box<SimpleExpr>),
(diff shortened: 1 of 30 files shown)


src/expr.rs
Huliiiiii, line 1: Maybe the deprecated methods should be removed as well?

Expurple (commented):
Hi. First of all, thank you. It's great to see my idea come to life.

Before I properly review the diff, I want to address a few issues that I immediately see in the changelog, and make the PR easier to review:

1. Don't rename `SimpleExpr` -> `Expr` in this PR. Let's keep this PR small and focused on the effects of type unification. Add a `pub type Expr = SimpleExpr` for compatibility. I planned to do that anyway, to avoid needlessly breaking the users. It's good that you want to clean up the codebase and rename the usages of `SimpleExpr` into `Expr` (so that only one name is used consistently). But that should be a separate PR afterwards, IMO.
2. Removal of inherent methods from the one remaining type should be a separate PR afterwards.
3. `#[non_exhaustive]` is unrelated to the type unification and should be a separate PR.
4. IMO, these breaking changes aren't important to the health of the codebase:
    > - Rename `new_with_left` to `new`
    > - Remove method `expr` because we can use `new`
    > - Replace generic with impl syntax on some method to simplify code
    
    So, I would keep these features around to avoid needlessly breaking the users for little benefit.
    
    I agree that `new_with_left` and `expr` no longer make sense semantically. But I would mark them as deprecated instead.

Aside from being easier to review, small pull requests have another important benefit. I want to test each PR on `sea_orm` and my app, to see its effects and unexpecting problems before merging it. This means that for every `sea_query` PR, we need to open a corresponding `sea_orm` PR that deals with the breaking changes. And that PR could be large in itself. It's better to localize the impact and make it easier to deal with the breakage in `sea_orm`.

If you want to contribute multiple changes, you don't have to sit idle until this first PR is merged. You can open multiple PRs that depend on each other, and we'll iterate on these in parallel. You can expect quick responses from me. I'm interested in landing these beautiful breaking changes!

src/expr.rs
Expurple: Could you point out the exact methods? Or at least the file where I should search for `deprecated`?

Expurple:
In the description, I would reword from

> Close: https://github.com/SeaQL/sea-query/discussions/771

to

> Close: https://github.com/SeaQL/sea-query/discussions/771 (partially)
>
> Also tracked/explained in #795.

Your PR doesn't fully resolve #771, because that issue also needs another PR to remove methods from `sea_orm::ColumnTrait` and make `ColumnTrait: Into<SimpleExpr>` (implies `ColumnTrait: ExprTrait`, which will provide the replacement methods)

src/expr.rs
Expurple: If you mean the deprecated methods in `expr.rs` (`asterisk` and `table_asterisk`), IMO these aren't harmful to the users or the health of the codebase. These methods make sense to me, behave well and only take like 10 lines of code. I'm not even sure why they're deprecated in the first place!

Huliiiiii:
See #889 #890

Expurple:
Thank you for splitting. Are you going to update this PR to do something not included in the others? Or do you want to close this one now? I think, we should close, because the "unification" from the name is now done in #889

PR #889, opened 2025-05-22
Unify `Expr` and `SimpleExpr` as one type

Huliiiiii:
## PR Info

- Closes 

- Dependencies:
  - 

- Dependents:
  - 

## New Features

- [x] SimpleExpr now has all methods from Expr.

## Bug Fixes

- [ ] 

## Breaking Changes

- [x] Expr is now an alias of SimpleExpr.
  - How to migrate: if your code doesn't rely on these being two different types, everything should just work. But if you had impl Foo for Expr and impl Foo for SimpleExpr, your code won't compile and you need to delete one of the impls.

## Changes

- [x] Deprecate `SimpleExpr::expr` and add `SimpleExpr::new` which is more consistent with Rust conventions.

### DIFF
--- a/src/expr.rs
+++ b/src/expr.rs
@@ -1401,17 +1401,8 @@ where
 }
 
 impl SimpleExpr {
-    fn new(e: impl Into<Self>) -> Self {
-        e.into()
-    }
-
-    // TODO: since
-    #[deprecated(note = "Please use the [`SimpleExpr::new`] method")]
-    fn new_with_left<T>(left: T) -> Self
-    where
-        T: Into<Self>,
-    {
-        left.into()
+    pub fn new(expr: impl Into<Self>) -> Self {
+        expr.into()
     }
 
     #[deprecated(since = "0.29.0", note = "Please use the [`Asterisk`]")]
@@ -1606,7 +1597,7 @@ impl SimpleExpr {
     where
         V: Into<Value>,
     {
-        Self::new(v)
+        Self::from(v)
     }
 
     /// Wrap an expression to perform some operation on it later.
@@ -1679,7 +1670,7 @@ impl SimpleExpr {
     where
         T: Into<Self>,
     {
-        Self::new(expr)
+        expr.into()
     }
 
     /// Express a [`Value`], returning a [`SimpleExpr`].
--- a/src/expr.rs
+++ b/src/expr.rs
@@ -1747,7 +1747,7 @@ impl SimpleExpr {
     where
         T: Into<String>,
     {
-        SimpleExpr::Custom(s.into())
+        Self::Custom(s.into())
     }
 
     /// Express any custom expression with [`Value`]. Use this if your expression needs variables.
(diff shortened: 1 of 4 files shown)


src/expr.rs
```diff
@@ -1683,11 +1674,12 @@ impl Expr {
     /// );
     /// ```
     #[allow(clippy::self_named_constructors)]
+    #[deprecated(since = "0.32.0", note = "Please use the [`SimpleExpr::new`] method")]
```
Huliiiiii, line 1677: `since` needs to be updated

src/expr.rs
```diff
@@ -1407,18 +1400,18 @@ where
...
 }
 
-impl Expr {
+impl SimpleExpr {
+    fn new(e: impl Into<Self>) -> Self {
+        e.into()
+    }
+
+    // TODO: since
+    #[deprecated(note = "Please use the [`SimpleExpr::new`] method")]
```
Huliiiiii, line 1409: `since` needs to be updated

src/expr.rs
```diff
@@ -1401,17 +1401,8 @@ where
...
-    // TODO: since
-    #[deprecated(note = "Please use the [`SimpleExpr::new`] method")]
-    fn new_with_left<T>(left: T) -> Self
-    where
-        T: Into<Self>,
-    {
-        left.into()
+    pub fn new(expr: impl Into<Self>) -> Self {
+        expr.into()
     }
```
Expurple, line 1406: `new` seems the same as `from`. Is it really needed?

Expurple (commented):
Thank you. Everything seems good, `cargo test` passes without warnings.

But before merging, I think we need to check the breakage in `sea_orm`. Can you make a draft PR in `sea_orm` that imports this branch of `sea_query` and fixes the breakage?

Also, I'd reword the changelog a little:

> ## New Features
> 
> - [x] `SimpleExpr` now has all methods from `Expr`.

> ## Breaking Changes
> 
> - [x] `Expr` is now an alias of `SimpleExpr`.
>     - **How to migrate**: if your code doesn't rely on these being two different types, everything should just work. But if you had `impl Foo for Expr` and `impl Foo for SimpleExpr`, your code won't compile and you need to delete one of the impls.

I assume, we'll need some "migration guide" for `sea_query 1.0`. So, I'm going to write a "How to migrate" section for every breaking change. You can also do this in your future PRs.

src/expr.rs
Expurple: Hmm, I'm not sure now to properly deal with this. But the first release with your changes will probably be called `1.0.0-rc1`

src/expr.rs
Huliiiiii: I think this aligns better with Rust conventions.
`new`, `expr`, and `from` essentially do the same thing.

Expurple: I think, `new` is kinda unnecessary (we already have variant constructors addition to `expr` and `from`). But we can have it. I'll leave this up to maintainers to decide

Expurple:
Waiting for a CI run at https://github.com/SeaQL/sea-orm/pull/2604, to make sure that we actually fixed all the breakage there.

After that, this PR can be merged!

Expurple:
Apart from unrelated clippy complaints, the CI passes on both PRs 🎉

This PR can be merged now.

After that, we can finish with merging https://github.com/SeaQL/sea-orm/pull/2604#issuecomment-2911999416

PR #892, opened 2025-05-29
Replace "DynIden" with "IdenImpl"

Huliiiiii:
Background: https://github.com/SeaQL/sea-query/discussions/795#discussioncomment-13311660

## PR Info

- Closes 

- Dependencies:
  - 

- Dependents:
  - 

## Breaking Changes

- [x] Replaced `DynIden` with concrete `struct IdenImpl`.
  - **How to migrate**:
    - If you use the derive macro `Iden`, you don't need to do anything.
    - If you implemented trait `Iden` manually, you need to replace it with `impl From<YourType> for IdenImpl`.

- [x] Unlike `DynIden`, `IdemImpl` is compared purely by string. Same strings rendered from different Rust identifier types will now compare as equal.

- [x] Remove `SeaRc`.
  - **How to migrate:**
    - In SeaORM and SeaQuery, `SeaRc` was only used for `DynIden`. So, after you fix the errors around `DynIden`, you shouldn't have any errors left.
    - But if you used `SeaRc` for your own purposes outside of `DynIden`, there can be more errors. There, you can try replacing `SeaRc` with `RcOrArc` (which was the underlying implementation of `SeaRc` anyway).

## Changes


Expurple:
If you don't mind, I'll review when the CI passes. But feel free to ask if you need my review earlier

Huliiiiii:
Because it is too troublesome to fix the tests and examples, I also implemented the corresponding behavior for the derive macro Iden in this PR.

Huliiiiii:
It seems that all errors are caused by deprecation.

src/extension/mysql/column.rs
```diff
@@ -18,3 +18,13 @@ impl Iden for MySqlType {
...
+
+impl From<MySqlType> for IdenImpl {
+    fn from(value: MySqlType) -> Self {
+        Self::from(match value {
+            MySqlType::TinyBlob => "tinyblob",
+            MySqlType::MediumBlob => "mediumblob",
+            MySqlType::LongBlob => "longblob",
+        })
+    }
+}
```
Expurple, line 30: This duplicated `match` (with `impl Iden` right above) doesn't look right. I think, in this iteration (this PR that doesn't delete `trait Iden` yet) we should blanket `impl<T> Iden for T where T: Into<IdenImpl>`.

Also, your PR description should mention the plan to delete the trait, so that it's easier for Chris to review and understand the whole picture.

src/types.rs
```diff
@@ -70,14 +75,80 @@ iden_trait!(Send, Sync);
...
+    pub fn unquoted(&self, s: &mut dyn fmt::Write) {
+        write!(s, "{}", self.value).unwrap()
+    }
+}
+
+// #[deprecated(
+//     since = "1.0.0-rc.1",
+//     note = "This type alias is kept for compatibility purposes. It may be removed in the future."
+// )]
+pub type DynIden = IdenImpl;
```
Expurple, line 135: I'd add some docs with an explanation:

```rust
/// A legacy compatibility alias. 
///
/// Identifiers used to be reference-counted `dyn` trait objects,
/// and we had this shorter alias for readability.
///
/// Now, identifiers are eagerly "rendered" into a concrete struct.
/// You can just use the struct.
```

src/types.rs
```diff
@@ -70,14 +75,80 @@ iden_trait!(Send, Sync);
...
+
+#[derive(Debug, Clone)]
 #[repr(transparent)]
 pub struct SeaRc<I>(pub(crate) RcOrArc<I>)
```
Expurple, line 139: `SeaRc` is no longer used anywhere. I've checked the usage is SeaORM too, it's only used where `DynIden` is used.

Personally, I'd just delete `SeaRc` right in this same PR

src/types.rs
```diff
@@ -86,36 +157,33 @@ impl ops::Deref for SeaRc<dyn Iden> {
...
-        SeaRc(RcOrArc::new(i))
+impl ops::Deref for SeaRc<IdenImpl> {
+    type Target = IdenImpl;
+
+    fn deref(&self) -> &Self::Target {
+        ops::Deref::deref(&self.0)
     }
 }
 
 pub trait IntoIden {
```
Expurple, line 174:
```suggestion
/// A legacy compatibility trait.
///
/// We used to have it instead of `Into<IdenImpl>`.
///
/// New code should use `Into<IdenImpl>`.
pub trait IntoIden {
```

src/types.rs
```diff
@@ -86,36 +157,33 @@ impl ops::Deref for SeaRc<dyn Iden> {
...
     fn into_iden(self) -> DynIden;
 }
 
+impl<T> IntoIden for T
```
Expurple, line 178:
```suggestion
/// Provide compatibility with the existing code that uses `IntoIden`.
///
/// New code should use `Into<IdenImpl>`.
impl<T> IntoIden for T
```

Expurple (changes requested):
I love this!

src/types.rs
```diff
@@ -70,14 +75,80 @@ iden_trait!(Send, Sync);
 #[cfg(not(feature = "thread-safe"))]
 iden_trait!();
 
-pub type DynIden = SeaRc<dyn Iden>;
+#[derive(Debug, Clone, PartialEq)]
+pub struct IdenImpl {
```
Expurple, line 79:
```suggestion
/// An SQL identifier (string)
#[derive(Debug, Clone, PartialEq)]
pub struct IdenImpl {
```

src/extension/mysql/column.rs
Huliiiiii: After this PR, the Iden trait is no longer used in the actual code. I think we can remove these impls?

Huliiiiii: You're right. My current choice is mainly for compatibility, but since users will have to re-implement `Into<IdenImpl>` anyway, it makes sense to replace `Iden`.

Huliiiiii: I'm having some issues while cleaning up the code. Keeping `IdenStatic` requires a lot of (dirty) fixes. But since the current `Iden` is already static, I think we can remove IdenStatic as well?

sea-query-derive/tests/pass/renamed_unit_struct.rs
```diff
@@ -5,5 +5,5 @@ use sea_query::Iden;
 pub struct CustomName;
 
 fn main() {
-    assert_eq!(CustomName.to_string(), "another_name");
+    assert_eq!(Iden::from(CustomName).to_string(), "another_name");
```
Expurple, line 8: Hmm. Is this pattern common in the actual user code? Perhaps, we should find a way to reduce this breakage and verbosity. E.g., add a trait `ToIdenSting` that's blanket-implemented for `Into<Iden>`?

sea-query-derive/tests/pass-static/3_level_flatten.rs
```diff
@@ -1,61 +0,0 @@
-use sea_query::{Iden, IdenStatic};
-use strum::{EnumIter, IntoEnumIterator};
-
-#[derive(Copy, Clone, IdenStatic, EnumIter)]
```
Expurple, line 4: Sorry for noticing this only now, but perhaps we should keep `IdenStatic` derive macro for compatibility? Just as an alias that does the same thing as `#[derive(Iden)]`.

Hmm, perhaps it shouldn't even be an alias, and we should keep the `trait IdenStatic` and blaknet-`impl<T> From<T> for Iden where T: IdenStatic`? That would be nice for compatibility, and also for providing a compile-time guarantee that you can get a `&static str`

Huliiiiii: We could make the derive(IdenStatic) do nothing and implement as_str and AsRef<str> for Iden.

Expurple: On this line of code that I have reviewed, `#[derive(Iden)]` is used without `#[derive(Iden)]`. It used to implicitly generate `impl Iden` in addition to `impl IdenStatic`.

If we make `#[derive(IdenStatic)]` do nothing, then code like this breaks completely, instead of generating a partially-compatible `impl From<_> for Iden` that at least still allows the type to be passed into methods that accept `IntoIden`

Huliiiiii: Oh, I see — IdenStatic actually implements both Iden and IdenStatic, so maybe we just need to re-export the derive(Iden) macro as derive(IdenStatic)?

sea-query-derive/tests/pass/renamed_unit_struct.rs
Huliiiiii: That's the issues with this implementation:

1. The Iden trait doesn't provide a method to get a `&'static str`, so we can only implement
`impl<T> From<T> for DynIden where T: IdenStatic`.

2. Therefore, we cannot solve the problem of `derive(Iden)` because this trait can be implemented in a non-static way.

3. One option is to add a method to `Iden` trait that returns a `Cow`, but in practice, I think this isn't much different from implementing the `Iden` trait for types that implement `Into<IdenImpl>`.

Huliiiiii:
1. First, I don't think it's a good practice to name this method `to_string` as it conflicts with the `Display` trait method, which violates Rust conventions. I suggest renaming it to `to_iden_string` or `to_sql_string` instead. 
Second, the current issue is that most existing code calls the trait method `Iden::to_string` on enums or zero size types. There's no clean migration path for the existing code with this approach.
3. `SeaRc` can be kept, but it will be unused.
4. Yes, we can add `into_iden` for `Alias`, but I think a better approach would be to remove `Alias`, since the current changes implement `Into<IdenImpl>` for both `&'static str` and `String`.

tyt2y3:
I would probably even not put in the `deprecated` notice to avoid causing tons of compiler warnings in the first release.

> I'm also strongly in favor of adding an impl for String and keeping Alias only for compatibility: https://github.com/SeaQL/sea-query/pull/882. And two more users share their frustrations and hacks in that thread.

well, I guess my argument is that String is data and I don't want it to be confused as identifier. example:

```rust
let a = String::new("a");
let b = String::new("b");

`a.eq(b)` would result in `"a" = 'b'`
`b.eq(a)` would result in `"b" = 'a'`
```

one is quoted as identifier, one is quoted as literal.

Expurple:
> I would probably even not put in the `deprecated` notice to avoid causing tons of compiler warnings in the first release.

Yeah, agree. Definitely not in `1.0.0`.

> I guess my argument is that String is data and I don't want it to be confused as identifier

That's an interesting argument and we should explore it. But it needs a convincing example. Your pseudocode example is incorrect. The non-pseudocode

```rust
// [dependencies]
// sea-query = { version = "0.32.3", features = ["tests-cfg"] }

use sea_query::{ExprTrait, PostgresQueryBuilder, QueryBuilder};

fn main() {
    let a = String::from("a");
    let b = String::from("b");

    let mut sql = String::new();
    PostgresQueryBuilder.prepare_simple_expr(&a.clone().eq(&b), &mut sql);
    println!("{}", sql);

    let mut sql = String::new();
    PostgresQueryBuilder.prepare_simple_expr(&b.eq(&a), &mut sql);
    println!("{}", sql);
}
```

prints

```txt
'a' = 'b'
'b' = 'a'
```

And to me, that's obvious and expected. Identifiers don't even implement `ExprTrait`. It's impossible to call `eq` directly on an identifier string. `"a" = 'b'` can be produced only from something like `Expr::col(Alias::new(a)).eq(b)`. Or `Expr::col(a).eq(b)`, after we `impl Iden for String`. The latter looks great to me.

sea-query-derive/tests/pass/renamed_unit_struct.rs
tyt2y3: `impl<T> From<T> for DynIden where T: Iden` we can, but we have to always return `Owned`, which is not ideal.
yeah, `IdenStatic` is somehow added later than `Iden` that made things complicated.

> for example, to ignore quote as a way to bypass escaping
I think we ultimately have to reject this use case as part of the breaking change

Expurple: > I think we ultimately have to reject this use case as part of the breaking change

Ok, that means that we must delete the provided `Iden` methods, so that they can not be overridden [1]. We can keep their *calls* working [2] by adding a blanket-implemented `trait IdenExt`. This solution makes sense to me.

[1]: Whether we decide to keep `trait Iden` for some compatibility, or not.
[2]: Actually, the calls will break, but the users only need to `use sea_query::IdenExt` as the compiler suggests.

Expurple:
@Huliiiiii, we've chatted with Chris and decided to split we PR in order to proceed.

Initially, we want to merge these two small changes as separate PRs:

- `impl Iden for String`
- Alter the `PartialEq` impl to be safe and compare `DynIden`s by string

That's a blocker for now.

Huliiiiii:
> @Huliiiiii, we've chatted with Chris and decided to split the PR in order to proceed.
> 
> Initially, we want to merge these two small changes as separate PRs:
> 
> * `impl Iden for String`
> * Alter the `PartialEq` impl to be safe, ignore the types and compare `DynIden`s as strings
> 
> That's a blocker for now.

I haven’t had time to work on this lately. I’m totally fine if you wants to take over or use this as a base for further changes.

tyt2y3:
> I’m totally fine if you wants to take over or use this as a base for further changes.

I will try some experiments

PR #897, opened 2025-06-09
Replace the usage of SimpleExpr with Expr and deprecate SimpleExpr

Huliiiiii:
## PR Info

Make the code cleaner.

- Closes 

- Dependencies:
  - 

- Dependents:
  - 

## New Features

- [ ] 

## Bug Fixes

- [ ] 

## Breaking Changes

- [ ] 

## Changes

- [x] Replace the usage of SimpleExpr with Expr and deprecate SimpleExpr

### DIFF
--- a/src/expr.rs
+++ b/src/expr.rs
@@ -1,23 +1,24 @@
 //! Building blocks of SQL statements.
 //!
-//! [`SimpleExpr`] is an arbitrary, dynamically-typed SQL expression.
+//! [`Expr`] is an arbitrary, dynamically-typed SQL expression.
 //! It can be used in select fields, where clauses and many other places.
 //!
 //! [`ExprTrait`] provides "operator" methods for building expressions.
 
 use crate::{func::*, query::*, types::*, value::*};
 
-/// A legacy compatibility alias for [`SimpleExpr`].
+/// A legacy compatibility alias for [`Expr`].
 ///
 /// It used to be a separate type with constructor methods.
-/// Now you can call these constructors directly on [`SimpleExpr`].
-pub type Expr = SimpleExpr;
+/// Now you can call these constructors directly on [`Expr`].
+#[deprecated(since = "1.0.0-rc.2", note = "Use [`Expr`] instead.")]
+pub type SimpleExpr = Expr;
 
 /// An arbitrary, dynamically-typed SQL expression.
 ///
 /// It can be used in select fields, where clauses and many other places.
 ///
-/// More concreterly, under the hood [`SimpleExpr`]s can be:
+/// More concreterly, under the hood [`Expr`]s can be:
 ///
 /// - Rust values
 /// - SQL identifiers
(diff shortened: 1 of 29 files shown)


Huliiiiii:
I'm considering removing `Expr::new`, since we already have `Expr::from`.

https://github.com/SeaQL/sea-query/pull/889#discussion_r2104129084

As @Expurple mentioned here, I checked my codebase and found that using `Expr::from` makes the intent clearer. Moreover, accepting `Into<Self>` in the `new` method is not idiomatic in Rust.

src/expr.rs
```diff
@@ -1,23 +1,24 @@
...
 use crate::{func::*, query::*, types::*, value::*};
 
-/// A legacy compatibility alias for [`SimpleExpr`].
+/// A legacy compatibility alias for [`Expr`].
 ///
 /// It used to be a separate type with constructor methods.
-/// Now you can call these constructors directly on [`SimpleExpr`].
-pub type Expr = SimpleExpr;
+/// Now you can call these constructors directly on [`Expr`].
+#[deprecated(since = "1.0.0-rc.2", note = "Use [`Expr`] instead.")]
```
Expurple, line 14: I like the more obvious shorter naming and the consistency of usage. So I'd accept your PR in general.

But I think, we shouldn't deprecate `SimpleExpr`. My codebase uses it a lot, and that would be a lot of warnings for very little practical benefit.

At least, I definitely wouldn't deprecate it right in SeaQuery 1.0.0. This release already forces the users to go through the errors from the breaking changes and fix them. Throwing these noisy warnings on top of errors is not helpful.

src/expr.rs
```diff
@@ -1,23 +1,24 @@
...
+pub type SimpleExpr = Expr;
```
Expurple, line 15: You've got this doc-comment backwards now. The old `Expr` type was the one with constructors, not the old `SimpleExpr` type. Anyway, we can just drop the details about their original difference and meaning:

```rust
/// A legacy compatibility alias for [`Expr`].
///
/// These used to be two separate (but very similar) types.
```

src/expr.rs
```diff
@@ -29,27 +30,27 @@ pub type Expr = SimpleExpr;
...
+    AsEnum(DynIden, Box<Expr>),
     Case(Box<CaseStatement>),
     Constant(Value),
 }
 
 /// "Operator" methods for building expressions.
 ///
 /// Before `sea_query` 0.32.0 (`sea_orm` 1.1.1),
-/// these methods were awailable only on [`Expr`]/[`SimpleExpr`]
+/// these methods were awailable only on [`Expr`]/[`Expr`]
```
Expurple, line 53: This looks like a mechanical search-and-replace that makes less sense and needs to be rolled-back. I think, the reference to both old types is useful here

Expurple (changes requested):
Thank you!

Expurple (approved):
@tyt2y3, you can merge this

tyt2y3 (approved):
thank you guys

PR #899, opened 2025-06-14
pub field of Alias

tyt2y3:
So that we can now do `Alias(format!("abc{}"))`.


Expurple (approved):
Sure, why not. `Alias` is just a string anyway

Issue #901, opened 2025-07-04
Enable using DISTINCT with any Function


remiposo:
## Motivation

Since the method to set `FuncArgMod` is not exposed, it appears that `DISTINCT` cannot be specified for functions other than `COUNT` (and `ARRAY_AGG`).
If I'm mistaken, it would be helpful if you could tell me how to specify it.

## Proposed Solutions

Change [`arg_with`](https://github.com/SeaQL/sea-query/blob/bab54410627b57c0ee1028f48f6dfc3a4108f527/src/func.rs#L68) from `pub(crate)` to `pub`.

PR #902, opened 2025-07-04
Expose arg_with

remiposo:
## PR Info

- Closes https://github.com/SeaQL/sea-query/issues/901

## Changes

- Changed `arg_with` to `pub` to specify `FuncArgMod`

### DIFF
--- a/src/func.rs
+++ b/src/func.rs
@@ -65,7 +65,7 @@ impl FunctionCall {
         self.arg_with(arg, Default::default())
     }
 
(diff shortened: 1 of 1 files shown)


tyt2y3:
what do you need this method for? can you instead add a method similar to `fn array_agg_distinct`?

tyt2y3:
sorry for the delay, if `count_distinct` is what you need, let's just add that specifically!

PR #905, opened 2025-07-07
Remove `into_sub_query_statement` from trait

tyt2y3:


Expurple (commented):
I like the shorter `.into()` calls, we can keep that part of the diff.

But I'd also keep `into_sub_query_statement` in the trait for now, to stay compatible. We can deprecate it later in `2.x.x`, give plenty of time to migrate, and then remove in `3.0.0`. `2.0.0` already requires to manually fix a lot of things at once.

I also wanted to do something similar with `Into*` traits. Replace them with `Into<*>` in our APIs, but also still keep the original traits as compatibility "aliases" (`trait IntoFoo: Into<Foo>`, `impl<T> IntoFoo for T where T: Into<Foo>`). Then deprecate later in `2.x.x` and remove in `3.0.0`.

Expurple:
1, 0, 0, 0 matches. In the last comment, I went for `Into\w+` first, found all the used traits, and then checked only their methods. But it's hard to tell how representative my project is

tyt2y3:
how about this? will be easy to deprecate and remove in the future.

Expurple (approved):
I like this. A default method is good.

> into_sub_query_statement is the worst because it's actually piggy-backing on the QueryStatementBuilder trait.

Yeah, I agree.

PR #909, opened 2025-07-12
New iden paradigm

tyt2y3:
As discussed, `Arc`, `vtable` tricks are removed.

The biggest behaviour change is late binding to early binding. previously the Iden is only rendered when the AST is being serialized, now that we pre-render it as we store that into the AST.

For majority of usecases in SeaORM though, since they're `'static str` anyway, there should be no difference.

It is still possible to have `struct MyIden(String)` with some dynamic behaviour, it's just that constructing new String on the fly is no longer possible.

### DIFF
--- a/sea-query-derive/Cargo.toml
+++ b/sea-query-derive/Cargo.toml
@@ -2,7 +2,7 @@
 name = "sea-query-derive"
 version = "0.4.3"
 authors = [ "Follpvosten <wolfi@karpador.xyz>", "Rene Leveille <rene@nestingsafe.com>" ]
-edition = "2024"
+edition = "2021"
 description = "Derive macro for sea-query's Iden trait"
 license = "MIT OR Apache-2.0"
 documentation = "https://docs.rs/sea-query"
--- a/src/types.rs
+++ b/src/types.rs
@@ -1,7 +1,7 @@
 //! Base types used throughout sea-query.
 
 use crate::{FunctionCall, ValueTuple, Values, expr::*, query::*};
-use std::{borrow::Cow, fmt, ops};
+use std::borrow::Cow;
 
 #[cfg(feature = "backend-postgres")]
 use crate::extension::postgres::PgBinOper;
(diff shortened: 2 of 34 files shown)


tyt2y3:
there is no way to read this code diff other than going commit by commit. every commit is compilable.

but summary of changes:

1. change unquoted() signature
image

2. change quoted() signature
```
fn quoted(&self) -> Cow<'static, str> {
```

3. removed prepare() from Iden trait (moved into builder backend)
4. DynIden is just a Cow, SeaRc is just a helper
5. impl IntoIden for String (most asked, I guess)

sea-query-derive/Cargo.toml
```diff
@@ -2,7 +2,7 @@
 name = "sea-query-derive"
 version = "0.4.3"
 authors = [ "Follpvosten <wolfi@karpador.xyz>", "Rene Leveille <rene@nestingsafe.com>" ]
-edition = "2024"
+edition = "2021"
```
Expurple, line 5: What's the reason for this?

src/types.rs
```diff
@@ -24,91 +24,69 @@ pub type RcOrArc<T> = std::sync::Arc<T>;
...
+    fn unquoted(&self) -> &str;
+}
+
+/// Identifier statically known at compile-time.
+pub trait IdenStatic: Iden + Copy + 'static {
+    fn as_str(&self) -> &'static str;
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct DynIden(pub(crate) Cow<'static, str>);
```
Expurple, line 72:
```suggestion
/// A prepared (quoted) identifier string.
///
/// The naming is legacy and kept for compatibility.
/// This used to be an alias for a `dyn Iden` object that's lazily rendered later.
///
/// Nowadays, it's an eagerly-rendered string.
/// Most identifiers are static strings that aren't "rendered" at runtime anyway.
#[derive(Debug, Clone, PartialEq, Eq)]
pub struct DynIden(pub(crate) Cow<'static, str>);
```

src/types.rs
```diff
@@ -24,91 +24,69 @@ pub type RcOrArc<T> = std::sync::Arc<T>;
...
 
 #[derive(Debug)]
-#[repr(transparent)]
-pub struct SeaRc<I>(pub(crate) RcOrArc<I>)
-where
-    I: ?Sized;
+pub struct SeaRc;
```
Expurple, line 75:
```suggestion
/// A legacy namespace for compatibility.
/// 
/// It's needed, so that most existing [`SeaRc::new`][SeaRc::new] calls keep working.
///
/// This used to be an actual type
/// (a reference-counted pointer with special impls for `dyn Iden` contents).
/// It's not needed anymore.
#[derive(Debug)]
pub struct SeaRc;
```

src/types.rs
```diff
@@ -24,91 +24,69 @@ pub type RcOrArc<T> = std::sync::Arc<T>;
...
-    fn eq(&self, other: &Self) -> bool {
-        let (self_vtable, other_vtable) = unsafe {
-            let (_, self_vtable) = mem::transmute::<&dyn Iden, (usize, usize)>(&*self.0);
-            let (_, other_vtable) = mem::transmute::<&dyn Iden, (usize, usize)>(&*other.0);
-            (self_vtable, other_vtable)
-        };
-        self_vtable == other_vtable && self.to_string() == other.to_string()
+impl SeaRc {
+    #[allow(clippy::new_ret_no_self)]
+    pub fn new<I>(i: I) -> DynIden
```
Expurple, line 79:
```suggestion
    /// A legacy method, kept for compatibility.
    ///
    /// Nowadays, instead of wrapping an `Iden` object,
    /// it eagerly "renders" it into a string and then drops the object.
    ///
    /// Note that most `Iden`s are statically known
    /// and their representations aren't actually "rendered" and allocated at runtime.
    #[allow(clippy::new_ret_no_self)]
    pub fn new<I>(i: I) -> DynIden
```

src/types.rs
```diff
@@ -630,33 +601,71 @@ impl Alias {
...
-        self.0.as_str().unquoted(s);
+impl IntoIden for String {
+    fn into_iden(self) -> DynIden {
+        DynIden(Cow::Owned(self))
     }
 }
 
 /// The "base" `impl` for writing arbitrary "raw" strings as identifiers.
 ///
 /// Reused for other string-like types.
```
Expurple, line 618:
```suggestion
```

The comment is out of date. `Alias` and `String` impls no longer delegate to the `str` impl

Expurple (approved):
I love this

sea-query-derive/Cargo.toml
tyt2y3: I don't know which `trybuild` version started supporting edition 2024. so I will have to specify a recent enough version.

tyt2y3:
special thanks to @Huliiiiii for bringing forward the proposal!

PR #915, opened 2025-07-25
Make `ConditionExpression` private

Expurple:
Previously described in https://github.com/SeaQL/sea-query/discussions/795.

## New Features

- [x] `impl From<Expr> for Condition`

## Breaking Changes

- [x] Made `ConditionExpression` private.
- [x] Replaced `Into<ConditionExpression>` with `Into<Condition>` in `Condition::add` and `Condition::add_option`.

### DIFF
--- a/src/query/condition.rs
+++ b/src/query/condition.rs
@@ -20,11 +20,14 @@ pub trait IntoCondition {
 
 pub type Cond = Condition;
 
-/// Represents anything that can be passed to an [`Condition::any`] or [`Condition::all`]'s [`Condition::add`] method.
+/// An internal representation of conditions.
+/// May be refactored away in the future if we can get our head around it.
 ///
-/// The arguments are automatically converted to the right enum.
+/// It used to be in the public interface of [`Condition::add`] and [`Condition::add_option`],
+/// in order to accept anything resembling a condition or an expression.
+/// Nowadays, we achieve that with traits.
 #[derive(Debug, Clone, PartialEq)]
-pub enum ConditionExpression {
+pub(crate) enum ConditionExpression {
     Condition(Condition),
     Expr(Expr),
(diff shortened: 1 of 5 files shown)


src/query/condition.rs
```diff
@@ -68,16 +71,11 @@ impl Condition {
...
-            // Skip the junction if there is only one.
-            if c.conditions.len() == 1 && !c.negate {
-                expr = c.conditions.pop().unwrap();
-            }
-        }
-        self.conditions.push(expr);
+        let condition: Condition = condition.into();
+        let condition: ConditionExpression = condition.into();
+        self.conditions.push(condition);
         self
```
Expurple, line 79: I removed that special case, and that had no impact on the tests at all. It's not even related to my other changes. If I just comment out

```rust
if c.conditions.len() == 1 && !c.negate {
    expr = c.conditions.pop().unwrap();
}
```

on `master`, the tests still pass.

src/query/condition.rs
```diff
@@ -312,6 +310,26 @@ impl From<Expr> for ConditionExpression {
...
+
+impl From<ConditionExpression> for Condition {
+    fn from(ce: ConditionExpression) -> Self {
+        Condition {
+            negate: false,
+            condition_type: ConditionType::All,
+            conditions: vec![ce],
+        }
+    }
+}
```
Expurple, line 331: `ConditionExpression` is now private, so we don't need to include this in the changelog

src/query/condition.rs
tyt2y3: iirc it's to reduce the extra parenthesis `()` around conditions. but since then there's much smarter logic around this.

Expurple:
I love it when the code makes sense :)

PR #917, opened 2025-07-27
Remove Expr::new

Huliiiiii:
https://github.com/SeaQL/sea-query/pull/897#issuecomment-2954569983

> I'm considering removing Expr::new, since we already have Expr::from.
> As @Expurple mentioned here, I checked my codebase and found that using Expr::from makes the intent clearer. Moreover, accepting Into<Self> in the new method is not idiomatic in Rust.

### DIFF
--- a/src/expr.rs
+++ b/src/expr.rs
@@ -1523,10 +1523,6 @@ where
 }
 
 impl Expr {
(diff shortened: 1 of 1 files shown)


Huliiiiii:
But this was added in 1.0.0-rc.1.

PR #922, opened 2025-07-28
Accept `IntoColumnRef` in `cast_as_quoted`

### LINKED ISSUE #827: Incorrect Quoting of Schema-Qualified Enums in `cast_as_quoted`
## Description

When using the `cast_as_quoted` function to cast schema-qualified enums (e.g., `schema_name.enum_name`), the generated SQL incorrectly quotes the entire identifier as `"schema_name.enum_name"`. Instead, each part of the schema and enum should be quoted separately, like `"schema_name"."enum_name"`, to ensure correctness.

## Steps to Reproduce

1. Define a schema-qualified enum:
   ```sql
   CREATE TYPE schema_name.enum_name AS ENUM ('value1', 'value2');
   ```
2. Use SeaQuery to cast a value using `cast_as_quoted`:
   ```rust
   use sea_query::{tests_cfg::*, *};

   let query = Query::select()
       .expr(Func::cast_as_quoted(
           "some_value",
           Alias::new("schema_name.enum_name"),
           '"'.into(),
       ))
       .to_owned();
   ```
3. Generate the SQL for PostgreSQL:
   ```rust
   assert_eq!(
       query.to_string(PostgresQueryBuilder),
       r#"SELECT CAST('some_value' AS "schema_name.enum_name")"#
   );
   ```

### Expected Behavior

The SQL should be generated as:
```sql
SELECT CAST('some_value' AS "schema_name"."enum_name")
```

### Actual Behavior

The SQL is incorrectly generated as:
```sql
SELECT CAST('some_value' AS "schema_name.enum_name")
```

### Reproduces How Often

Always.

## Versions

- SeaQuery: 0.31.1
- PostgreSQL: Irrelevant
- OS: Irrelevant

## Additional Information

The current `cast_as_quoted` function does not handle schema-qualified names correctly. Below is the fix to ensure that both schema and enum names are properly quoted.

```diff
    pub fn cast_as_quoted<V, I>(expr: V, iden: I, q: Quote) -> FunctionCall
    where
        V: Into<SimpleExpr>,
        I: IntoIden,
    {
        let expr: SimpleExpr = expr.into();
        let mut quoted_type = String::new();
-       iden.into_iden().prepare(&mut quoted_type, q);
+       let parts: Vec<&str> = iden.into_iden().to_string().split('.').collect();
+       let quoted_type = parts.iter().map(|s| format!("\"{}\"", s)).collect::<Vec<_>>().join(".");
        FunctionCall::new(Function::Cast)
            .arg(expr.binary(BinOper::As, Expr::cust(quoted_type.as_str())))
    }
```

This fix ensures that schema-qualified enums are quoted correctly by splitting the identifier on the dot and quoting each part separately.

Expurple:
As discussed here: https://github.com/SeaQL/sea-query/issues/827#issuecomment-3125338008

Merge after #921. Marking as draft until then. The first commit comes as a dependency from that PR. If you want to review the diff before merging that PR, review only the second commit.

## PR Info

- Closes #827

- Dependencies:
  - #921

## Changes

- [x] Accept `IntoColumnRef` as the type name in `cast_as_quoted`.

This is backwards-compatible because there's an

```rust
impl<T> IntoColumnRef for T
where
    T: IntoIden
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## 1.0.0 - pending
 
+### Enhancements
+
+* Removed unnecessary `'static` bounds from type signatures.
+
 ### New features
 
 * Unify `Expr` and `SimpleExpr` as one type. `SimpleExpr` is kept as an alias of `Expr`, but they can now be used interchangably. There may be a few compile
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Enhancements
 
 * Removed unnecessary `'static` bounds from type signatures.
+* `cast_as_quoted` now allows you to [qualify the type
+  name](https://github.com/SeaQL/sea-query/issues/827).
 
 ### New features
 
--- a/src/func.rs
+++ b/src/func.rs
@@ -537,6 +537,8 @@ impl Func {
 
     /// Call `CAST` function with a case-sensitive custom type.
     ///
+    /// Type can be qualified with a schema name.
+    ///
     /// # Examples
     ///
     /// ```
@@ -558,15 +560,34 @@ impl Func {
     ///     query.to_string(SqliteQueryBuilder),
     ///     r#"SELECT CAST('hello' AS "MyType")"#
     /// );
+    ///
+    /// // Also works with a schema-qualified type name:
+    ///
+    /// let query = Query::select()
+    ///     .expr(Func::cast_as_quoted("hello", ("MySchema", "MyType")))
+    ///     .to_owned();
+    ///
+    /// assert_eq!(
+    ///     query.to_string(MysqlQueryBuilder),
+    ///     r#"SELECT CAST('hello' AS `MySchema`.`MyType`)"#
+    /// );
+    /// assert_eq!(
+    ///     query.to_string(PostgresQueryBuilder),
+    ///     r#"SELECT CAST('hello' AS "MySchema"."MyType")"#
+    /// );
+    /// assert_eq!(
+    ///     query.to_string(SqliteQueryBuilder),
+    ///     r#"SELECT CAST('hello' AS "MySchema"."MyType")"#
+    /// );
     /// ```
-    pub fn cast_as_quoted<V, I>(expr: V, iden: I) -> FunctionCall
+    pub fn cast_as_quoted<V, I>(expr: V, col: I) -> FunctionCall
     where
         V: Into<Expr>,
-        I: IntoIden,
+        I: IntoColumnRef,
     {
         let expr: Expr = expr.into();
         FunctionCall::new(Function::Cast)
-            .arg(expr.binary(BinOper::As, Expr::TypeName(iden.into_iden())))
+            .arg(expr.binary(BinOper::As, Expr::Column(col.into_column_ref())))
     }
 
     /// Call `COALESCE` function.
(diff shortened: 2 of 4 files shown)


Expurple:
Hold up. Why did I rename the parameter to `col`? It's a type name! `ColumnRef` doesn't make sense here. `Asterisk` and `TableAsterisk` aren't valid type names! That's exactly what I was talking about here: https://github.com/SeaQL/sea-query/discussions/795#discussioncomment-13917480.

I need to implement and merge `MaybeQualified`, and then use that here instead of `ColumnRef`.

**Don't merge this PR as is.**

tyt2y3:
I see your point. it works but is kind of wrong in a type system sense. also I think this parameter can appear in other places as well, e.g. ALTER TYPE

Expurple:
@tyt2y3 @Huliiiiii

Rebased on top of #953 as an experiment. The first commit comes from that branch. The second commit is the actual content of this PR.

The `cast_as_quoted` API finally seems right. But there are still many things that I don't like:

1. I'd like to get rid of `Expr::TypeName` altogether. AFAIK, a type name is not a valid expression. We should model the SQL grammar more faithfully.

2. I'm not sure what's the deal with `Expr::AsEnum`. Why is it specifically about enums? This needs to be documented.

3. While browsing `AsEnum` references, I found this code in the Postgres backend:

    ```rust
    fn prepare_simple_expr(&self, simple_expr: &Expr, sql: &mut dyn SqlWriter) {
        match simple_expr {
            Expr::AsEnum(type_name, expr) => {
                write!(sql, "CAST(").unwrap();
                self.prepare_simple_expr_common(expr, sql);
                let q = self.quote();
                let type_name = type_name.to_string();
                let (ty, sfx) = if type_name.ends_with("[]") {
                    (&type_name[..type_name.len() - 2], "[]")
                } else {
                    (type_name.as_str(), "")
                };
                write!(sql, " AS {}{}{}{})", q.left(), ty, q.right(), sfx).unwrap();
            }
            _ => QueryBuilder::prepare_simple_expr_common(self, simple_expr, sql),
        }
    }
    ```

    `type_name` here is a `DynIden`, but the code also handles `[]` as part of the type name. Valid identifiers don't include `[]`! If we want to support that, we need to model `TypeName` / `TypeRef` as something richer than `(Option<SchemaName>, DynIden)` (which it is in this version of the PR)

Expurple:
I've just googled type names in Posgtres. Boy, is it complicated even without `[]`. They can also span several words and have `(size)` / `(precision)`: https://www.postgresql.org/docs/current/datatype.html#DATATYPE-TABLE

I think, we need to agree what exactly a `DynIden` is and what it can and can not contain. Based on the naming, I'd prefer the clarity of it being just a single identifier, and type names being a separate, more complex type

Huliiiiii:
`Expr::SubQuery`, `Expr::AsEnum`, and `Expr::Case` were all added as workarounds to support certain kinds of expressions that were previously unsupported, without breaking compatibility.

As I mentioned here: https://github.com/SeaQL/sea-query/discussions/926
There are many SQL expressions with unique structures, we need a more general representation to support them properly.

In Rust, one common approach to representing an AST—especially when aiming to reduce allocations—is to store all nodes in a Vec (serving as an arena) and use an enum whose variants hold typed handles (e.g., usize) pointing into that arena.

But for 1.0, this change is too large.

tyt2y3:
Yeah it's complicated I don't think a universal solution is easy. I would just stick with the low hanging fruit for now, which basically is the draft PR you have.

for context, `AsEnum` was added to support ActiveEnum in SeaORM, and the problem was:
1. we can't quote primitive type names `"integer"`
2. we have to quote  custom types (including enum) `"my_enum"`
3. however if we want to cast to `my_enum[]` the correct syntax is `"my_enum"[]`

PR #923, opened 2025-07-28
Remove `SeaRc::new` from docs and internal code

Expurple:
`SeaRc::new` is a compatibility hack that's going to be deprecated in the future. Let's start moving away from it.

### DIFF
--- a/src/query/select.rs
+++ b/src/query/select.rs
@@ -606,11 +606,9 @@ impl SelectStatement {
     /// ```
     /// use sea_query::{tests_cfg::*, *};
     ///
-    /// let alias: String = "C".into();
-    ///
     /// let query = Query::select()
     ///     .from(Char::Table)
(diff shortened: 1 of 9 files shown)


src/query/select.rs
```diff
@@ -1586,15 +1580,14 @@ impl SelectStatement {
     /// ```
     /// use sea_query::{tests_cfg::*, audit::*, *};
     ///
-    /// let sub_glyph: DynIden = SeaRc::new("sub_glyph");
```
tyt2y3, line 1589: I wonder what's your view on `static sub_glyph: &str = "sub_glyph"?

Expurple: Do you want me to change that back to a variable?

tyt2y3 (approved):

src/query/select.rs
tyt2y3: it's okay. but better to have multiple variants through the docs example!

PR #925, opened 2025-07-28
Unbox values

Huliiiiii:
## PR Info

Unbox values, benchmarks show a 10–15% performance improvement in query construction

- Closes #863

- Dependencies:
  - 

- Dependents:
  - 

## New Features

- [ ] 

## Bug Fixes

- [ ] 

## Breaking Changes

- [x] values inside the `Value` enum are no longer boxed.

## Changes

- [ ]

### DIFF
--- a/tests/size.rs
+++ b/tests/size.rs
@@ -1,11 +1,9 @@
 use sea_query::Value;
 
-/// If this check fails, you need to run `benches/value.rs`
-/// and see if the increased `Value` size hurts performance.
+/// This test is to check if the size of [`Value`] exceeds the limit.
+/// If the size exceeds the limit, you should try boxing the variant, update and run the benchmark of [`Value`] to compare performance.
 ///
-/// If it does, box the big variant.
-///
-/// If it doesn't, bump the constant here in the test.
+/// If the boxed variant causes a greater performance loss, update the size limit instead.
 #[test]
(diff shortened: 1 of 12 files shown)


Expurple:
I'd also add a comment right there, next to the assertion. To document our approach and the existence of the benchmark

Huliiiiii:
Based on the benchmark results, I didn't see any performance impact from the size of Value (24 vs 40). I think boxing the type on the stack would have a larger performance overhead.

tyt2y3:
@Huliiiiii I agree with you. memory in-direction is almost always worse. but I think we should put a guard in place to prevent it growing unchecked.

tests/size.rs
```diff
@@ -0,0 +1,9 @@
+use sea_query::Value;
+
+#[test]
+fn value_size() {
```
Expurple, line 4:
```suggestion
/// If this check fails, you need to run `benches/value.rs`
/// and see if the increased `Value` size hurts performance.
///
/// If it does, box the big variant.
///
/// If it doesn't, bump the constant here in the test.
#[test]
fn value_size() {
```

Expurple (commented):
Still needs a changelog entry. LGTM otherwise

tyt2y3:
I'll probably make some small tweaks

PR #927, opened 2025-07-29
Rework `ColumnRef` and `TableRef`

Expurple:
Extracted the `foo.bar.baz` qualification logic that was duplicated across several variants of `ColumnRef` and `TableRef`.

I need this to implement `schema.type`: #827, #922

I'll also need this to implement `schema.function` later: https://github.com/SeaQL/sea-query/discussions/795#discussioncomment-13917480

There are also immediate benefits. IMO, `ColumnRef` and `TableRef` (and their pattern-matching) look considerably cleaner now. The `database.` prefix is guaranteed to be supported everywhere where it may be needed.

The detailed changelog is in the diff. I fully preserved (and documented) the existing codegen behavior, as evidenced by unchanged output in tests. If this documented behavior looks suspicious (it kinda does to me), let's fix that later in a separate PR. Here, the only changes in tests are the refactoring that's needed for the new variants to compile.

This is a relatively big breaking change for those users who dig within the internals of `ColumnRef`, `TableRef`, and `SchemaTable`. But that's very low-level and should be rare. My work project never does that, despite depending on SeaORM/SeaQuery/SQLx very closely and frequently getting generic or low-level.

## PR Info

- Dependents:
  - https://github.com/SeaQL/sea-query/pull/922

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -1,7 +1,7 @@
 //! Base types used throughout sea-query.
 
 use crate::{FunctionCall, ValueTuple, Values, expr::*, query::*};
-use std::borrow::Cow;
+use std::{borrow::Cow, fmt::Debug, iter::Flatten};
 
 #[cfg(feature = "backend-postgres")]
 use crate::extension::postgres::PgBinOper;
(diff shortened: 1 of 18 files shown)


src/types.rs
```diff
@@ -161,18 +211,8 @@ pub trait IntoColumnRef {
...
-    /// Table identifier with database and schema prefix
-    DatabaseSchemaTable(DynIden, DynIden, DynIden),
-    /// Table identifier with alias
-    TableAlias(DynIden, DynIden),
-    /// Table identifier with schema prefix and alias
-    SchemaTableAlias(DynIden, DynIden, DynIden),
-    /// Table identifier with database and schema prefix and alias
-    DatabaseSchemaTableAlias(DynIden, DynIden, DynIden, DynIden),
+    /// A table identifier. Potentially qualified. Potentially remaned using an alias
+    Table(TableName, Option<DynIden>),
```
Expurple, line 215: Ideally, I'd use named fields for clarity (`alias: Option<DynIden>`).

But I chose to stay consistent with the other existing variants and not touch those. SeaQuery 1.0 breaks a lot of things already

Huliiiiii:
Oh, I forgot about that.

tyt2y3 (approved):

Issue #259, opened 2022-02-09
SQLite insert performance


nhenezi:
I'm trying to get `sea_query` to be as performant as native SQLite prepared statement as possible.  I've only tested `TEXT` columns and here's the fastest I've managed to do it:

```rust
    let mut q = sea_query::Query::insert()
        .into_table(S::S("test_table".to_string()))
        .columns(vec![
            S::S("col1".to_string()),
            S::S("col2".to_string()),
            S::S("col3".to_string()),
            S::S("col4".to_string()),
            S::S("col5".to_string()),
            S::S("col6".to_string()),
            S::S("col7".to_string()),
            S::S("col8".to_string()),
            S::S("col9".to_string()),
            S::S("col10".to_string()),

        ])
        .to_owned();
    q.values_panic(data[0].clone());
    let (sql, _) = q.build(SqliteQueryBuilder);
    let mut prepared_stmt = conn.prepare(&sql).unwrap();
    let start = Instant::now();
    conn.execute("BEGIN TRANSACTION;", []).unwrap();
    for d in data {
        let rq_vals = RusqliteValues::from(Values(d));
        prepared_stmt.execute(rq_vals.as_params().as_slice()).unwrap();

    }
    conn.execute("COMMIT;", []).unwrap();

```
by using the fact that I know that all values have the same structure so I use only the first one to construct a prepared statement that is utilized during the insertion - that seemed to produce the equivalent to the fastest SQLite insertion and consequentially fastest `sea_query` insertions:

```
Using SQLite prepared statement
1000000 entries inserted in ~1.149 sec. ~870000 inserts/sec
Using sea_query with guard and prepared statement
1000000 entries inserted in ~1.541 sec. ~648000 inserts/sec
```

It's still 25% slower which I would like to bring down to 5-10% if possible. A few different approaches and a more extensive set of experiments you can find at https://gitlab.com/nhenezi/rust-sea-query-perf-test/-/blob/master/src/main.rs .

I'm suspecting that the [`.as_params()`](https://github.com/SeaQL/sea-query/blob/master/src/driver/rusqlite.rs#L19) is the bottleneck, as results gets worse the more columns I add to the performance test and `.as_params()` has to iterate over all values and convert them to something that SQLite understands.

I see two approaches that I can take here, but neither one is doable in my case:
  -  parallelization approach to serializing to SQLite where I wouldn't use `.as_params()` but rather in a parallelized manner convert every `Value` individually to its SQLite representation
[...]

PR #931, opened 2025-07-30
Value: box BigDecimal and Array

tyt2y3:
These two types are already internally heap-allocated.

And boxing these two would bring down Value to 32 bytes.

I hope this is fair trade off.


tyt2y3:
32 is a magic power-of-two number. so that most value variants don't get pessemisied by `Array` and `BigInt`. they are using heap allocation anyway, so the penalty is relatively smaller. 

if you are concerned about ergonomics, I think we can add a few `From` impl.

Expurple:
Yeah, `From` impls need attention, regardless of boxing. I have some personal TODOs regarding those painful conversions into `Value`. I'll sort it out and publish later.

> 32 is a magic power-of-two number that's more cache line friendly.

Sure, but how important is that in practice?  I could only find a single open performance-related issue across SeaORM and SeaQuery repos: https://github.com/SeaQL/sea-query/issues/259. On the other hand, ergonomics has always been a big issue for me and my team.

I'm vary of micro-optimizing without representative benchmarks. And I doubt that `Value`s are moved or accessed so often that it dominates a separate allocation/deallocation for *every* `Value` (of a boxed variant). I'd assume that, most of the time, `Value`s are constucted close to the query, then moved into some internal `Vec`, then *sequentially* iterated for serialization, and that's it. Cache lines shouldn't matter for sequential iteration over a `Vec`, due to pre-fetching.

tyt2y3:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024

> Btw, the size of `Json` is also 40, so boxing them won’t make `Value` smaller.

it's 32

> ergonomics has always been a big issue

I think a `Value::array<T>()` method would solve the issue

tyt2y3:
Thanks to both of you! @Huliiiiii @Expurple

PR #930, opened 2025-07-30
Forbid unsafe code

Expurple:
## Changes

- [x] `#![forbid(unsafe_code)]` in all workspace crates.

Now that we don't have that weird `PartialEq` impl, we can do this and advertise it.

PR #932, opened 2025-07-31
Fix the Diesel crates

Expurple:
Fixing the CI failures on master. See #930


Expurple:
Eh, `sqlx` uses `ipnetwork` 0.20. We can't upgrade it without upgrading `sqlx`

Huliiiiii:
I think you could also fix the workflow file here — the paths don't include src.

Huliiiiii:
Yes.
https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#patterns-to-match-file-paths

Huliiiiii:
https://github.com/SeaQL/sea-query/blob/f985330c1a8705626a9258691305fd8fe668cdda/.github/workflows/diesel.yml#L3-L19

They should include `src/**`.

Expurple:
> They should include `src/**`.

Ah, I see now. I thought that you meant `src` subdirectories, like `examples/diesel_*/src/**`

Expurple:
Up-to-date and ready for review now

tyt2y3 (approved):
thank you so much

PR #934, opened 2025-07-31
Refactor `value.rs`

Huliiiiii:
## PR Info

Refactor value.rs, which was previously the longest file in the codebase at 2,300 loc. I split it into separate modules and reorganized the contents for better logical structure.

## Changes

- [x] Refactored value.rs
- [x] Renamed `with_vector` to `with_pgvector` to improve naming accuracy, and made the module private as it doesn't export any items.


Huliiiiii:
These are the remaining 10 longest files in the codebase:
```
2225 ./tests/postgres/query.rs
1705 ./tests/sqlite/query.rs
1442 ./tests/mysql/query.rs
1426 ./src/backend/query_builder.rs
676 ./src/query/select.rs
668 ./tests/postgres/table.rs
649 ./src/token.rs
603 ./src/types.rs
530 ./src/value.rs
489 ./src/table/column.rs
```

Expurple (approved):
Yeah, I wanted to similarly modularize `types.rs` and some other large modules. Good to see this getting started

PR #938, opened 2025-08-01
Enable clippy::nursery

Huliiiiii:
## PR Info
I’ve enabled `clippy::nursery`, which helps us write more readable code.
for example, from:

```
if from.is_empty() {
    self.prepare_iden(column, sql);
} else {
    if let Some(table) = table {
        if let TableRef::Table(table) = table.deref() {
            self.prepare_column_ref(
                &ColumnRef::TableColumn(table.clone(), column.clone()),
                sql,
            );
            return;
        }
    }
    self.prepare_iden(column, sql);
}
```

to:
```rust
if !from.is_empty()
    && let Some(table) = table
    && let TableRef::Table(table) = table.deref()
{
    self.prepare_column_ref(&ColumnRef::TableColumn(table.clone(), column.clone()), sql);
} else {
    self.prepare_iden(column, sql);
}
```

## Changes

- [x] Some functions are marked as const

### DIFF
--- a/src/query/select.rs
+++ b/src/query/select.rs
@@ -592,11 +592,7 @@ impl SelectStatement {
         T: IntoColumnRef,
         I: IntoIterator<Item = T>,
     {
-        self.exprs(
-            cols.into_iter()
-                .map(|c| Expr::Column(c.into_column_ref()))
-                .collect::<Vec<Expr>>(),
-        )
+        self.exprs(cols.into_iter().map(|c| Expr::Column(c.into_column_ref())))
     }
 
     /// Select column.
@@ -1794,11 +1790,7 @@ impl SelectStatement {
         T: IntoColumnRef,
         I: IntoIterator<Item = T>,
     {
-        self.add_group_by(
-            cols.into_iter()
-                .map(|c| Expr::Column(c.into_column_ref()))
-                .collect::<Vec<_>>(),
-        )
+        self.add_group_by(cols.into_iter().map(|c| Expr::Column(c.into_column_ref())))
     }
 
     /// Add a group by column.
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,0 +1,6 @@
+[toolchain]
+channel = "1.85"
+components = [
+	"rustfmt",
+	"clippy",
+]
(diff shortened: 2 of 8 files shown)


src/expr.rs
```diff
@@ -2315,7 +2315,7 @@ impl Expr {
     ///     r#"SELECT CURRENT_DATE"#
     /// );
     /// ```
-    pub fn current_date() -> Self {
+    pub const fn current_date() -> Self {
```
Expurple, line 2318: I wouldn't bother with `pub const fn` anywhere. It can impose semver restrictions on us, for little benefit

src/extension/postgres/interval.rs
```diff
@@ -6,19 +6,19 @@ use crate::PgInterval;
...
+            Self::Hour => "HOUR",
+            Self::Minute => "MINUTE",
+            Self::Second => "SECOND",
+            Self::YearToMonth => "YEAR TO MONTH",
+            Self::DayToHour => "DAY TO HOUR",
+            Self::DayToMinute => "DAY TO MINUTE",
+            Self::DayToSecond => "DAY TO SECOND",
+            Self::HourToMinute => "HOUR TO MINUTE",
+            Self::HourToSecond => "HOUR TO SECOND",
+            Self::MinuteToSecond => "MINUTE TO SECOND",
```
Expurple, line 21: I disagree with this lint. The type name is clearly more readable in this case

src/query/select.rs
```diff
@@ -1632,7 +1628,7 @@ impl SelectStatement {
     pub fn join_subquery<T, C>(
         &mut self,
         join: JoinType,
-        query: SelectStatement,
+        query: Self,
```
Expurple, line 1631: One of the many more places where I disagree with this lint

Expurple (changes requested):

Expurple:
Regarding MSRV. I (and my team at work) always use the latest stable, and I assume that our set of dependencies (especially in SeaORM) is always going to float close to that anyway. That said, my opinion may not be representative and we should probably ask the users on Discord

Expurple (approved):

tyt2y3:
okay, let's merge this for now. we can definitely bump MSRV anytime. I usually lag behind the latest stable by 1 version.

PR #939, opened 2025-08-01
impl<T> IntoCondition for T where T: Into<Condition

Huliiiiii:
## PR Info

Make into_condition a shorthand for Into::<Condition>::into

## New Features

- [ ] 

## Bug Fixes

- [ ] 

## Breaking Changes

- [ ] 
 ## Changes

- [ ]

### DIFF
--- a/src/query/condition.rs
+++ b/src/query/condition.rs
@@ -14,11 +14,41 @@ pub struct Condition {
     pub(crate) conditions: Vec<ConditionExpression>,
 }
 
+pub type Cond = Condition;
+
+impl From<Expr> for Condition {
+    fn from(expr: Expr) -> Self {
+        Condition {
+            negate: false,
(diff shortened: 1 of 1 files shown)


Expurple (commented):
I would also:

- Make `trait IntoCondition: Into<Condition>`, so that both imply each other. That way, user generics with `IntoCondition` bounds can still be passed into SeaQuery methods with `Into<Condition>` bounds.

- Document that `IntoCondition` is kept for compatibility and serves no other purpose. Reference https://github.com/SeaQL/sea-orm/discussions/2327

- Document the breaking change in the changelog (user `IntoCondition` impls will no longer compile if the type implements `Into<Condition>`)

tyt2y3 (approved):

tyt2y3:
thank you @Huliiiiii @Expurple

PR #941, opened 2025-08-02
Support jiff

### LINKED ISSUE #851: Support jiff types
Ref to #714.

Ref to https://github.com/burntsushi/jiff/.

I use jiff and now convert all timestamps to string for construct query. We may add first-class support for jiff types.

In short, IMO jiff is better chrono.

Huliiiiii:
## PR Info

- Closes #851

- Dependencies:
  - 

- Dependents:
  - 

## New Features

- [ ] 

## Bug Fixes

- [ ] 

## Breaking Changes

- [ ] 

## Changes

- [ ]

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,7 @@ rust_decimal = { version = "1", default-features = false, optional = true }
 bigdecimal = { version = "0.4", default-features = false, optional = true }
 uuid = { version = "1", default-features = false, optional = true }
 time = { version = "0.3.36", default-features = false, optional = true, features = ["macros", "formatting"] }
+jiff = { version = "0.2.15", default-features = false, optional = true, features = ["std", "perf-inline"] }
 ipnetwork = { version = "0.20", default-features = false, optional = true }
 mac_address = { version = "1.1", default-features = false, optional = true }
 ordered-float = { version = "4.6", default-features = false, optional = true }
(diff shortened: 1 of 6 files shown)


Huliiiiii:
Is there anything else I need to implement?
@tyt2y3

Cargo.toml
```diff
@@ -67,6 +68,7 @@ with-rust_decimal = ["rust_decimal"]
 with-bigdecimal = ["bigdecimal"]
 with-uuid = ["uuid"]
 with-time = ["time"]
+jiff = ["dep:jiff"]
```
tyt2y3, line 71: may be we should following the existing convention of `with-jiff`?

Huliiiiii: There’s a [Clippy lint](https://rust-lang.github.io/rust-clippy/master/index.html#redundant_feature_names) related to this.  Meanwhile, naming the feature directly after the dependency is more idiomatic. If you prefer, you could also rename it to with-jiff.

One more thing I forgot to mention: the types in jiff are quite large and are mostly stack-allocated type. I think we should consider reverting the previous PR that boxed big decimal and array. Alternatively, we could box the jiff types instead—though I wouldn’t recommend that.

tyt2y3: I see. got some clarification from https://github.com/BurntSushi/jiff/discussions/400 but still 96 bytes is quite big.

PR #942, opened 2025-08-02
Add constructors for `Value` (#935)

tyt2y3:
Just tweaking the Array interface.
Well, I originally meant to add jiff support too, but it was already added!


tyt2y3:
I realised there is no easy way to construct an `Array(None)`, so have to add another constructor.

Expurple (commented):
I'd try to utilize `From` as much as possible, instead of inherent constructors

Huliiiiii:
We can implement the `From` trait manually for most concrete types.

PR #946, opened 2025-08-03
Update value bench and unbox variants based on the result

Huliiiiii:
## PR Info

Update benchmarks and unbox Value based on the results.

Benchmarks show that in typical cases, making `Value` smaller does not improve performance. On the contrary, and each heap-allocated value will cause a 1-1.5% performance loss.

This makes sense, since `Value` is usually stored inside a Vec or a Box, such as in `Expr::Binary`.

### DIFF
--- a/benches/value.rs
+++ b/benches/value.rs
@@ -1,3 +1,4 @@
+use core::fmt;
 use std::hint::black_box;
 
 use criterion::{Criterion, criterion_group, criterion_main};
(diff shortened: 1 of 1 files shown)


Huliiiiii:
@Expurple @tyt2y3 Could you try running the benchmarks locally to see if you can reproduce my results?

Huliiiiii:
Well, BurntSushi/jiff#400 explains why I didn’t see any performance impact — in release mode, it’s effectively a comparison between 32 and 40.

PR #947, opened 2025-08-03
No write macro

Huliiiiii:
## PR Info

Replace `write!` with `write_str`

## Why

https://github.com/rust-lang/rust/issues/76490

This one-line change resulted in a 15% performance improvement in the current benchmark😅.
```
impl Write for SqlWriterValues {
    fn write_str(&mut self, s: &str) -> std::fmt::Result {
        // write!(self.string, "{s}")
        self.string.write_str(&s)
    }
}
```

### DIFF
--- a/src/backend/query_builder.rs
+++ b/src/backend/query_builder.rs
@@ -1688,24 +1688,33 @@ where
                 if v.is_empty() {
                     write!(s, "'{{}}'").unwrap()
                 } else {
-                    write!(
-                        s,
-                        "ARRAY [{}]",
-                        v.iter()
-                            .map(|element| self.value_to_string(element))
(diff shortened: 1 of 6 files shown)


src/backend/query_builder.rs
```diff
@@ -1728,11 +1847,32 @@ pub(crate) fn common_well_known_left_associative(op: &BinOper) -> bool {
...
+
+            for $item in $iter {
+                $join
+                $do
+            }
+
+            $($last)?
+        }
+    };
+}
```
tyt2y3, line 1864: I was wondering where this macro come from? I think this is brilliant, although may be we should some short examples for it.

tyt2y3 (approved):

PR #948, opened 2025-08-07
Enable clippy::pedantic

Huliiiiii:
## PR Info

Enable clippy::pedantic

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -55,7 +55,7 @@ jobs:
         with:
           toolchain: stable
           components: clippy
-      - run: cargo clippy --features=all-features --workspace -- -D warnings
+      - run: cargo clippy --features=all-features --workspace -- -D warnings -A clippy::semicolon_if_nothing_returned
       - run: cargo clippy --manifest-path sea-query-sqlx/Cargo.toml --workspace --features runtime-async-std-rustls --features=with-chrono,with-json,with-rust_decimal,with-bigdecimal,with-uuid,with-time,with-ipnetwork,with-mac_address,postgres-array,postgres-vector -- -D warnings
       - run: cargo clippy --manifest-path sea-query-rusqlite/Cargo.toml --all-features --workspace -- -D warnings
       - run: cargo clippy --manifest-path sea-query-postgres/Cargo.toml --all-features --workspace -- -D warnings


tyt2y3:
I think adding some panic does is easiest right now. Converting panic to Error is not possible, unless ... we convert the API to return `Result`, which is another big breaking change.

src/prepare.rs
```diff
@@ -106,7 +107,7 @@ where
                         }
                     }
                 }
-                output.push(mark.clone())
+                output.push(mark.clone());
```
Expurple, line 110: If you're gonna do something like this, you need to configure `rustfmt` to add it automatically. I'm not doing this manually

src/query/window.rs
```diff
@@ -103,7 +105,7 @@ impl WindowStatement {
     }
 
     /// Construct a new [`WindowStatement`] with PARTITION BY custom
-    pub fn partition_by_custom<T>(col: T) -> Self
+    pub fn partition_by_custom<T>(col: &T) -> Self
```
Expurple, line 108: This breaking change needs to be in the changelog. But I'm not sure if it's worth it in the first place

src/types.rs
```diff
@@ -100,13 +101,14 @@ impl SeaRc {
     /// Note that most `Iden`s are statically known
     /// and their representations aren't actually "rendered" and allocated at runtime.
     #[allow(clippy::new_ret_no_self)]
-    pub fn new<I>(i: I) -> DynIden
+    pub fn new<I>(i: &I) -> DynIden
```
Expurple, line 104: This breaking change goes against the whole point of keeping this method around

src/prepare.rs
Huliiiiii: This is [a clippy lint](https://rust-lang.github.io/rust-clippy/master/index.html#semicolon_if_nothing_returned).

Expurple: I understand, but I wouldn't enable it in CI, unless `rustfmt` can add this semicolon automatically. I don't want to manually add these semicolons as the second commit on every PR

src/backend/query_builder.rs
```diff
@@ -13,7 +11,7 @@ pub trait QueryBuilder:
     }
 
     /// Prefix for tuples in VALUES list (e.g. ROW for MySQL)
-    fn values_list_tuple_prefix(&self) -> &str {
+    fn values_list_tuple_prefix(&self) -> &'static str {
```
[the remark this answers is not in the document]
Huliiiiii: I don't think this is a breaking change, the static lifetime is the subtype of all lifetimes.

src/query/window.rs
Huliiiiii: This only calls the `ToString` method, updating the trait bound is also a breaking change.

Huliiiiii: How about changing the internal value of Expr::Custom to Cow?

src/prepare.rs
Huliiiiii: We can disable this lint in CI. Anyway，this makes the return type much clearer.

Expurple: Yeah, I don't mind if it's disabled in CI

src/audit/mod.rs
```diff
@@ -58,6 +58,7 @@ pub type SchemaTable = TableName;
 
 impl QueryAccessAudit {
     /// This filters the selects from access requests.
+    #[must_use]
```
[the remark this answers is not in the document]
Huliiiiii: Sorry, that was another lint and I forgot about it, it should be fixed by now.

Huliiiiii:
This is ready to be merged.

PR #953, opened 2025-08-12
Refactor conversions around qualified identifiers

Expurple:
There was some annoying repetition in this area. `TableName` and `ColumnName` each had three copy-pasted `From` impls. Now they have one, which will be easier to copy-paste when I add `TypeName` and `FunctionName`. I came up with this while re-doing https://github.com/SeaQL/sea-query/pull/922 to add `TypeName`.

Related to https://github.com/SeaQL/sea-query/pull/927.

`types.rs` is getting too long and starting to have visible disjoint "submodules", so I started splitting it.

## PR Info

- Dependents:
  - https://github.com/SeaQL/sea-query/pull/922

## New Features

- [x] The new traits are technically an addition, but I wouldn't advertise it. They are not meant to be used directly. I've documented this on the traits.

### DIFF
--- a/src/types/mod.rs
+++ b/src/types/mod.rs
@@ -8,6 +8,10 @@ use crate::extension::postgres::PgBinOper;
 #[cfg(feature = "backend-sqlite")]
 use crate::extension::sqlite::SqliteBinOper;
 
+mod qualification;
+
+pub use qualification::{MaybeQualifiedOnce, MaybeQualifiedTwice};
+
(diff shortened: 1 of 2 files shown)


Huliiiiii:
How about renaming `into_2_parts` to `into_pair` and `into_3_parts` to `into_triplet`?

Expurple:
Wait a second. Now I'm not sure if this will actually help me to model `TypeName` conversions. I'll post an update under https://github.com/SeaQL/sea-query/pull/922, let's discuss there. Changed to draft for now

PR #956, opened 2025-08-16
Remove the condition on "cross_join"

### LINKED ISSUE #858: cross_join constructs invalid SQL
## Description

In the [example](https://github.com/SeaQL/sea-query/blob/master/src/query/select.rs#L1121), SQL like `CROSS JOIN ... ON ...` is being constructed, but (at least in PostgreSQL) `CROSS JOIN` does not accept `ON`, leading to a syntax error. 

## Versions

0.32.1

Huliiiiii:
## PR Info

- Closes #858

## Bug Fixes

- Removed invalid condition requirement on `SelectStatement::cross_join`

## Breaking Changes

- `SelectStatement::cross_join` no longer accepts a condition

### DIFF
--- a/src/query/select.rs
+++ b/src/query/select.rs
@@ -1142,34 +1142,34 @@ impl SelectStatement {
     ///         .column(Char::Character)
     ///         .column((Font::Table, Font::Name))
     ///         .from(Char::Table)
-    ///         .cross_join(
-    ///             Font::Table,
-    ///             all![
-    ///                 Expr::col((Char::Table, Char::FontId)).equals((Font::Table, Font::Id)),
-    ///                 Expr::col((Char::Table, Char::FontId)).equals((Font::Table, Font::Id)),
-    ///             ]
-    ///         )
+    ///         .cross_join(Font::Table)
     ///         .to_owned();
     ///
     /// assert_eq!(
     ///     query.to_string(MysqlQueryBuilder),
-    ///     r#"SELECT `character`, `font`.`name` FROM `character` CROSS JOIN `font` ON `character`.`font_id` = `font`.`id` AND `character`.`font_id` = `font`.`id`"#
+    ///     r#"SELECT `character`, `font`.`name` FROM `character` CROSS JOIN `font`"#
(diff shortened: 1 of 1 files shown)


src/query/select.rs
```diff
@@ -1712,17 +1684,17 @@ impl SelectStatement {
         )
     }
 
-    fn join_join(
+    fn push_join(
         &mut self,
         join: JoinType,
         table: TableRef,
-        on: JoinOn,
+        on: impl Into<Option<JoinOn>>,
```
tyt2y3, line 1691: umm what does a generic here enables that `Option<Condition>` doesn't ?
it's an internal api so it doesn't matter too much, but just being curious

src/query/select.rs
```diff
@@ -1117,59 +1117,32 @@ impl SelectStatement {
...
-    ///     query.to_string(PostgresQueryBuilder),
-    ///     r#"SELECT "character", "font"."name" FROM "character" CROSS JOIN "font" ON "character"."font_id" = "font"."id" AND "character"."font_id" = "font"."id""#
-    /// );
-    /// assert_eq!(
-    ///     query.to_string(SqliteQueryBuilder),
-    ///     r#"SELECT "character", "font"."name" FROM "character" CROSS JOIN "font" ON "character"."font_id" = "font"."id" AND "character"."font_id" = "font"."id""#
-    /// );
     /// ```
-    pub fn cross_join<R, C>(&mut self, tbl_ref: R, condition: C) -> &mut Self
+    pub fn cross_join<R>(&mut self, tbl_ref: R) -> &mut Self
```
tyt2y3, line 1140: yes I think this is a proper fix.
I will copy this to a new method and back port this to 0.3x

tyt2y3 (approved):

Expurple (approved):
Thanks

src/query/select.rs
Huliiiiii: This allows JoinOn instead of Some(JoinOn) as the parameter

PR #961, opened 2025-08-17
Rework `ColumnSpec`

Huliiiiii:
It should be a struct instead of a vec of enum.

The fix for tests is a bit dirty, but I couldn’t think of any other way than implementing `prepare_column_spec` individually for each backend.

- Close SeaQL/sea-orm#2396
- Close SeaQL/sea-orm#2337


tyt2y3:
sorry for the conflicts, I agree that this should be fixed.

why the order of `AUTO_INCREMENT PRIMARY KEY` was changed? I'd like to keep things unchanged if not too much trouble

Huliiiiii:
> sorry for the conflicts, I agree that this should be fixed.
> 
> why the order of `AUTO_INCREMENT PRIMARY KEY` changed? I'd like to keep things unchanged if not too much trouble

Since order is not stored now, we cannot preserve it. Therefore, I chose the ordering that is most compatible across backends.

Huliiiiii:
> > sorry for the conflicts, I agree that this should be fixed.
> > why the order of `AUTO_INCREMENT PRIMARY KEY` changed? I'd like to keep things unchanged if not too much trouble
> 
> Since order is not stored now, we cannot preserve it. Therefore, I chose the ordering that is most compatible across backends.

Another upside is that it prevents invalid SQL generation from user-provided order.

tyt2y3 (approved):
looks good!

PR #964, opened 2025-08-20
feat(indexes): Support for index column operator class 

### LINKED ISSUE #907: Support for specifying operator class in expression indexes (e.g. gist_trgm_ops)
## Motivation

Currently, sea-query cannot generate index creation statements that specify operator classes for expression indexes, such as creating a GiST index with `gist_trgm_ops` on a column or expression. This limits usage of PostgreSQL features like trigram indexes on expressions.

## Proposed Solutions

Add support for specifying operator classes on columns and expressions in index definitions, allowing statements like:

```sql
CREATE INDEX idx_email_trgm ON users USING GIST (email gist_trgm_ops);
```

or expression indexes with operator classes.

hackermondev:
## New Features
- Adds support for specifying operator classes with index columns (closes #907)

Postgres only.

### DIFF
--- a/src/backend/postgres/index.rs
+++ b/src/backend/postgres/index.rs
@@ -162,6 +162,9 @@ impl IndexBuilder for PostgresQueryBuilder {
                     }
                 }
             }
+            if let Some(operator_class) = col.operator_class() {
+                write!(sql, " {}", operator_class.to_string()).unwrap();
+            }
             false
         });
         write!(sql, ")").unwrap();
--- a/src/index/common.rs
+++ b/src/index/common.rs
@@ -20,12 +20,14 @@ pub struct IndexColumnTableColumn {
     pub(crate) name: DynIden,
     pub(crate) prefix: Option<u32>,
     pub(crate) order: Option<IndexOrder>,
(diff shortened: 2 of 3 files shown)


src/backend/postgres/index.rs
```diff
@@ -167,10 +167,13 @@ impl IndexBuilder for PostgresQueryBuilder {
                         }
                     }
                 }
+              
+                if let Some(operator_class) = col.operator_class() {
+                    write!(sql, " {}", operator_class).unwrap();
```
Huliiiiii, line 172: Please use `write_str` instead of the `write!` macro, `format_args` has a performance cost.

src/backend/postgres/index.rs
```diff
@@ -167,10 +167,13 @@ impl IndexBuilder for PostgresQueryBuilder {
...
+                }
             }
-        );
-
-        sql.write_str(")").unwrap();
+        });
+        write!(sql, ")").unwrap();
```
Huliiiiii, line 176: Also here.

Huliiiiii (changes requested):

PR #975, opened 2025-08-26
Refactor `IntoColumnDef` into a sub trait of `Into<ColumnDef>`

Expurple:
Same as the other `Into*` PRs


Expurple:
Merging without a review, since this is a small mechanical change, with many similar PRs merged recently for other traits.

Issue #954, opened 2025-08-13
Redesign `TypeName`, `DynIden` and `Expr` for correctness


Expurple:
See the thread starting from [this comment](https://github.com/SeaQL/sea-query/pull/922/#issuecomment-3179811630).

Let's continue that discussion about redesigning these types. They should be able to represent most valid SQL type names, identifiers and expressions, and none of the invalid ones.

Expurple:
For flexibility, we already have `cust*` methods and `Cust`/`Custom` variants. I adore them. When something isn't supported natively, they usually provide a working solution. I'm not sure if we need a different, semi-structured "arbitrary node" workaround in addition to that.

I know that you need #926. But if we look closer at it, an "expression list" isn't itself a valid expression that can be freely composed with other expressions. Which means that `ExprList` should be a separate type, rather than a new variant of `Expr`. I think, we should add that type, then add a backend function to render that into an SQL string, and then you can safely format that rendered string into a bigger raw string that comprises your `JSON_EXISTS` function call.

> I think one possible approach is to split different types of expressions into separate structs, and define their serialization to SQL through traits.

I don't see how that's relevant here. We need some "general expression node" in order to compose expressions. That's just how the SQL grammar works. If the `Expr` non-terminal was expressed as `Box<dyn Buildable>` instead of `enum Expr` , I would still argue that `TypeName` and `Keyword` shouldn't implement `Buildable` and shouldn't be usable as expression nodes.

Expurple:
It's really simple. Only the variants *that are actually expressions* should stay in `Expr`. Use `expr + 1` as a quick test. "Real" expressions should be composable.

I don't have time to consult the grammar and check every variant that we have today. But it's obvious that `Expr::TypeName` doesn't pass this test and should be removed. In practice, it's only used to implement casts, so we should replace it with a `Cast(Box<Expr>, TypeName)` variant. It's very similar to the existing `Expr::AsEnum` variant, so we should dive into that and see if these variants can be unified.

Sorry, I caused some confusion about `Keyword` previously. I should've checked the `Keyword` contents. Turns out, it's already "only the keywords that are usable as expressions", not "arbitrary SQL keywords". `Expr::Keyword` can stay.

Huliiiiii:
Maybe this can be removed?
```rust
// src/extension/postgres/types.rs
#[derive(Clone, Debug)]
#[non_exhaustive]
pub enum TypeRef {
    Type(DynIden),
[...]

PR #977, opened 2025-08-27
Rework `CAST` expressions

Expurple:
See https://github.com/SeaQL/sea-query/issues/954.

I think, this version is more correct and easier to understand (regarding case-sensitivity, `[]` handling, etc). The summary is in the changelog, as always

### DIFF
--- a/src/backend/postgres/query.rs
+++ b/src/backend/postgres/query.rs
@@ -42,7 +42,7 @@ impl QueryBuilder for PostgresQueryBuilder {
         let (ty, sfx) = if let Some(base) = type_name.strip_suffix("[]") {
             (base, "[]")
         } else {
-            (type_name.as_ref(), "")
+            (type_name, "")
         };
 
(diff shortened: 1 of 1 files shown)


Expurple:
But now I'm stuck with this failing `ExprTrait::as_enum` doctest. I've read https://github.com/SeaQL/sea-query/pull/922/#issuecomment-3183979509 and I understand the need to separate quoted (case-sensitive) and unquoted (case-insensitive) casts, but I still don't understand why `as_enum` needs to do a cast only on Postgres and do nothing on MySQL and SQLite. This behavior is undocumented, seems very weird, and requires that special `AsEnum` node that I try to remove. I this something that SeaORM needs? Perhaps, SeaORM should apply this DB-specific logic? I find it weird that I can explicitly call `my_val.as_enum(my_type)` (a seemingly low-level quary builder method) and get no cast at all depending on the database.

PR #980, opened 2025-08-29
Change the shared lock syntax to 'LOCK IN SHARE MODE' in MSQL for MariaDB support

brqnko:
## PR Info

- Closes https://github.com/SeaQL/sea-orm/issues/2107

## New Features

- [ ] None 

## Bug Fixes

- [x] Fixes a syntax error when using `lock_shared()` on MariaDB. 

## Breaking Changes

- [ ] None 

## Changes

- [x] Create a `lock_phrase` method in `QueryBuilder` so that it can be overridden in `MySqlQueryBuilder` to use `LOCK IN SHARE MODE`.

### DIFF
--- a/src/backend/query_builder.rs
+++ b/src/backend/query_builder.rs
@@ -548,7 +548,6 @@ pub trait QueryBuilder:
     fn prepare_table_sample(&self, _select: &SelectStatement, _sql: &mut impl SqlWriter) {}
 
     /// Translate [`LockType`] into SQL statement.
-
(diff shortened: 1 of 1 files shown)


brqnko:
Yes, this LOCK IN SHARE MODE syntax is compatible with MySQL.

While FOR SHARE is the recommended syntax in MySQL 8.0+, LOCK IN SHARE MODE is still supported for backward compatibility and works on both MariaDB and MySQL.

- https://dev.mysql.com/doc/refman/9.4/en/innodb-locking-reads.html

Huliiiiii:
Hi, can you fix this conflict?

PR #984, opened 2025-08-31
Implement the `EscapeBuilder` trait specifically for Postgres

### LINKED ISSUE #694: Escape string \z meaningless for postgres
## Description

The default string escaper [uses](https://github.com/SeaQL/sea-query/blob/36c03c36a360cb1c4a68cf8059b630ddcda62c33/src/backend/mod.rs#L43) `\z` to escape `0x1a` ("substitute character"). However, this is not a valid escape sequence in postgresql. The list of valid escape sequences can be found [here](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-BACKSLASH-TABLE).

## Steps to Reproduce

```rust
#[test]
fn test_substitute_char() {
    let ch: SimpleExpr = '\u{1a}'.into();
    let query = Query::select()
        .expr(ch)
        .to_owned();
    // Just one possible escape (other possibilities include e.g. \u001A, etc.)
    assert_eq!(query.to_string(PostgresQueryBuilder), r#"SELECT E'\x1A'"#);
}
```

### Expected Behavior

The output is a valid escape of `\x1a`.

### Actual Behavior

Sea query attempts to escape using `\z`, but this is not a valid escape character.

### Reproduces How Often

Deterministic

## Versions

Sea query 0.30.1

Huliiiiii:
I didn’t escape Unicode because Postgres can handle Unicode input directly.

I also renamed `need_escape` to `needs_escape` to better follow English grammar.

Closes #694

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -63,7 +63,7 @@ pub trait QuotedBuilder {
 
 pub trait EscapeBuilder {
     /// Return if string literal needs to be escaped
-    fn need_escape(&self, s: &str) -> bool {
+    fn needs_escape(&self, s: &str) -> bool {
         s.chars().any(|c| {
             matches!(
                 c,
--- a/src/backend/postgres/mod.rs
+++ b/src/backend/postgres/mod.rs
@@ -160,7 +160,7 @@ mod tests {
     fn test_unescape_string() {
         let escaper = PostgresQueryBuilder;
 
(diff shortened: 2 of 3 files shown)


Huliiiiii:
Is the test wrong?
```
> psql -d postgres
psql (17.5)
Type "help" for help.

postgres=# select E' "abc" ';
 ?column? 
----------
  "abc" 
(1 row)

postgres=# select E'a"b';
 ?column? 
----------
 a"b
(1 row)
```

PR #987, opened 2025-09-01
Add pseudo tables

### LINKED ISSUE #842: No way to express "excluded" in upsert where clause in SeaQL
## Description

I want to express a specific update pattern in SeqQL, but there is no way to do so in the query builder.
The equivalent query I want to build is as so:
```sql
INSERT INTO kv (key, value)
VALUES ('hi', 'blah')
ON CONFLICT (key) DO UPDATE SET value   = excluded.value,
                                version = version + 1
WHERE excluded.value <> value
```

however I am unable to generate the `excluded.value` method within the `WHERE` clause here. The closest I can get is
```rust
let query = Kv::insert(model)
    .on_conflict(
        OnConflict::column(kv::Column::Key)
            .update_column(kv::Column::Value)
            .value(kv::Column::Version, Expr::col(kv::Column::Version).add(1))
            .action_and_where(Expr::col(kv::Column::Value).not_equals(kv::Column::Value))
            .to_owned(),
    )
    .into_query()
    .to_string(SqliteQueryBuilder);
```

which evaluates to

```sql
INSERT INTO "kv" ("key", "value")
VALUES ('hi', 'blah')
ON CONFLICT ("key") DO UPDATE SET "value"   = "excluded"."value",
                                  "version" = "version" + 1
WHERE "value" <> "value"
```

### SQL
```sql
CREATE TABLE IF NOT EXISTS "kv" ( "key" text NOT NULL PRIMARY KEY, "value" text NULL, "version" integer NOT NULL DEFAULT 0 );
```

### Entity
```rust
//! `SeaORM` Entity, @generated by sea-orm-codegen 1.1.1

use sea_orm::entity::prelude::*;

#[derive(Clone, Debug, PartialEq, DeriveEntityModel, Eq)]
#[sea_orm(table_name = "kv")]
pub struct Model {
    #[sea_orm(primary_key, auto_increment = false, column_type = "Text")]
    pub key: String,
    #[sea_orm(column_type = "Text", nullable)]
    pub value: Option<String>,
    pub version: i32,
}

#[derive(Copy, Clone, Debug, EnumIter, DeriveRelation)]
pub enum Relation {}

impl ActiveModelBehavior for ActiveModel {}
```

As far as I can tell from parusing the source code, the only time the special `excluded` table can be generated as part of a query is in the assignment.

## Versions

OS: Linux
Arch: AMD64
Database: SQLite, 

libsqlite3-sys: 0.30.1
Bundled SQLite version: 3.46.0
Bundled SQLite source ID: `2024-05-23 13:25:27 96c92aba00c8375bc32fafcdf12429c58bd8aabfcadab6683e35bbb9cdebf19e`

sea-orm: 1.1.1
sea-orm-macros: 1.1.1
sea-bae: 0.2.1
sea-query: 0.32.0
sea-orm-migration: 1.1.1
sea-orm-cli: 1.1.1
sea-schema: 0.16.0
sea-schema-derive: 0.3.0

Huliiiiii:
Close #842

### DIFF
--- a/src/types/iden/compound.rs
+++ b/src/types/iden/compound.rs
@@ -135,12 +135,12 @@ pub enum ColumnRef {
     Column(ColumnName),
     /// An `*` expression, potentially qualified as `(database.)(schema.)(table.)*`.
     Asterisk(Option<TableName>),
-    /// NEW.*
+    /// NEW.<column_name>
     New(DynIden),
-    /// OLD.*
+    /// OLD.<column_name>
     Old(DynIden),
     #[cfg(feature = "backend-postgres")]
-    /// excluded.*
+    /// excluded.<column_name>
     Excluded(DynIden),
 }
 


src/types/iden/compound.rs
```diff
@@ -135,6 +135,13 @@ pub enum ColumnRef {
     Column(ColumnName),
     /// An `*` expression, potentially qualified as `(database.)(schema.)(table.)*`.
     Asterisk(Option<TableName>),
+    /// NEW.*
+    New(DynIden),
+    /// Old.*
+    Old(DynIden),
+    #[cfg(feature = "backend-postgres")]
+    /// excluded.*
+    Excluded(DynIden),
```
Expurple, line 144: To we really need these special variants? These table names look like they can be already expressed as "regular" table names: `ColumnRef::from(("OLD", MyColumn))`

It may be enough to define some unit-struct idens for these tables and document the approach.

Huliiiiii:
> Also, your current approach doesn't support asterisk, like `NEW.*` (literally `*` in the query)

This means the glob pattern, not the SQL asterisk.

Expurple:
I understand. I'm pointing out that the SQL asterisk is also valid with these tables. I'd like to find a clean way to support it right away

Issue #995, opened 2025-10-06
Add drop_column_if_exists


ctron:
The SeaORM migration manager right now has:

* `add_column` and `add_column_if_not_exists`
* `drop_column`

But not `drop_column_if_exists`. Despite PostgreSQL supporting this.

## Motivation

If makes it easier to deal with certain cases where the column does or does not exist.

## Proposed Solutions

Add `drop_column_if_exists`

PR #996, opened 2025-10-10
implemented drop column if exists for postgres

kev1N916:
## PR Info

Closes https://github.com/SeaQL/sea-query/issues/995

## New Features

This pull request implements DROP COLUMN IF EXISTS for Postgres which is not supported by sea-query currently.


tyt2y3 (approved):
I will remove the panics

PR #1000, opened 2025-10-15
Add `null` method for `Expr`

### LINKED ISSUE #999: Add `Expr::null` method analogously as for other keywords
## Motivation

It's not obviously how to specify `NULL` value correctly.
Expected to have similar method to `Expr::current_date` for `NULL`, but there don't it.

In `Expr::current_date` function used such code: `Self::Keyword(Keyword::CurrentDate)`, so we will try to use something like `SimpleExpr::Keyword(Keyword::Null)` according to expectations to write "similar code". It's too long and inconvenient.

There are also variant to use `Keyword::Null` directly, but docs doesn't show it and without read the code (`From` implementations for `SimpleExpr`), it will be difficult to find this variant.

## Proposed Solutions

Add `Expr::null` method that will be friendly for users

Desiders:
## PR Info

- Closes #999

## New Features

- Add `null` method for `Expr`

### DIFF
--- a/src/expr/enum.rs
+++ b/src/expr/enum.rs
@@ -842,14 +842,8 @@ impl Expr {
     /// let query = Query::select().expr(Expr::null()).to_owned();
     ///
     /// assert_eq!(query.to_string(MysqlQueryBuilder), r#"SELECT NULL"#);
-    /// assert_eq!(
-    ///     query.to_string(PostgresQueryBuilder),
-    ///     r#"SELECT NULL"#
-    /// );
-    /// assert_eq!(
-    ///     query.to_string(SqliteQueryBuilder),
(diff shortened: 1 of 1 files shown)


Huliiiiii:
IMO, for a query builder library, the trait bound `T: Into<Expr>` is clear enough to indicate what parameters it can accept.

Desiders:
> You can claim a free [Rustacean Sticker Pack](https://www.sea-ql.org/rust-sticker-pack/) if you want it haha.

I'm embarrassed by such encouragement for such change in the code, even if it's beautiful "date". I'd want to get it and learn how to do it, but I would not like to cause inconvenience.

Congratulations the team on 1000th repository thread :D

PR #1002, opened 2025-10-18
Move aggregate methods and if_null to ExprTrait

Huliiiiii:


tyt2y3:
there is conflict, I will resolve this offline

PR #1004, opened 2025-10-28
Refactor Value::Array and add Value::Enum

Huliiiiii:
Changes:

- Generate `ARRAY[` instead of `ARRAY [` to match SQL conventions
- Arrays now support null values

### DIFF
--- a/sea-query-rusqlite/Cargo.toml
+++ b/sea-query-rusqlite/Cargo.toml
@@ -3,7 +3,7 @@
 
 [package]
 name = "sea-query-rusqlite"
-version = "0.8.0-rc.10"
+version = "0.8.0-rc.12"
 authors = [ "Ivan Krivosheev <py.krivosheev@gmail.com>" ]
 edition = "2024"
 description = "Binder traits for connecting sea-query with Rusqlite"


sea-query-postgres-types/src/range.rs
```diff
@@ -109,3 +109,43 @@ fn hash_range_bound<H: Hasher>(rb: &RangeBound<f64>, state: &mut H) {
         RangeBound::Unbounded => (),
     }
 }
+
+impl<T: Clone + Display> From<&RangeBound<T>> for std::ops::Bound<T> {
```
Huliiiiii: Why don't we use std::ops::Bound?

tyt2y3 (approved):
thank you. this is longer than I can dissect in one go, but looks good in general

tyt2y3 (approved):

src/value/array.rs
```diff
@@ -0,0 +1,743 @@
...
+#[cfg(feature = "with-json")]
+use crate::backend::ValueEncoder;
+use std::sync::Arc;
+
+type EnumArray = Box<(Arc<str>, Box<[Option<Arc<Enum>>]>)>;
+
+#[derive(Debug, Clone)]
+#[cfg_attr(not(feature = "hashable-value"), derive(PartialEq))]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[non_exhaustive]
```
tyt2y3: this doesn't have to be non_exhaustive I guess?

Issue #69, opened 2021-07-19
Question: support for range types


Milias:
Hello,

I have been testing this crate for the past several days and after some searching in the documentation and in the code I cannot find any references to [range support](https://www.postgresql.org/docs/13/rangetypes.html). I have only used Sea Query with a postgres database, together with the [postgres crate](https://docs.rs/postgres/0.19.1/postgres/index.html), which does support ranges. Additionally I used [this crate](https://github.com/sfackler/rust-postgres-range).

Are there any plans to support range types? Could you point me to where would it be the best way to start supporting them in the code?

Thank you.

tyt2y3:
Hi,

Thank you for reaching out.

Yes we can add support of it. But how'd you intend to use it?

Can you make some snippets on the schema & queries for intended use?

With that being done, I will be able to set out an implementation plan.

Milias:
Hello, thank you for answering so quickly!

The use case I have in mind for ranges is for scheduling applications, that is, representing a period of time. Postgres already has a few built-in range types for dates and times (`tsrange`, `tstzrange`, `daterange`: see [here](https://www.postgresql.org/docs/13/rangetypes.html)). In my case I need to use a custom data type exactly like the example given in section 8.17.8 of that link, that is:

```
CREATE FUNCTION time_subtype_diff(x time, y time) RETURNS float8 AS
'SELECT EXTRACT(EPOCH FROM (x - y))' LANGUAGE sql STRICT IMMUTABLE;

CREATE TYPE timerange AS RANGE (
    subtype = time,
    subtype_diff = time_subtype_diff
);

SELECT '[11:10, 23:00]'::timerange;
```

Then, if you have a table "events" with a column "range" with this data type you could insert new row as

```
insert into "events" ( "range" ) values ('[10:30, 11:00]');
```

Return all events happening at a given time:

```
select * from "events" where "range" @> '8:45'::time;
```

Or an exclusion constraint so that two events cannot overlap:

```
alter table events add constraint no_overlapping_events exclude using gist ("range" with &&);
```

These are simplified examples that I used in developing this toy application. Mainly, the range support that I would really use in Sea Query would be regarding `insert` and `select` queries, both in generating the queries themselves as well as converting their results back to rust types.

On the rust side this is a simplified example code to illustrate the usecase:

```
#[derive(Iden)]
enum Events {
    Table,
    Id,
    Range,
}

#[derive(Debug)]
pub struct Event {
    pub id: i64,
[...]

PR #1007, opened 2025-11-04
Add postgres range type

jonasrichard:
## PR Info

- Closes https://github.com/SeaQL/sea-query/issues/69

## New Features

- [ ] Provide a new type `RangeType` to implement Postgres range type
- [ ] Provide range type as an element type of Postgres Array

## Doubts

- [ ] Not sure where to implement `hashable-value` features. I put that in the `with_range.rs`
- [ ] It implements time ranges with Chrono and Time I left out Jiff now
- [ ] Not sure how it plays together with `with-json`
- [ ] How to make RangeType from std Rust types (tuple, std range, how to describe exclusive and inclusive bounds?)
- [ ] A refactor PR is going on which may affect this branch (https://github.com/SeaQL/sea-query/pull/1004)

### DIFF
--- a/sea-query-postgres-types/src/range.rs
+++ b/sea-query-postgres-types/src/range.rs
@@ -11,7 +11,7 @@ use postgres_types::{IsNull, Kind, ToSql, Type, to_sql_checked};
 
 #[derive(Clone, Debug, Eq, PartialEq, Hash, serde::Serialize, serde::Deserialize)]
 //#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
-pub enum RangeBound<T: Display + ToSql> {
+pub enum RangeBound<T: Clone + Display + ToSql> {
     Exclusive(T),
     Inclusive(T),
     Unbounded,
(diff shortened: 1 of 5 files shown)


tyt2y3:
i see, shall we move that to a separate type crate? something like `sea-query-types` / `sea-query-postgres-types`?

jonasrichard:
I will create a create for postgres only now, so the 2nd. And addressing the `sea-query-sqlx` compilation error.

jonasrichard:
@tyt2y3 I started to move Range to the sea-query-postgres-types crate but I am struggling a bit.

* I cannot reach this crate from the root crate sea-query. I think I would like to re-export these two types from sea-query-postgres crate and that crate is enough to be visible from sea-query.
* Another one, I tried to put a serde feature - because it is needed by the type definition - in the Cargo.toml of sea-query-postgres-types and say: use the same features as the root crate. Hope it is correct.
* I included serde in types crate as well, I feel that it is a duplication to say the serde version twice (this and in the root crate).

If you can tell me what to change to fix the 1st point I can go on.

jonasrichard:
Okay, I think I am ready with that. I managed to run almost all checks with build-tools/rustclippy.sh

tyt2y3 (approved):
thank you! if this works for you let's get this merged for now.
can create a followup PR for SQLx.

I think the strategy is to convert our own `RangeType` to sqlx's `Range` on the fly and call `bind()`

Issue #993, opened 2025-10-05
Postgres WITH/CTE with only a VALUES list - No query


BenGatewood:
## Motivation

Hey all - Opening this at Huli's request after a discussion on Discord:

It's not currently possible to construct SQL like the following where the CTE consists only of a VALUES list without  a query: 

```
WITH data (
    field1,
    field2,
    field3,
    field4,
    field5
    ) AS (
        VALUES (
                'a',
                123,
                'blah',
                333,
                1
               )
), ins1 AS (
    INSERT INTO tableA (col1, col2, col3) SELECT d.field1, d.field2, d.field3 FROM data d RETURNING id as tableA_id
) INSERT INTO tableB (table_a_id, colx, coly) SELECT ins1.tableA_id, d.field4, d.field5 FROM data d CROSS JOIN ins1;
```

jonasrichard:
As I see after the values there can be more rows, and a values can be a scalar query like:

```sql
... VALUES ( (1, 2), (2, 4), (3, (SELECT 6)) ) ...
```

Thinking adding an `Option<Vec<Value>>` to `CommonTableExpression` but if we have a `Value | ScalarQuery` that would be also handy. Maybe at first it can be just a scalar value.

PR #1008, opened 2025-11-18
Implement VALUES clause for Postgres CTE

jonasrichard:
## PR Info

- Closes https://github.com/SeaQL/sea-query/issues/993

## New Features

- [ ] Add a `values()` fn to `CommonTableExpression` in order to feed Postgres VALUES clause. 

## Changes

- [ ] During building the CTE struct, either query or values should be specified, ~~otherwise code can panic, see documentation~~.

### DIFF
--- a/src/query/with.rs
+++ b/src/query/with.rs
@@ -6,12 +6,12 @@ use crate::{
 use inherent::inherent;
 
 #[derive(Debug, Clone, PartialEq)]
-pub enum CTEQuery {
-    Query(Box<SubQueryStatement>),
+pub(crate) enum CteQuery {
+    SubQuery(Box<SubQueryStatement>),
     Values(Vec<Values>),
 }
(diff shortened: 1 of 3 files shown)


tyt2y3:
I agree with @Huliiiiii

src/query/with.rs
```diff
@@ -5,6 +5,18 @@ use crate::{
 };
 use inherent::inherent;
 
+#[derive(Debug, Clone, PartialEq)]
+pub enum CTEQuery {
```
Huliiiiii: CteQuery is the more idiomatic. (e.g. Http, Uuid).

jonasrichard: I changed name and made it create-private since query builder needs it.

Huliiiiii (approved):
Thanks!

Issue #1011, opened 2025-11-26
[Flaw] Types from stronger-pinned dependencies should be re-exported


GenuaGSchulz:
## Description

When working with Sea-Query, especially with optional features such as 'ipnetwork' we would like to be able to use the exact types (such as IpNetwork) that Sea-Query uses without having to include a pinned-version dependency to the (ipnetwork) crate to an old version ourselves. The types in question mostly come from optionally-feature-enabled dependencies that are included with a minor-version-pinned version in Sea-Query's Cargo.toml.

The proposal is to re-export types that fit the above description, so application code can use these exact types.

## Versions

sea-query = "0.32.7"

PR #1012, opened 2025-11-26
Re-exporting stronger-pinned third-party types

GenuaGSchulz:
## PR Info

- Closes https://github.com/SeaQL/sea-query/issues/1011

## Changes

- [x] modified use -> pub use for some of the optional dependency types in value.rs


tyt2y3:
I am open to pub-ing types, however I want to point out that minor version requirement doesn't mean the exact version has to be used, but any higher version before the next major. https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#default-requirements

In case of `time` crate, because there was a minor breaking change.

tyt2y3:
I will pub all types

PR #1024, opened 2025-12-19
Escape array suffix in prepare_type_ref

### LINKED ISSUE #1023: Invalid SQL generated with arrays and TypeRef
## Description

Invalid SQL generated from functions taking `TypeRef` when given a string for an array type.

```rust
        let (sql, _) = Query::select()
            .column(Asterisk)
            .from_function(
                PgFunc::json_table(
                    Expr::val(r#"{"foo": "some value", "bar": ["x", "y", "z"]}"#),
                    "$",
                )
                .column("foo", "text")
                .path("$.foo")
                .build_column()
                .column("bar", "text[]")
                .path("$.bar")
                .wrapper(WrapperKind::WithUnconditional)
                .empty_array_on_empty()
                .error_on_error()
                .build_column()
                .build(),
                "stuff",
            )
            .build_sqlx(PostgresQueryBuilder);
```

produces a SQL fragment like:

```sql
SELECT * FROM JSON_TABLE('{"foo": "some value", "bar": ["x", "y", "z"]}', '$' COLUMNS (foo "text" PATH '$.foo', bar "text[]" PATH '$.bar' WITH UNCONDITIONAL WRAPPER EMPTY ARRAY ON EMPTY ERROR ON ERROR)) AS "stuff"
```

which gives the following error: `ERROR:  type "text[]" does not exist`.

Removing the quotes around `"text[]"` in the generated SQL fixes the issue.

## Steps to Reproduce

See above.

### Expected Behavior

Valid SQL

### Actual Behavior

Invalid SQL when using arrays as they are quoted.

### Reproduces How Often

Always.

## Versions

sea-query: 1.0.0-rc.22
PostgreSQL: 17.4
OS: RHEL8

Huliiiiii:
Closes #1023

### DIFF
--- a/src/backend/query_builder.rs
+++ b/src/backend/query_builder.rs
@@ -1820,7 +1820,6 @@ mod tests {
     #[cfg(feature = "with-chrono")]
     use chrono::{DateTime, FixedOffset, NaiveDate, NaiveDateTime, NaiveTime, Utc};
 
-    #[cfg(feature = "with-chrono")]
     use crate::QueryBuilder;
 
     /// [Postgresql reference](https://www.postgresql.org/docs/current/datatype-datetime.html#DATATYPE-DATETIME-INPUT-TIMES)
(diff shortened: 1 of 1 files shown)


Expurple (commented):
Looks good, but we also need a test for this

PR #1027, opened 2025-12-21
Redesign array-related APIs

Huliiiiii:
Changes:
- Remove `NotU8`

Refactoring:
- Use `ArrayElement` to simplify code
- Move hash mod to a separate file

### DIFF
--- a/src/value/array.rs
+++ b/src/value/array.rs
@@ -452,11 +452,8 @@ impl From<Array> for Value {
 ///
 /// When implemented, SeaQuery will provide:
 /// - `ValueType` for `Vec<T>` and `Vec<Option<T>>`
-/// - `From<Vec<T>> for Value` / `From<Vec<Option<T>>> for Value`
-/// - `From<Vec<T>> for Array` / `From<Vec<Option<T>>> for Array`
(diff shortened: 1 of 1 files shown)


src/value/array.rs
```diff
@@ -562,183 +554,76 @@ where
...
-                (Self::BigDecimal(l0), Self::BigDecimal(r0)) => l0 == r0,
-                #[cfg(feature = "with-ipnetwork")]
-                (Self::IpNetwork(l0), Self::IpNetwork(r0)) => l0 == r0,
-                #[cfg(feature = "with-mac_address")]
-                (Self::MacAddress(l0), Self::MacAddress(r0)) => l0 == r0,
-                (Self::Null(l0), Self::Null(r0)) => l0 == r0,
-                _ => false,
-            }
-        }
+impl<T> std::iter::FromIterator<T> for Array
```
Huliiiiii: Do we need to provide FromIterator?

tyt2y3: well I am not sure but good for convenience?

tyt2y3: it's needed in sea-orm

PR #1037, opened 2026-01-01
Use named columns in sqlx examples

hrimov:
## PR Info

A more resilient approach using named column access instead of numeric indices. This will be even more consistent given the example as a whole - `query_as_with` and `FromRow` are used throughout.

I don't know if this was intended or not, but I left a `try_get` in the comment as an alternative. Let me know if its needed.


hrimov:
Seems that for MySQL and SQLite we need to have an alias for aggregate functions result-column

Issue #737, opened 2024-01-08
`Table::drop()` with multiple `.table` when using sqlite3 causes invalid sql


uttarayan21:
## Description
```rust
    let down = Table::drop()
        .table(RootFolderIden::Table)
        .table(FolderIden::Table)
        .table(FileIden::Table)
        .build(SqliteQueryBuilder);
```
That generates the following sql
```sql
DROP TABLE "root_folder", "folder", "file"
```

## Steps to Reproduce

1. Generate the sql from the above code snippet.

### Expected Behavior

Ideally it should be the same as when creating a table. (i.e. overwrites the table name with the last one used )

### Actual Behavior

` DROP TABLE "root_folder", "folder", "file"`
Generated a drop table statement with all the tables which isn't supported by sqlite (https://www.sqlite.org/lang_droptable.html)

### Reproduces How Often

Always

## Versions

```
│   └── sea-query v0.30.6
│       ├── sea-query-attr v0.1.1 (proc-macro)
│       └── sea-query-derive v0.4.1 (proc-macro)
│   ├── sea-query v0.30.6 (*)
├── sea-query v0.30.6 (*)
```

thgoebel:
I ran into this because the README told me to: https://github.com/SeaQL/sea-query/tree/0.32.0#table-drop

Issue #471, opened 2022-10-07
create index on non-unique text field throws error for Sqlite


boan-anbo:
## Description

Migrate with index created on text field without unique throws errors for Sqlite.

## Steps to Reproduce

1. Define migration:
```rust
.col(
    ColumnDef::new(MyEntity::MyColumn)
        .text()
        .null()
        .default("NULL")
)
.index(
    Index::create()
        .name("my_column")
        .col(MyEntity::MyColumn)
)
```
3. the cli throws
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Exec("error returned from database: (code: 1) near \"(\": syntax error")', migration\src\m20220929_111159_create_db.rs:225:20

```
6. It works if you change migration to
```rust
.col(
    ColumnDef::new(MyEntity::MyColumn)
        .text()
        .null()
        .default("NULL")
)
.index(
    Index::create()
        .name("my_column")
        .col(MyEntity::MyColumn)
        .unique() // add this
)
```

### Expected Behavior

Not to throw as Sqlite supports indexing non-unique field.

## Versions

```toml
sea-orm={version = "0.9.2", features = [
    "sqlx-sqlite",
    "runtime-tokio-rustls",
    "mock",
    "macros",
    "debug-print",
    "uuid"
]}
```

billy1624:
I just transfer this issue to SeaQuery

ikrivosheev:
Hmmm, @boan-anbo @billy1624 @oonxt , after read docs:

1. https://www.sqlite.org/lang_createtable.html
2. https://www.postgresql.org/docs/current/sql-createtable.html

what behavior is expected?

0xpr03:
I for one would expect that it either errors out from the start (you have to make it unique), or just make two SQL commands with some docs that index-creation may be separated to a second command. The latter one may be a bad idea when looking at atomic commands errors-for-command relations.

PR #1038, opened 2026-01-08
Fix backend guards for inline index and SQLite DROP TABLE

Summpot:
Fixes the two items tracked under #1031:

- #471: `TableCreateStatement::index()` is now deprecated for any build that isn’t *MySQL-only* (i.e. if Postgres/SQLite is enabled, or if multiple backends are enabled). MySQL-only builds keep the method non-deprecated.
- #737: when `backend-sqlite` is enabled, `Table::drop()` uses a typestate builder so the first `.table()` call is well-defined. A *second* `.table()` call now compiles but emits a deprecation warning (to avoid a confusing "method not found" error) and will panic at runtime, because SQLite strictly forbids dropping multiple tables in a single statement. Non-sqlite builds keep the legacy multi-`.table()` API.

Also updates tests to match the new behavior.

Local verification: `cargo test -p sea-query --test test-sqlite`.

### DIFF
--- a/src/table/drop.rs
+++ b/src/table/drop.rs
@@ -147,6 +147,22 @@ impl TableDropStatement<TableDropPending> {
 }
 
 #[cfg(feature = "backend-sqlite")]
+impl TableDropStatement<TableDropDefined> {
+    #[deprecated(
+        since = "1.0.0",
+        note = "SQLite strictly forbids dropping multiple tables in a single statement. Please split this into multiple Table::drop() calls."
+    )]
+    pub fn table<T>(self, _table: T) -> Self
+    where
+        T: IntoTableRef,
+    {
+        panic!(
+            "Attempted to drop multiple tables in SQLite mode. This is not supported. See compiler warnings."
+        );
+    }
+}
+
+#[cfg(feature = "backend-sqlite")]
 impl<S> TableDropStatement<S> {
     pub(crate) fn is_if_exists(&self) -> bool {
         self.inner.is_if_exists()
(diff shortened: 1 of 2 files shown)


Summpot:
I'd like to discuss a potential refinement to the SQLite strict mode implementation.

Currently, the Typestate pattern causes a **hard compile-time error** (`method not found`) when `.table()` is called a second time. While safe, this can sometimes be abrupt.

**The Alternative Idea:**
Instead of removing the method entirely, we could implement it for the `Defined` state but mark it as `#[deprecated]`.

```rust
#[deprecated(note = "SQLite does not support dropping multiple tables. Please split into multiple statements.")]
pub fn table(self, ...) { ... }

```

**The Motivation (AI/LLM Coding):**
In the era of AI-assisted coding (Copilot, Cursor, etc.), a hard compilation error often confuses agents, leading them to hallucinate non-existent APIs. A `deprecated` warning with a clear actionable note serves as a much better "prompt" for agents to automatically refactor the code (e.g., splitting one statement into two).

**The Implementation Question (Panic vs. No-op):**
If we allow the code to compile (with warnings), we need to decide on the runtime behavior for that second call:

1. **Silent No-op:** Simply ignore the second table.
* *Risk:* This is dangerous as it leads to silent data inconsistency (user thinks table is dropped, but it isn't).


2. **Panic:** Explicitly `panic!` inside the builder.
* *Pros:* Since generating invalid SQL (`DROP TABLE a, b`) results in a runtime database error on SQLite anyway, a Rust-level panic is functionally equivalent in terms of safety, but provides a clearer error message than a generic SQL syntax error.

PR #1054, opened 2026-02-05
Add constraint type

### LINKED ISSUE #1053: Add "alter table constraint" support separately from table definition
## Motivation

Currently, there is no support of create constraints to existing tables. We have similar functionality for foreign keys using `ForeignKeyCreateStatement` (and other operations like drop), but for constraints it's not implemented.

It can be helpful in migrations when we need to modify tables' constraints to avoid writing raw SQL.

## Proposed Solutions

I think we need to add new type `TableConstraint`, `ConstraintStatement` enum and `ConstraintStatement::Create` variant with `ConstraintCreateStatement` value.

To build SQL from new types, we need to create new trait `ConstraintBuilder` and implement it for supported databases.

## Additional Information

I have implemented some of the proposed code and would be ready to continue if the feature request is approved.

Desiders:
## PR Info

- Closes #1053

## New Features

- [ ] 

## Bug Fixes

- [ ] 

## Breaking Changes

- [ ] 

## Changes

- [ ]


Desiders:
As I see, there was an added `DeleteConstraint` variant for alter table: https://github.com/SeaQL/sea-query/commit/f3c0b8365e4378cf4de5c9426023e0d9a260494b.

I thought too about adding a `Constraint` variant for delete and using something like https://github.com/SeaQL/sea-query/blob/f3c0b8365e4378cf4de5c9426023e0d9a260494b/src/backend/mysql/table.rs#L20000 for table alter statement

PR #1055, opened 2026-02-09
Wrap part of on_conflict_target in parentheses

Huliiiiii:
@tisonkun mentioned to me that in PostgreSQL, some statements used as the `index_expression` in a `conflict_target` need to be wrapped in parentheses.

https://www.postgresql.org/docs/current/sql-insert.html
```
where conflict_target can be one of:

    ( { index_column_name | ( index_expression ) } [ COLLATE collation ] [ opclass ] [, ...] ) [ WHERE index_predicate ]
    ON CONSTRAINT constraint_name
```

### DIFF
--- a/src/backend/query_builder.rs
+++ b/src/backend/query_builder.rs
@@ -1,6 +1,6 @@
 use std::{borrow::Cow, fmt, ops::Deref};
 
-use crate::*;
+use crate::{utils::JoinWrite, *};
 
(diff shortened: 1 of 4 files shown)


tyt2y3 (approved):
Looks good. The Postgres ON CONFLICT target expression now gets the needed parentheses, with no public API change.

PR #1067, opened 2026-04-16
Generate invalid sql for `IndexCreateStatement::cond_where` on MySQL …

### LINKED ISSUE #1066: MySQL backend silently drops cond_where on IndexCreateStatement
## Description

The MySQL backend silently ignores `cond_where` on `IndexCreateStatement`. When building a partial/filtered unique index, the WHERE clause is dropped without any error or warning, producing a semantically different index than intended.

The root cause is that `MysqlQueryBuilder::prepare_index_create_statement` (`sea-query/src/backend/mysql/index.rs`, lines 27-50) never calls `prepare_filter`, while both `PostgresQueryBuilder` and `SqliteQueryBuilder` do.

## Steps to Reproduce

1. Build an `IndexCreateStatement` with `cond_where`:
   ```rust
   Index::create()
       .name("ux_one_primary_per_identity")
       .table(Emails::Table)
       .col(Emails::IdentityId)
       .unique()
       .cond_where(Expr::col(Emails::Primary).eq(true))
       .to_owned()
   ```
2. Execute the migration against a MySQL backend.

### Expected Behavior

An error or panic indicating that MySQL does not support partial indexes, since silently dropping the WHERE clause produces a plain unique index with completely different semantics. In this example, instead of enforcing "one primary email per identity," it enforces "one email per identity", which is a data-integrity hazard.

### Actual Behavior

The WHERE clause is silently dropped. The generated SQL is `CREATE UNIQUE INDEX ... ON ... (identity_id)` with no WHERE clause and no diagnostic of any kind.

### Reproduces How Often

100%. any `cond_where` on any index against the MySQL backend is silently ignored.

### Workaround

Check the backend before creating the index and skip it on MySQL:

```rust
if manager.get_database_backend() != DatabaseBackend::MySql {
    manager
        .create_index(
            Index::create()
                .name("ux_one_primary_per_identity")
                .table(Emails::Table)
                .col(Emails::IdentityId)
                .unique()
                .cond_where(Expr::col(Emails::Primary).eq(true))
                .to_owned(),
        )
        .await?;
}
```

This means MySQL simply won't have the constraint enforced at the database level.

Huliiiiii:
Closes #1066

### DIFF
--- a/src/backend/index_builder.rs
+++ b/src/backend/index_builder.rs
@@ -103,6 +103,6 @@ pub trait IndexBuilder: QuotedBuilder + TableRefBuilder {
     }
 
     #[doc(hidden)]
-    // Write WHERE clause for partial index. This function is not available in MySQL.
+    // Write WHERE clause for partial index.
     fn prepare_filter(&self, _condition: &ConditionHolder, _sql: &mut impl SqlWriter) {}
 }
--- a/src/backend/mysql/index.rs
+++ b/src/backend/mysql/index.rs
@@ -22,6 +22,7 @@ impl IndexBuilder for MysqlQueryBuilder {
         }
 
         self.prepare_index_columns(&create.index.columns, sql);
+        self.prepare_filter(&create.r#where, sql);
     }
 
     fn prepare_index_create_statement(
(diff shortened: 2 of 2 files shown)


Expurple (commented):
I would explicitly document on those methods why we intentionally generate invalid SQL for MySQL, and reference the issue.

The decision itself is OK to me. Probably makes more sense to crash the query, than to crash the entire app with a panic.

tyt2y3 (approved):
Thanks!

PR #1069, opened 2026-04-30
Add Postgres-only SQLx bindings crate

caniko:
## Summary

This adds a small Postgres-only SQLx binding crate for SeaQuery.

The existing `sea-query-sqlx` crate is intentionally database-agnostic, but that also means its manifest has to mention every SQLx backend it can support. Cargo records optional dependencies from package manifests in downstream lockfiles, so a project that only enables Postgres can still end up with unrelated SQLx packages such as `sqlx-mysql` in `Cargo.lock`.

That is awkward for projects that keep a tight dependency/audit surface. In our case, the MySQL edge pulls in `rsa` even though no MySQL code is compiled or used.

## What changed

- Added `sea-query-sqlx-postgres`, a Postgres-only companion crate.
- Kept the familiar `SqlxBinder` and `SqlxValues` shape so callers can continue to build a query and pass the values into SQLx.
- Wired the crate directly to `sqlx-core` and `sqlx-postgres`, without depending on the top-level `sqlx` facade.
- Preserved the common SeaQuery value feature flags for JSON, UUID, chrono, decimals, time, IP networks, MAC addresses, and Postgres arrays.

One limitation is called out in the implementation: `pgvector`'s SQLx integration currently depends on the `sqlx` facade, so this crate does not wire that integration yet. I kept that separate from the initial change so this PR can stay focused on the dependency shape.

## Why this shape

I first tried changing `sea-query-sqlx` itself to use direct SQLx driver crates, but that does not solve the lockfile issue. Cargo still records optional dependencies listed by the package, even when those features are not enabled. A separate per-driver package avoids that entirely because a Postgres-only consumer never sees the MySQL dependency in the first place.

## Checks

- `cargo check --manifest-path sea-query-sqlx-postgres/Cargo.toml --no-default-features --features with-json,with-uuid,with-chrono,with-rust_decimal,with-bigdecimal,postgres-array,runtime-tokio-rustls`
- `cargo check --manifest-path sea-query-sqlx-postgres/Cargo.toml --all-features`
- `cargo fmt --manifest-path sea-query-sqlx-postgres/Cargo.toml -- --check`
- Verified in a temporary downstream crate that enabling the new Postgres crate does not add `sqlx`, `sqlx-mysql`, `sqlx-macros`, or `rsa` to `Cargo.lock`.

### DIFF
--- a/sea-query-sqlx-postgres/Cargo.toml
+++ b/sea-query-sqlx-postgres/Cargo.toml
@@ -1,0 +1,65 @@
+[workspace]
+# A separate workspace
+
+[package]
+name = "sea-query-sqlx-postgres"
+version = "0.1.0"
+authors = ["SeaQL Contributors"]
+edition = "2024"
+description = "PostgreSQL-only SQLx bindings for SeaQuery"
+license = "MIT OR Apache-2.0"
+documentation = "https://docs.rs/sea-query"
+repository = "https://github.com/SeaQL/sea-query"
+categories = ["database"]
+keywords = ["database", "sql", "postgres", "sqlx"]
+rust-version = "1.88.0"
+
+[lib]
+
+[dependencies]
+sea-query = { version = "1.0.0-rc.33", path = "..", default-features = false, features = [
+    "backend-postgres",
+    "thread-safe",
+] }
+sqlx-core = { version = "0.8", default-features = false }
+sqlx-postgres = { version = "0.8", default-features = false }
+
+[features]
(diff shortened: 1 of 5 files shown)


caniko:
Marking as ready for review. @tyt2y3 @Huliiiiii @Expurple — would appreciate any thoughts on the per-driver crate shape vs. trying to slim `sea-query-sqlx` itself. Happy to iterate on naming or structure.

PR #1070, opened 2026-04-30
Fix null element in array panic (#1068)

mohs8421:
When a value in the array is null, the system panicked. Now which behavior is expected? Ignoring the value seems more appropriate in the given context. The proper approach to this would probably be to implement a conversion to Vec<Option<T>> instead.

## PR Info

- Closes [#1068](https://github.com/SeaQL/sea-query/issues/1068)

## Bug Fixes

- [Fixing panic, when an array contains null elements]

## Breaking Changes

- [It might be considered breaking, when users actually expect this to panic, as the null array element is now ignored in this conversion]


mohs8421:
@Huliiiiii take a look at this fix, it might not be the desired behavior as it is due to the fact that this now ignores the null value.

Huliiiiii:
I think this can return a ValueTypeErr instead.

mohs8421:
> I think this can return a ValueTypeErr instead.

Might be fine in a context, where you expect to have values in the arrays, if you are in a different context where a null value might occur and be valid, this would still be too strict.
The proper way to resolve this would be to implement a translation into `Vec<Option<T>>` and either stick with the original behavior, or warn / fail at compiletime, when a nullable item is used in a non-nullable context.
That way users would have the decision on their end.

Huliiiiii:
Yes, a conversion between `Vec<Option<T>>` should be added.

I think changing the current behavior to return an error is acceptable. A runtime error is better than a panic.

mohs8421:
alright, I can do that

mohs8421:
This should be the desired behavior now, it might still be interesting to see, if the original test of the issue still panics due to the unwrapping happening there now a further level up.

src/value/postgres_array.rs
```diff
@@ -104,11 +104,15 @@ where
...
-                .map(|e| e.unwrap())
-                .collect()),
+            Value::Array(ty, Some(v)) if T::array_type() == ty => {
+                if v.iter().any(|p| p == &p.as_null()) {
+                    return Err(ValueTypeErr);
+                }
+                Ok(v.into_iter()
+                    .filter(|p| p != &p.as_null())
+                    .map(|e| e.unwrap())
+                    .collect())
```
Huliiiiii:
```suggestion
                v.into_iter()
                    .map(|e| {
                        if e == e.as_null() {
                            Err(ValueTypeErr)
                        } else {
                            Ok(e.unwrap())
                        }
                    })
                    .collect()
```

src/value/postgres_array.rs
```diff
@@ -99,20 +99,54 @@ where
...
 impl<T> ValueType for Vec<T>
+where
+    T: NotU8 + ValueType,
+{
+    fn try_from(v: Value) -> Result<Self, ValueTypeErr> {
+        match v {
+            Value::Array(ty, Some(v)) if T::array_type() == ty => v
+                .into_iter()
+                .map(|e| {
+                    if e == e.as_null() {
```
Huliiiiii: I think we could add an is_null method.

Huliiiiii (changes requested):

Huliiiiii:
I think it would be better to write this explicitly as `Option::is_some`, so we won’t forget to update this method when adding new variants to Value.

mohs8421:
I did that change, the error you see on the last commit is comming from 0581311ea9445cf622e789b704e24075b901e907 as I had to resync the branch.

src/value.rs
```diff
@@ -503,128 +503,128 @@ impl Value {
...
             #[cfg_attr(docsrs, doc(cfg(feature = "with-mac_address")))]
-            Self::MacAddress(None) => true,
+            Self::MacAddress(None) => false,
 
             #[cfg(feature = "postgres-range")]
             #[cfg_attr(docsrs, doc(cfg(feature = "postgres-range")))]
-            Self::Range(None) => true,
-            _ => false,
+            Self::Range(None) => false,
+            _ => true,
```
Huliiiiii: Sorry, my explanation wasn't clear enough. What I meant was `Self::Variant(v) => v.is_none()`, not renaming this method to `is_some`.

Huliiiiii: I also got the conditions backwards, sorry.

Huliiiiii (approved):

mohs8421:
@tyt2y3 this change is still waiting for your review.

PR #1071, opened 2026-05-13
Fix RETURNING for MariaDB

### LINKED ISSUE #2709: Allow `exec_with_returning` with MariaDB
## Motivation

Currently, when using a MariaDB as backend, it is hidden behind the MySql backend, which is normal. However, there are things that are not available in MySQL but in MariaDB, like the `INSERT ... RETURNING` syntax. Nevertheless, the methods building queries with the "returning" syntax are configured to be only available with the Postgres and Sqlite backend (see https://docs.rs/sea-orm/latest/src/sea_orm/database/db_connection.rs.html#585).

## Proposed Solutions

Just add a match arm in the [`DbBackend::support_returning`](https://docs.rs/sea-orm/latest/sea_orm/type.DbBackend.html#method.support_returning) method, for the MySql backend, with an additional feature, similar to `"sqlite-use-returning-for-3_35"`, e.g. `"mariadb"`.

This way, developers using MariaDB will be able to use the "returning" syntax, by enabling this feature.

Tnze:
## PR Info

- Closes [Allow exec_with_returning_many with MariaDB](https://github.com/SeaQL/sea-orm/issues/2709)

## New Features

- [ ] Allow exec_with_returning_many with MariaDB

### DIFF
--- a/src/backend/mysql/query.rs
+++ b/src/backend/mysql/query.rs
@@ -281,8 +281,6 @@ impl QueryBuilder for MysqlQueryBuilder {
 
     fn prepare_on_conflict_condition(&self, _: &ConditionHolder, _: &mut impl SqlWriter) {}
 
-    fn prepare_returning(&self, _returning: &Option<ReturningClause>, _sql: &mut impl SqlWriter) {}
-
     fn random_function(&self) -> &str {
         "RAND"
     }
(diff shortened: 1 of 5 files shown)


Tnze:
I'm not sure if we should make a new feature to control this behavior.

PR #1072, opened 2026-05-15
Remove `inherent`

Suficio:
I recently wanted to out this project however the additional dependency was not desirable, especially since `inherent` utilizes proc-macros making it a relatively heavy dependency to import.

I realized you can rather easily cover the current `inherent` coverage with a couple manual implementations without relying on shims as in the past: #614.

It also allows going back to the original claim of "no dependencies".

## Changes

- Removes `inherent` in favor of manual function implementations.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,6 @@ name = "sea_query"
 path = "src/lib.rs"
 
 [dependencies]
-inherent = "1.0"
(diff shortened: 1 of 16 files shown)


tyt2y3 (approved):
that's fair. we originally use this on many expr methods, but we don't anymore

PR #1074, opened 2026-05-15
Fix: tokenizer treating `[` as a quote start for postgres (#1073)

### LINKED ISSUE #1073: PostgreSQL `cust_with_values` fails on ARRAY[...] literals (placeholder swallowing)
# Bug Report: `cust_with_values` drops placeholders inside `[...]` groups (e.g. `ARRAY[...]`)

## Description

`Expr::cust_with_values` silently drops bind values whose `$N` (or `?`) placeholders are written inside a `[...]` group, most notably a PostgreSQL `ARRAY[...]` literal or a subscript expression. The rendered SQL contains a stray `[...]` substring instead of the substituted placeholders, the values vector is shorter than the prepared statement expects, and execution fails with a parameter-count mismatch (or, in the worst case, binds against the wrong column when the count happens to line up).

Root cause: the SQL tokenizer in `src/token.rs` treats `[` as a string-quoting delimiter for every backend. Inside `Expr::cust_with_values`, the tokenizer is used to walk the input and substitute `$N` / `?`. Anything between `[` and `]` is consumed as a single `Token::Quoted` and never reaches the placeholder-substitution loop, so the placeholders are emitted verbatim and the corresponding values stay in the input vec instead of being routed into the bind vec at the right offset.

The `[col]` bracketed-identifier convention only exists in MSSQL / T-SQL, and as an MS Access-compatibility mode in SQLite. MySQL has no `[col]` syntax (it uses backticks). PostgreSQL not only doesn't use `[col]` but actively uses `[...]` for `ARRAY` literals and subscript syntax (`arr[1]`, `arr[1:3]`), which is where this bug is most visible.

## Steps to Reproduce

```rust
let (statement, values) = Query::select()
    .column(Char::Character)
    .from(Char::Table)
    .and_where(Expr::cust_with_values(
        "tags @> ARRAY[$1, $1]",
        ["needle"],
    ))
    .build(PostgresQueryBuilder);

println!("{statement}");
println!("{values:?}");
```

### Expected Behavior

```text
SELECT "character" FROM "character" WHERE tags @> ARRAY[$1, $2]
values: ["needle", "needle"]
```

### Actual Behavior (1.0.0-rc.34, current master)

```text
SELECT "character" FROM "character" WHERE tags @> ARRAY[$1, $1]
values: []
```

PostgreSQL rejects the prepared statement with a parameter-count mismatch. The same failure shape applies to any `cust_with_values` input that places `$N` / `?` inside `[...]`, e.g. paradedb / pg_search predicates like:

```sql
id @@@ paradedb.boolean(should => ARRAY[
    paradedb.match('body', $1, distance => 1),
    paradedb.match('body_ngram', $1)
])
```

### Reproduces How Often

Always, whenever `cust_with_values` input contains a placeholder between `[` and `]`.

## Versions

Reproduced on `sea-query 1.0.0-rc.34` (current `master`).

The same bug exists in the released `0.32.x` line, but that release line lacks the `TokenizerBackend` machinery added since and needs a different fix shape. A separate issue / PR could cover the 0.32.x situation; this one is scoped to `master`.

Database: PostgreSQL 18 with paradedb / pg_search. OS: Linux.

## Fix shape

A PR against `master` is on the way. The backend-aware `TokenizerBackend::{Mysql, Postgres, Sqlite}` plumbing already exists, so the fix is small: `is_string_delimiter_start` becomes a `&self` method that returns `false` for `[` when `self.backend == TokenizerBackend::Postgres`. MySQL and SQLite are left byte-identical to today's behavior. A new tokenizer-level test (`ARRAY[$1, $2]` survives as separate tokens under Postgres) and an `Expr::cust_with_values` end-to-end regression test are included.

## Additional Information

Workaround in user code: render the SeaQuery to SQL via `build_any(...)`, splice the `[...]` bearing fragment in manually, and push the values onto the bind vec at the right offsets. It works but is messy and reqires handling raw string query fragments. Fixing this issue would eliminate such workarounds.

cavebatsofware:
# Stop treating `[` as a string-quote start under PostgreSQL

Fixes #1073.

## Summary

Stops the SQL tokenizer from consuming `ARRAY[...]` / `arr[i]` and similar bracket-bounded fragments as a single `Token::Quoted` when rendering under the PostgreSQL backend. This fixes the `Expr::cust_with_values` bug for any input that places `$N` placeholders between `[` and `]`. MySQL and SQLite are left byte-identical to today.

## What changed

* `Tokenizer::is_string_delimiter_start` is now a `&self` method (matching the existing `&self` shape of `uses_backslash_escape_for` and `next_single_quote_is_postgres_escape_string`). It returns `false` for `[` when `self.backend == TokenizerBackend::Postgres`, and otherwise returns the historical `matches!(c, ...)` result.
* Two call sites in `Tokenizer::quoted` / `Tokenizer::unquote` updated to `self.is_string_delimiter_start(c)`.
* `is_string_delimiter_end_for` is unchanged. Once a `[` quote has been opened (only possible for MySQL/SQLite under the new gate), the close is still `]`.

## Why Postgres-only

`[` is overloaded across dialects:

| Backend | `ARRAY[...]` literal | `[col]` identifier quoting |
| --- | --- | --- |
| Postgres | yes | no |
| MySQL | no | no (MySQL uses backticks) |
| SQLite | no | yes (documented MS Access compat) |

The historical tokenizer treated `[` as a quote start unconditionally for every backend which is a quirk that affected MySQL incidentally even though MySQL has no `[col]` syntax. This PR keeps that historical behavior for MySQL and SQLite, and limits the change to PostgreSQL where `[` is core syntax for `ARRAY[...]` and subscript expressions. 

A follow-up that also removes `[` for MySQL (where it seems to serve no purpose) is reasonable but out of scope here, mostly because I'm not 100% sure that there is no case where MySQL would ever use any [$X] expression without further research that I have not done. I know they use [,] in regex but those are quoted already within `` so while I do think it would be safe I don't know for sure and my use case and expertise is with PostgreSQL.

## Tests

* `token::tests::test_9` unchanged (default backend is MySQL, `[ab]` is still one `Quoted` token).
* `token::tests::test_9_postgres_does_not_treat_brackets_as_quoted` new. Asserts `ARRAY[$1, $2]` tokenizes such that the `$1` and `$2` survive as separate tokens under `TokenizerBackend::Postgres`.
* `token::tests::test_9_mysql_and_sqlite_keep_bracketed_quoting` new. Locks in that MySQL and SQLite still produce `Token::Quoted("[ab]")`.
* `tests/postgres/query.rs::cust_with_values_inside_array_literal_renumbers_placeholders` new. End-to-end through `Expr::cust_with_values` + `PostgresQueryBuilder`:

```rust
let (statement, values) = Query::select()
    .column(Char::Character)
    .from(Char::Table)
    .and_where(Expr::cust_with_values(
        "tags @> ARRAY[$1, $1]",
        ["needle"],
    ))
    .build(PostgresQueryBuilder);

assert_eq!(statement, r#"SELECT "character" FROM "character" WHERE tags @> ARRAY[$1, $2]"#);
assert_eq!(values, Values(vec!["needle".into(), "needle".into()]));
```

## Changelog

```markdown
### Fixes

* Fixed `Expr::cust_with_values` dropping placeholders inside `ARRAY[...]` literals and other `[...]` groups under `PostgresQueryBuilder` https://github.com/SeaQL/sea-query/pull/<n>
```

### DIFF
--- a/src/token.rs
+++ b/src/token.rs
@@ -152,7 +152,7 @@ impl<'a> Tokenizer<'a> {
         let mut uses_backslash_escape = false;
         while !self.end() {
             let c = self.get();
-            if first && Self::is_string_delimiter_start(c) {
+            if first && self.is_string_delimiter_start(c) {
                 b = self.p_c(c);
                 first = false;
                 start = c;
@@ -195,7 +195,7 @@ impl<'a> Tokenizer<'a> {
         let mut uses_backslash_escape = false;
         while !self.end() {
             let c = self.get();
-            if first && Self::is_string_delimiter_start(c) {
+            if first && self.is_string_delimiter_start(c) {
                 first = false;
                 start = c;
                 uses_backslash_escape = self.uses_backslash_escape_for(start);
@@ -286,8 +286,12 @@ impl<'a> Tokenizer<'a> {
         c.is_alphabetic() || c.is_ascii_digit()
     }
 
-    fn is_string_delimiter_start(c: char) -> bool {
-        matches!(c, '`' | '[' | '\'' | '"')
+    fn is_string_delimiter_start(&self, c: char) -> bool {
+        match c {
+            '`' | '\'' | '"' => true,
+            '[' => self.backend != TokenizerBackend::Postgres,
+            _ => false,
+        }
     }
 
     fn is_string_escape_for(start: char, c: char) -> bool {
@@ -566,6 +570,41 @@ mod tests {
             string,
             tokens.iter().map(|x| x.as_str()).collect::<String>()
         );
+    }
+
+    #[test]
+    fn test_9_postgres_does_not_treat_brackets_as_quoted() {
+        let string = r"ARRAY[$1, $2]";
+        let tokenizer = Tokenizer::new(string).for_backend(TokenizerBackend::Postgres);
+        let tokens: Vec<Token> = tokenizer.iter().collect();
+        assert_eq!(
+            tokens,
+            vec![
+                Token::Unquoted("ARRAY"),
+                Token::Punctuation("["),
+                Token::Punctuation("$"),
+                Token::Unquoted("1"),
+                Token::Punctuation(","),
+                Token::Space(" "),
+                Token::Punctuation("$"),
+                Token::Unquoted("2"),
+                Token::Punctuation("]"),
+            ]
+        );
+        assert_eq!(
+            string,
+            tokens.iter().map(|x| x.as_str()).collect::<String>()
+        );
+    }
+
+    #[test]
+    fn test_9_mysql_and_sqlite_keep_bracketed_quoting() {
+        for backend in [TokenizerBackend::Mysql, TokenizerBackend::Sqlite] {
+            let string = r"[ab]";
+            let tokenizer = Tokenizer::new(string).for_backend(backend);
+            let tokens: Vec<Token> = tokenizer.iter().collect();
+            assert_eq!(tokens, vec![Token::Quoted("[ab]")], "backend={:?}", backend);
+        }
     }
 
     #[test]
--- a/tests/postgres/query.rs
+++ b/tests/postgres/query.rs
@@ -1204,6 +1204,21 @@ fn select_61_custom_values_numbered_placeholders_after_escape_string() {
 }
 
 #[test]
+fn cust_with_values_inside_array_literal_renumbers_placeholders() {
+    let (statement, values) = Query::select()
+        .column(Char::Character)
+        .from(Char::Table)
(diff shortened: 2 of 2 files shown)


cavebatsofware:
Force pushed ^ cargo fmt fix that got left out. 
`cargo clippy --features=all-features --workspace -- -D warnings -A clippy::semicolon_if_nothing_returned` doesn't show any issues for files touched by this PR.

Huliiiiii (approved):

cavebatsofware:
@tyt2y3 Is this something that you want in the next update?

PR #1080, opened 2026-06-03
Feat/postgres create function and triggers

Ayanrocks:
Gives users the ablity to create functions and triggers on database by using sea-query instead of raw SQL.
Added examples of all the functionality in `examples/postgres`

## PR Info

- Closes  #403 

- Dependencies:
  -  NO

- Dependents:
  -  NO

## New Features

- [x] Added support for CREATE FUNCTION in Postgres
- [x] Added support for ALTER FUNCTION in Postgres
- [x] Added support for DROP FUNCTION in Postgres
- [x] Added support for CREATE TRIGGER in Postgres
- [x] Added support for ALTER TRIGGER in Postgres
- [x] Added support for DROP TRIGGER in Postgres

## Bug Fixes
N/A
 

## Breaking Changes
No
 

## Changes
No

### DIFF
--- a/src/backend/postgres/mod.rs
+++ b/src/backend/postgres/mod.rs
@@ -7,7 +7,9 @@ pub(crate) mod types;
 
 use super::*;
 use crate::extension::postgres::{
-    function::alter::{FunctionAlterStatement, FunctionConfigOption, FunctionDependsOption},
+    function::alter::{
+        FunctionAlterOption, FunctionAlterStatement, FunctionConfigOption, FunctionDependsOption,
+    },
     function::create::{
         FunctionArgMode, FunctionBehavior, FunctionCreateStatement, FunctionReturns,
     },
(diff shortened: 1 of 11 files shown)


src/extension/postgres/function/create.rs
```diff
@@ -0,0 +1,145 @@
+use crate::{ColumnType, DynIden, Expr, IntoIden};
+
+/// Represents PostgreSQL function argument modes
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum FunctionArgMode {
+    In,
+    Out,
+    Inout,
```
Huliiiiii: InOut

src/extension/postgres/function/alter.rs
```diff
@@ -0,0 +1,211 @@
...
+    ResetAll,
+}
+
+/// Creates a new "ALTER FUNCTION" statement for PostgreSQL
+#[derive(Debug, Default, Clone, PartialEq)]
+pub struct FunctionAlterStatement {
+    pub(crate) name: Option<DynIden>,
+    pub(crate) if_exists: bool,
+    pub(crate) arg_types: Option<Vec<ColumnType>>,
+    pub(crate) options: Vec<FunctionAlterOption>,
```
Huliiiiii: I think it's better to use a set or a struct here to prevent duplicate options in the current api.

src/backend/postgres/mod.rs
```diff
@@ -6,11 +6,432 @@ pub(crate) mod table;
...
+                if i > 0 {
+                    sql.write_str(", ").unwrap();
+                }
+                self.prepare_column_type(ty, sql);
+            }
+            sql.write_str(")").unwrap();
+        }
+        for (i, option) in alter.options.iter().enumerate() {
+            if i > 0 {
+                sql.write_str(", ").unwrap();
```
Huliiiiii: They should not be separated by commas.

src/extension/postgres/trigger/create.rs
```diff
@@ -0,0 +1,221 @@
...
+    pub(crate) execution_type: Option<TriggerExecutionType>,
+}
+
+impl TriggerCreateStatement {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Set the trigger name
+    pub fn name<T: IntoIden>(&mut self, name: T) -> &mut Self {
```
Huliiiiii: In simple cases, you can use `impl trait` instead of generics.

src/extension/postgres/trigger.rs
```diff
@@ -0,0 +1,536 @@
...
+impl PgTriggerStmt {
+    /// Creates a new [`TriggerCreateStatement`]
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use sea_query::{*, extension::postgres::*, tests_cfg::*};
+    ///
+    /// let create = PgTriggerStmt::create()
+    ///     .name(Alias::new("my_trigger"))
```
Huliiiiii: &'static str already implements IntoIden, so we no longer need to construct an Alias manually.

src/extension/postgres/trigger/create.rs
```diff
@@ -0,0 +1,221 @@
...
+
+    /// Set trigger execution function
+    pub fn function<T: IntoIden>(&mut self, name: T) -> &mut Self {
+        self.function = Some(name.into_iden());
+        self.execution_type = Some(TriggerExecutionType::Function);
+        self
+    }
+
+    /// Set trigger execution procedure
+    pub fn procedure<T: IntoIden>(&mut self, name: T) -> &mut Self {
```
Huliiiiii: `procedure` has been deprecated by postgres.
> In the syntax of CREATE TRIGGER, the keywords FUNCTION and PROCEDURE are equivalent, but the referenced function must in any case be a function, not a procedure. The use of the keyword PROCEDURE here is historical and deprecated.

src/extension/postgres/function/alter.rs
Ayanrocks: We can add a deduplication logic or go with struct as for hashset, `f64` doesn't emplement the `Eq/Hash` traits. Let me know which one sounds good

src/extension/postgres/trigger/create.rs
Huliiiiii: I think we can just remove it. There's no need to add an api that postgres has already deprecated and that doesn't affect functionality.

src/extension/postgres/function/alter.rs
Ayanrocks: @Huliiiiii Any update? should I go ahead with deduplication or we want to keep it as a struct?

Huliiiiii: I think options that can only appear once would be better modeled as fields on the struct. For others, it may be cleaner to split it based on mutual exclusivity rather than putting everything into a single enum.

Ayanrocks:
Hey @Expurple, Got a chance to review it?

src/extension/postgres/function/alter.rs
```diff
@@ -0,0 +1,250 @@
...
+/// Creates a new "ALTER FUNCTION" statement for PostgreSQL
+#[derive(Debug, Default, Clone, PartialEq)]
+pub struct FunctionAlterStatement {
+    pub(crate) name: Option<DynIden>,
+    pub(crate) if_exists: bool,
+    pub(crate) arg_types: Option<Vec<ColumnType>>,
+
+    pub(crate) rename_to: Option<DynIden>,
+    pub(crate) owner_to: Option<DynIden>,
+    pub(crate) set_schema: Option<DynIden>,
```
Huliiiiii: `action`, `RENAME TO`, `OWNER TO`, `SET SCHEMA`, and `DEPENDS ON EXTENSION` are mutually exclusive.
https://www.postgresql.org/docs/current/sql-alterfunction.html

src/extension/postgres/function/create.rs
```diff
@@ -0,0 +1,145 @@
...
+
+/// Represents the return type of a PostgreSQL function
+#[derive(Debug, Clone, PartialEq)]
+pub enum FunctionReturns {
+    Type(ColumnType),
+    Table(Vec<(DynIden, ColumnType)>),
+}
+
+/// Creates a new "CREATE FUNCTION" statement for PostgreSQL
+#[derive(Debug, Default, Clone, PartialEq)]
```
Huliiiiii, line 75: Add an optional schema field

Ayanrocks:
Any update on this PR? @Huliiiiii @Expurple

src/backend/postgres/mod.rs
```diff
@@ -6,11 +6,498 @@ pub(crate) mod table;
...
+                FunctionBehavior::ParallelUnsafe => "PARALLEL UNSAFE",
+                FunctionBehavior::ParallelRestricted => "PARALLEL RESTRICTED",
+                FunctionBehavior::ParallelSafe => "PARALLEL SAFE",
+            })
+            .unwrap();
+        }
+        if let Some(definition) = &create.as_definition {
+            sql.write_str(" AS ").unwrap();
+            self.write_string_quoted(definition, sql);
+        } else if let Some(sql_body) = &create.sql_body {
```
Huliiiiii, line 162: > sql_body
> 
>     The body of a LANGUAGE SQL function. This can either be a single statement
> 
>     RETURN expression
> 
>     or a block
> 
>     BEGIN ATOMIC
>       statement;
>       statement;
>       ...
>       statement;
>     END

Just simply write the user input directly into the SQL.

src/extension/postgres/trigger/create.rs
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    /// Set trigger execution function
+    pub fn function(&mut self, name: impl IntoIden) -> &mut Self {
+        self.function = Some(name.into_iden());
+        self
+    }
+
+    /// Add a trigger execution argument
+    pub fn function_arg(&mut self, arg: impl Into<Expr>) -> &mut Self {
```
Huliiiiii: > arguments
> 
>     An optional comma-separated list of arguments to be provided to the function when the trigger is executed. The arguments are literal string constants. Simple names and numeric constants can be written here, too, but they will all be converted to strings. Please check the description of the implementation language of the trigger function to find out how these arguments can be accessed within the function; it might be different from normal function arguments.

This should be `&'static str` or `String`, not `Expr`.

Ayanrocks:
@Huliiiiii Can you review it please?

Ayanrocks:
Any update on this PR? @Huliiiiii @Expurple

Huliiiiii:
I will continue working on this.

PR #1083, opened 2026-06-11
support `join ... using` (#1046)

### LINKED ISSUE #1046: Support `JOIN ... USING`
## Motivation

I'd like to use shorthand joins, so instead of

```sql
INNER JOIN table ON (table.x = other.x AND table.y = other.y)
```

I could do

```sql
INNER JOIN table USING (x, y)
```

This makes for easier to read queries in logs, but is ofc just syntactic sugar (with one rarely useful feature of not duplicating the columns in `SELECT *`).

## Proposed Solutions

Glancing at the code, looks like making `JoinOn::Columns` generate code should be enough (as long as the struct/field naming isn't a problem) - https://github.com/SeaQL/sea-query/blob/master/src/backend/query_builder.rs#L1012
With a new `SelectStatement` method like `.join_using`.

## Additional Information

Postgres, MySQL and SQLite all support this. MS SQL doesn't, but afaik that's not in the scope of this repo.

w1ll-i-code:
## PR Info

- Closes #1046

## New Features

- [x] Supports using `JOIN ... USING`

### DIFF
--- a/src/query/select.rs
+++ b/src/query/select.rs
@@ -1925,6 +1925,43 @@ impl SelectStatement {
         )
     }
 
+    /// Join with a "USING" clause.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use sea_query::{tests_cfg::*, *};
+    ///
+    /// let query = Query::select()
+    ///     .column(Char::Character)
+    ///     .from(Char::Table)
+    ///     .join_using(JoinType::LeftJoin, Font::Table, [Char::FontId])
+    ///     .to_owned();
+    ///
+    /// assert_eq!(
+    ///     query.to_string(MysqlQueryBuilder),
+    ///     r#"SELECT `character` FROM `character` LEFT JOIN `font` USING (`font_id`)"#
+    /// );
+    /// assert_eq!(
+    ///     query.to_string(PostgresQueryBuilder),
+    ///     r#"SELECT "character" FROM "character" LEFT JOIN "font" USING ("font_id")"#
+    /// );
+    /// ```
+    pub fn join_using(
+        &mut self,
+        join: JoinType,
+        tbl_ref: impl IntoTableRef,
+        columns: impl IntoIterator<Item = impl IntoColumnRef>,
+    ) -> &mut Self {
+        let columns = columns
+            .into_iter()
+            .map(IntoColumnRef::into_column_ref)
+            .map(Expr::from)
+            .collect();
+
+        self.push_join(join, tbl_ref.into(), JoinOn::Columns(columns), false)
+    }
+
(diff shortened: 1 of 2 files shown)


Huliiiiii (approved):

src/query/select.rs
```diff
@@ -1925,6 +1925,46 @@ impl SelectStatement {
...
+    /// assert_eq!(
+    ///     query.to_string(MysqlQueryBuilder),
+    ///     r#"SELECT `character` FROM `character` LEFT JOIN `font` USING (`font_id`)"#
+    /// );
+    /// assert_eq!(
+    ///     query.to_string(PostgresQueryBuilder),
+    ///     r#"SELECT "character" FROM "character" LEFT JOIN "font" USING ("font_id")"#
+    /// );
+    /// ```
+    pub fn join_using<T, R>(&mut self, join: JoinType, tbl_ref: R, columns: Vec<T>) -> &mut Self
```
Huliiiiii: Maybe `impl IntoIterator<Item = impl IntoColumnRef>` would be better? In most cases, using array syntax would be more concise.

w1ll-i-code: That's a great idea. I was already wondering if there was a better solution when trying it in my own code base. Fixed, thank you.

src/query/select.rs
```diff
@@ -1925,6 +1925,47 @@ impl SelectStatement {
...
+    /// assert_eq!(
+    ///     query.to_string(PostgresQueryBuilder),
+    ///     r#"SELECT "character" FROM "character" LEFT JOIN "font" USING ("font_id")"#
+    /// );
+    /// ```
+    pub fn join_using<R, T, I>(&mut self, join: JoinType, tbl_ref: R, columns: I) -> &mut Self
+    where
+        R: IntoTableRef,
+        T: IntoColumnRef,
+        I: IntoIterator<Item = T>,
```
Huliiiiii:
```suggestion
    pub fn join_using(&mut self, join: JoinType, tbl_ref: impl IntoTableRef, columns: impl IntoIterator<Item = impl IntoColumnRef>) -> &mut Self
```
Does this work on sea-query's MSRV?

w1ll-i-code: It does, but it'd be inconsistent with the surrounding code, which all use the explicit syntax. I still can change that, if that's the direction the project wants to take.

Huliiiiii: Using `impl Trait` in new code is fine. The old code was written before Rust supported this syntax.

w1ll-i-code: Fixed, as well as the formatting issue in the tests. There seem to be problems with the builds however. Looks like a dependency broke.

PR #1084, opened 2026-06-23
Make integer ValueType::try_from lenient via checked conversion

devgony:
Integer types only accept their own `Value` variant on read, so
`i32::try_from(Value::BigInt(..))` fails even when the number fits. This breaks
the mock/proxy backends on databases with a single integer width — e.g. Spanner
only has `INT64`, so the proxy can only emit `Value::BigInt` and an `i32` field
can't be read.

This adds an `int_type_to_value!` macro for the 8 integer types. It's identical
to `type_to_value!` except `try_from` accepts any integer variant and converts
with the std checked `TryFrom`. Out-of-range or NULL still returns
`ValueTypeErr`, so nothing is truncated. Writes, `Nullable` and column types are
unchanged.

Note: this is a behavior change (not an API break) — reads that used to fail now
succeed when the value fits, including across the signed/unsigned boundary
(still range-checked). Happy to restrict it to same-signedness variants if you
prefer.

Tested with a new `integer_value_type_is_lenient_and_checked` case;
`cargo test --lib`, `--doc`, `clippy` and `fmt` all pass.

### DIFF
--- a/src/value.rs
+++ b/src/value.rs
@@ -1135,15 +1135,81 @@ macro_rules! type_to_value {
 #[allow(unused_imports)]
 use type_to_value;
 
+/// Like [`type_to_value!`], but for the integer scalar types. Identical in every
+/// respect (`From`, `Nullable`, `type_name`, `array_type`, `column_type`) except
+/// that `try_from` accepts *any* integer `Value` variant and converts it with a
+/// *checked* std conversion (`<$type>::try_from`). This lets an integer be read
+/// from a variant other than its own — e.g. an `i32` model field from a
+/// `Value::BigInt` emitted by a backend whose database has only one integer width
+/// (such as Spanner's `INT64`). NULL and out-of-range both map to `ValueTypeErr`,
+/// preserving the original strict behaviour for those cases (no silent truncation).
+macro_rules! int_type_to_value {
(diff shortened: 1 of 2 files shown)


Huliiiiii:
Please write the PR description in your own language, not in the LLM-generated text, which is difficult for humans to read.

src/value.rs
```diff
@@ -1135,15 +1135,81 @@ macro_rules! type_to_value {
...
 use type_to_value;
 
+/// Like [`type_to_value!`], but for the integer scalar types. Identical in every
+/// respect (`From`, `Nullable`, `type_name`, `array_type`, `column_type`) except
+/// that `try_from` accepts *any* integer `Value` variant and converts it with a
+/// *checked* std `TryFrom` conversion. This lets an integer be read
+/// from a variant other than its own — e.g. an `i32` model field from a
+/// `Value::BigInt` emitted by a backend whose database has only one integer width
+/// (such as Spanner's `INT64`). NULL and out-of-range both map to `ValueTypeErr`,
+/// preserving the original strict behaviour for those cases (no silent truncation).
```
Huliiiiii: Likewise, please rewrite the comments in a more natural, human-readable way.

Comments are for maintainers and other people to read.

PR #1093, opened 2026-08-04
Fix sea_value_to_json_value returning string "NULL" for None date/time values

### LINKED ISSUE #1092: `sea_value_to_json_value` returns the string "NULL" for `None` values in chrono date/time `Value` variants
## Description

`sea_value_to_json_value` does not special-case `None` for chrono date/time `Value` variants (`ChronoDate`, `ChronoTime`, `ChronoDateTime`, `ChronoDateTimeWithTimeZone`, `ChronoDateTimeUtc`, `ChronoDateTimeLocal`, and the equivalent `with-time` variants). Every other nullable variant (`Bool(None)`, `String(None)`, `Int(None)`, etc.) has an explicit arm returning `Json::Null`, but the chrono/time variants fall through to `CommonSqlQueryBuilder.value_to_string(value)`, which renders a `None` value as the SQL literal `NULL` — so the JSON representation becomes the **string** `"NULL"` instead of a JSON `null`.

This breaks `sea_orm::ActiveModelTrait::from_json` / `set_from_json` (sea-orm 2.0.1) for any entity with a nullable chrono column, whenever that column's key is omitted from the input JSON. `from_json` builds a "defaults" object (containing e.g. ChronoTime(None) as the default for an optional time column) and calls `sea_value_to_json_value` on every column created. It then merges the caller's JSON on top of it; since the buggy `"NULL"` string default is never overridden (because the column key is missing in the JSON input), it gets fed into chrono's `Deserialize` impl and fails to parse.

Relevant code: `src/value/with_json.rs`, function `sea_value_to_json_value`. The `None` arms for `Bool`/`TinyInt`/.../`Json`/`Enum` are listed explicitly at the top of the match; the chrono/time arms further down have no such case and unconditionally call `value_to_string`.

## Steps to Reproduce

1. Construct a `Value::ChronoDateTimeWithTimeZone(None)` (or any other nullable chrono `Value` variant set to `None`).
2. Call `sea_query::sea_value_to_json_value(&value)`.
3. Observe the returned `serde_json::Value` is `Value::String("NULL")` instead of `Value::Null`.

This can also be observed indirectly through sea-orm: define an entity with an `Option<DateTimeWithTimeZone>` column, then call `ActiveModel::from_json(json!({ /* column key omitted */ }))` — it fails with `DbErr::Json("input contains invalid characters")` (a chrono `ParseError` triggered by trying to parse the literal string `"NULL"` as a date).

A minimal, self-contained repro project (no database required) is attached under **Additional Information**.

### Expected Behavior

`sea_value_to_json_value(&Value::ChronoDateTimeWithTimeZone(None))` returns `serde_json::Value::Null`, consistent with every other nullable `Value` variant.

### Actual Behavior

`sea_value_to_json_value(&Value::ChronoDateTimeWithTimeZone(None))` returns `serde_json::Value::String("NULL")`.

Downstream in sea-orm, this causes `ActiveModelTrait::from_json`/`set_from_json` to fail with:

```
Json Error: input contains invalid characters
```

for any entity with a nullable chrono column whose key is missing from the input JSON.

### Reproduces How Often

Always, it's a logic gap in a pure function, not timing- or environment-dependent.

## Versions

```
sea-orm v2.0.1
├── sea-orm-macros v2.0.1 (proc-macro)
│   ├── sea-bae v0.2.2 (proc-macro)
├── sea-query v1.0.1
│   ├── sea-query-derive v1.0.0 (proc-macro)
```

- `sea-orm = "2.0.1"`
- `sea-query = "1.0.1"`
- OS: Linux Mint 22.3 (Zena), kernel 6.17.0-35-generic, x86_64
- No database connection required to reproduce: the bug is in a pure value-conversion function and was also confirmed indirectly via `sea_orm::ActiveModelTrait::from_json`, which likewise needs no live DB connection.

## Additional Information

Minimal reproducible example (single `cargo run`, no DB needed):

```toml
# Cargo.toml
[package]
name = "sea-orm-bug-repro"
version = "0.1.0"
edition = "2021"

[dependencies]
sea-orm = { version = "2.0", features = ["macros"] }
serde = { version = "1.0", features = ["derive"] }
serde_json = "1.0"
```

```rust
// src/main.rs
use sea_orm::entity::prelude::*;
use sea_orm::sea_query::sea_value_to_json_value;
use sea_orm::{ActiveModelTrait, Value};
use serde::{Deserialize, Serialize};
use serde_json::json;

#[derive(Clone, Debug, PartialEq, DeriveEntityModel, Serialize, Deserialize)]
#[sea_orm(table_name = "simple")]
pub struct Model {
    #[sea_orm(primary_key)]
    #[serde(skip_deserializing)]
    pub id: i64,
    pub name: String,
    pub created_at: Option<DateTimeWithTimeZone>,
}

#[derive(Copy, Clone, Debug, EnumIter, DeriveRelation)]
pub enum Relation {}

impl ActiveModelBehavior for ActiveModel {}

fn main() {
    // 1. Direct repro of the value-conversion bug
    let none_date: Value = Value::ChronoDateTimeWithTimeZone(None);
    let json_value = sea_value_to_json_value(&none_date);
    println!("{json_value:?}"); // actual: String("NULL")  |  expected: Null

    // 2. Downstream effect on ActiveModel::from_json
    let input = json!({ "name": "foo" }); // `created_at` key omitted
    match ActiveModel::from_json(input) {
        Ok(am) => println!("OK: {am:?}"),
        Err(e) => println!("ERR: {e}"), // "Json Error: input contains invalid characters"
    }
}
```

Expected output line 1: `Null`
Actual output line 1: `String("NULL")`

Expected output line 2: `OK: ActiveModel { id: NotSet, name: Set("foo"), created_at: NotSet }`
Actual output line 2: `ERR: Json Error: input contains invalid characters`

### Suggested fix

Add explicit `None => Json::Null` arms for the affected chrono/time variants in `sea_value_to_json_value`, mirroring the existing arms for `Bool(None)`, `String(None)`, etc.:

```rust
#[cfg(feature = "with-chrono")]
Value::ChronoDate(None)
| Value::ChronoTime(None)
| Value::ChronoDateTime(None)
| Value::ChronoDateTimeWithTimeZone(None)
| Value::ChronoDateTimeUtc(None)
| Value::ChronoDateTimeLocal(None) => Json::Null,
```

(and the equivalent for the `with-time` variants), placed before the existing `Value::ChronoDateTimeWithTimeZone(_) => CommonSqlQueryBuilder.value_to_string(value).into()` arms so the `None` case is caught first.

sescer:
## PR Info

- Closes #1092
- Dependencies: None
- Dependents: None

## New Features

None

## Bug Fixes

- `sea_value_to_json_value` now returns JSON `null` for `None` chrono, time, and jiff `Value` variants instead of the string `"NULL"`

Previously these variants fell through to `value_to_string`, which emits the SQL literal `NULL` and then becomes a JSON string. That broke sea-orm `ActiveModelTrait::from_json` / `set_from_json` when an optional date/time column key was omitted from the input.

Companion fix: the same gap existed for `with-jiff` variants and is covered here as well.

## Breaking Changes

None

## Changes

- Add explicit `None => Json::Null` arms for chrono / time / jiff variants in `sea_value_to_json_value`
- Add feature-gated unit tests for each family


src/value/with_json.rs
```diff
@@ -119,3 +136,67 @@ pub fn sea_value_to_json_value(value: &Value) -> Json {
         Value::Range(Some(_)) => CommonSqlQueryBuilder.value_to_string(value).into(),
     }
 }
+
+#[cfg(test)]
+mod tests {
```
Huliiiiii: These tests simply mirror the implementation.

Huliiiiii:
Thanks!

PacificBird:
Not to be too pushy, but is there any way this PR can be released in a hotfix sooner rather than later? I can't push any changes to my production system with this bug in place. Even regressing my sea-orm version back to one that used to work for whatever reason does not seem to fix things, although maybe this is a weird stale build artifacts bug. Whatever the case, this bug in completely bricks the ORM as a whole for me at the moment and I think justifies a hotfix.

PR #1100, opened 2026-08-19
Add MySQL -> and ->> JSON field operators

### LINKED ISSUE #631: Add support for `get_json_field` and `cast_json_field` for all backends
## Motivation

With the merging of PR #630, JSON path traversal is now supported as an extension in both SQLite and Postgres. However, as far as I can tell from the [MySQL docs](https://dev.mysql.com/doc/refman/8.0/en/json-search-functions.html), the same syntax is also supported in MySQL.

## Proposed Solutions

Move the json path traversal operators and the related expression functions definitions from the extensions modules to the general sea-query modules.

ChrisJr404:
MySQL supports the `->` and `->>` JSON field operators, just like SQLite and Postgres, but sea-query only exposed them through the SQLite and Postgres extensions. This adds the same capability for MySQL so JSON path traversal can be expressed against a MySQL backend too, which is what #631 asks for.

I went with an additive `MysqlExpr` extension trait that mirrors the existing `SqliteExpr` and `PgExpr` traits (`get_json_field` -> `->`, `cast_json_field` -> `->>`), rather than moving the operators into the general modules as the issue originally suggested. Keeping it as a per-backend extension matches how the other two backends already work and avoids any breaking change to the public API. The operators map to `JSON_EXTRACT` / `JSON_UNQUOTE(JSON_EXTRACT(...))`, and the right-hand side is a JSON path such as `"$.a"`.

Added integration tests for both operators (via `binary` and via the new trait methods) plus doctests on the trait, and confirmed `cargo test`, `cargo fmt --check` and `cargo clippy --all-targets` are clean.

Closes #631


src/extension/mysql/mod.rs
```diff
@@ -1,13 +1,35 @@
...
 pub(crate) use explain::MySqlExplainSchemaSpec;
+
+use crate::types::BinOper;
+
+/// MySQL-specific binary operators.
+///
+/// For all supported operators (including the standard ones), see [`BinOper`].
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum MysqlBinOper {
```
Huliiiiii, line 24: Should this be Mysql or MySql?

src/extension/mysql/mod.rs
```diff
@@ -1,13 +1,35 @@
...
+    /// `->`. Retrieves a JSON value referenced by a path as a JSON document.
```
Huliiiiii, line 25: These should be consistent with the official names.

PR #1103, opened 2026-08-30
feat(index): add Gin and Gist variants to IndexType and helper methods

itsbalamurali:
## Description
Adds `IndexType::Gin` and `IndexType::Gist` variants to `IndexType`, along with `.gin()` and `.gist()` helper methods on `IndexCreateStatement`.

## Motivation
Simplifies creation of Postgres GIN and GIST indexes on table definitions without relying on `IndexType::Custom`.

## Example
```rust
use sea_query::*;

let index = Index::create()
    .gin()
    .name("idx-glyph-image-gin")
    .table(Glyph::Table)
    .col(Glyph::Image)

assert_eq!(
    index,
    r#"CREATE INDEX "idx-glyph-image-gin" ON "glyph" USING GIN ("image")"#
);
```

### DIFF
--- a/src/backend/postgres/index.rs
+++ b/src/backend/postgres/index.rs
@@ -123,7 +123,8 @@ impl IndexBuilder for PostgresQueryBuilder {
             sql.write_str(" USING ").unwrap();
             match index_type {
                 IndexType::BTree => sql.write_str("BTREE"),
-                IndexType::FullText => sql.write_str("GIN"),
+                IndexType::FullText | IndexType::Gin => sql.write_str("GIN"),
+                IndexType::Gist => sql.write_str("GIST"),
                 IndexType::Hash => sql.write_str("HASH"),
                 IndexType::Custom(custom) => sql.write_str(&custom.0),
             }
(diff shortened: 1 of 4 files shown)


src/backend/postgres/index.rs
```diff
@@ -123,7 +123,8 @@ impl IndexBuilder for PostgresQueryBuilder {
             sql.write_str(" USING ").unwrap();
             match index_type {
                 IndexType::BTree => sql.write_str("BTREE"),
-                IndexType::FullText => sql.write_str("GIN"),
+                IndexType::FullText | IndexType::Gin => sql.write_str("GIN"),
```
Huliiiiii, line 126: Is FullText currently just an alias for Gin? 
Why is it necessary to add a new variant?

tests/postgres/index.rs
```diff
@@ -225,3 +225,29 @@ fn drop_5() {
         .table(("database", "schema", Glyph::Table))
         .to_string(PostgresQueryBuilder);
 }
+
```
Huliiiiii, line 228: These two tests are clearly unnecessary.

PR #1106, opened 2026-08-30
fix(backend/postgres): emit USING clause directly after TYPE in modify_column (#1099)

### LINKED ISSUE #1099: (PG) `using()` combined with `not_null()` emits invalid SQL — USING placed after SET NOT NULL
## Description

On Postgres, `ColumnDef::using()` combined with `not_null()` in a single
`modify_column` emits a statement Postgres cannot parse: the `USING` clause is
appended after `SET NOT NULL` instead of directly after `TYPE …`.

`USING` is only valid as part of the `ALTER COLUMN … TYPE …` clause, so once any
other clause follows the type change the output is a syntax error. This makes
`using()` unusable for the case it is most needed — converting a `NOT NULL`
`varchar` column to an enum, where Postgres refuses the implicit cast and the
`USING` clause is mandatory.

`using()` on its own is fine; only the combination is affected.

## Steps to reproduce

```rust
use sea_query::*;

#[derive(Iden)]
enum Character {
    Table,
    Id,
}

fn main() {
    let with_not_null = Table::alter()
        .table(Character::Table)
        .modify_column(
            ColumnDef::new(Character::Id)
                .integer()
                .not_null()
                .using(Expr::col(Character::Id).cast_as("integer")),
        )
        .to_owned();
    println!("{}", with_not_null.to_string(PostgresQueryBuilder));
}
```

## Expected Behavior

`USING` stays attached to the type change, e.g.

```sql
ALTER TABLE "character"
  ALTER COLUMN "id" TYPE integer USING CAST("id" AS integer),
  ALTER COLUMN "id" SET NOT NULL
```

## Actual Behavior

```sql
ALTER TABLE "character" ALTER COLUMN "id" TYPE integer, ALTER COLUMN "id" SET NOT NULL USING CAST("id" AS integer)
```

Postgres 16 rejects it:

```
ERROR:  syntax error at or near "USING"
LINE 1: ... "id" TYPE integer, ALTER COLUMN "id" SET NOT NULL USING CAS...
                                                             ^
```

Without `not_null()` the output is correct and accepted:

```sql
ALTER TABLE "character" ALTER COLUMN "id" TYPE integer USING CAST("id" AS integer)
```

## Reproduces How Often

Every time, for any `ColumnDef` in `modify_column` that sets both `using` and
any of `nullable` / `default` / `unique` / `primary_key` / `check` — `using` is
written after all of them.

## Versions

sea-query 1.0.1, PostgresQueryBuilder, Postgres 16, macOS (builder output is
platform-independent).

## Additional Information

Cause looks to be `prepare_modify_column` in `src/backend/postgres/table.rs`.
The `TYPE` clause is written first, the remaining specs are appended as
comma-separated `ALTER COLUMN …` clauses, and `using` is written last, outside
any clause:

```rust
if let Some(column_type) = &column_def.types { /* ALTER COLUMN … TYPE … */ }
...
if let Some(nullable) = column_def.spec.nullable { /* , ALTER COLUMN … SET NOT NULL */ }
...
if let Some(expr) = &column_def.spec.using {
    write!(sql, " USING ").unwrap();   // appended at the very end
    QueryBuilder::prepare_expr(self, expr, sql);
}
```

Writing `USING` immediately after `prepare_column_type` (inside the
`column_def.types` branch) would keep it in the clause it belongs to and make
the combination valid.

The doc example on `using()` only covers the single-spec case, which is why the
output looks correct there.

This is a follow-up to #626, which added the clause.

Happy to open a PR if the suggested placement is the direction you'd want.

itsbalamurali:
## Description
Fixes PostgreSQL statement preparation in `prepare_modify_column` so that `USING` expression clauses are emitted directly after the `ALTER COLUMN ... TYPE ...` clause rather than at the end of the entire `modify_column` statement after `SET NOT NULL` / `DEFAULT`.

Fixes #1099.

## Motivation
When modifying a column in Postgres to change both its type and nullability (e.g. converting a column to an enum with `USING` cast and `NOT NULL`), Postgres rejects `SET NOT NULL USING ...` with a syntax error because `USING` is only valid directly attached to the `TYPE` clause.

## Example
```rust
use sea_query::*;

let statement = Table::alter()
    .table(Character::Table)
    .modify_column(
        ColumnDef::new(Character::Id)
            .integer()
            .not_null()
            .using(Expr::col(Character::Id).cast_as("integer")),
    )

assert_eq!(
    statement,
    r#"ALTER TABLE "character" ALTER COLUMN "id" TYPE integer USING CAST("id" AS integer), ALTER COLUMN "id" SET NOT NULL"#
);
```

### DIFF
--- a/src/backend/postgres/table.rs
+++ b/src/backend/postgres/table.rs
@@ -370,6 +370,10 @@ impl PostgresQueryBuilder {
             self.prepare_iden(&column_def.name, sql);
             write!(sql, " TYPE ").unwrap();
             self.prepare_column_type(column_type, sql);
+            if let Some(expr) = &column_def.spec.using {
+                write!(sql, " USING ").unwrap();
+                QueryBuilder::prepare_expr(self, expr, sql);
+            }
             is_first = false;
         }
 
@@ -420,9 +424,11 @@ impl PostgresQueryBuilder {
             let _ = x;
         }
 
-        if let Some(expr) = &column_def.spec.using {
-            write!(sql, " USING ").unwrap();
-            QueryBuilder::prepare_expr(self, expr, sql);
+        if column_def.types.is_none() {
+            if let Some(expr) = &column_def.spec.using {
+                write!(sql, " USING ").unwrap();
+                QueryBuilder::prepare_expr(self, expr, sql);
+            }
         }
 
         if let Some(extra) = &column_def.spec.extra {
--- a/tests/postgres/table.rs
+++ b/tests/postgres/table.rs
@@ -871,3 +871,21 @@ fn create_partition_values_less_than_panics() {
         .values_less_than([10])
         .to_string(PostgresQueryBuilder);
 }
+
+#[test]
+fn test_alter_column_using_with_not_null() {
+    let statement = Table::alter()
+        .table(Char::Table)
+        .modify_column(
+            ColumnDef::new(Char::Id)
+                .integer()
+                .not_null()
+                .using(Expr::col(Char::Id).cast_as("integer")),
+        )
+        .to_string(PostgresQueryBuilder);
+
+    assert_eq!(
+        statement,
+        r#"ALTER TABLE "character" ALTER COLUMN "id" TYPE integer USING CAST("id" AS integer), ALTER COLUMN "id" SET NOT NULL"#
+    );
+}


src/backend/postgres/table.rs
```diff
@@ -420,9 +424,11 @@ impl PostgresQueryBuilder {
             let _ = x;
         }
 
-        if let Some(expr) = &column_def.spec.using {
-            write!(sql, " USING ").unwrap();
-            QueryBuilder::prepare_expr(self, expr, sql);
+        if column_def.types.is_none() {
```
Huliiiiii, line 427: I don't think we need to emit using clause when types is none.

tests/postgres/table.rs
```diff
@@ -871,3 +871,21 @@ fn create_partition_values_less_than_panics() {
         .values_less_than([10])
         .to_string(PostgresQueryBuilder);
 }
+
+#[test]
+fn test_alter_column_using_with_not_null() {
```
Huliiiiii, line 876: I think the name of this test does a poor job of describing its purpose.

Also, don't we already have tests covering this? Given how this feature is implemented, I think we should call as many methods as possible for each mutually exclusive case and assert their outputs.

